PR #9, opened 2025-03-03
4 initialize server

Christiantyemele:


Ogenbertrand:
Please @Hermann-Core can you take a look at this, Christian wants to proceed.

src/database/README.md
```diff
@@ -0,0 +1,49 @@
...
+|--------|------|-------------|
+| `id`        | SERIAL | Auto-incrementing primary key |
+| `issuer`    | TEXT  | Unique identifier for the issuer |
+| `statuslisttoken`| JSONB | status list token|
+
+## Usage
+
+```rust
+// implement Repository for your Store<T>
+impl<T> Repository<T> for Store<T>
```
Hermann-Core, line 28: Where does the `Store<T>` type defined? You should provide it along with the Trait bounds required (if any).

src/database/README.md
```diff
@@ -0,0 +1,49 @@
...
+    T: Serialize + for<'de> Deserialize<'de>,
+{
+    fn get_table(&self) -> Table<T> {
+        self.table.clone()
+    }
+}
+
+// create a store instance of credentials
+let store: Store<Credentials> = Store {
+
```
Hermann-Core, line 42: unnecessary line.

src/database/README.md
```diff
@@ -0,0 +1,49 @@
...
+            table: Table::new(conn, "credentials", "issuer".to_string()),
+        };
+
+// insert into database
+store.insert_one(credential).await.unwrap()
```
Hermann-Core, line 47: In the documentation, we generally avoid using `unwrap` to encourage users to handle errors.
```suggestion
store.insert_one(credential).await?;
```

src/database/connection.rs
```diff
@@ -0,0 +1,20 @@
+use std::env;
+
+use sqlx::{postgres::PgPoolOptions, Executor, PgPool};
+
+/// establish database connection with migrations  
+#[allow(unused)]
```
Hermann-Core, line 6: Why writing a function if it is not going to be used?

src/database/connection.rs
```diff
@@ -0,0 +1,20 @@
...
+pub async fn establish_connection() -> PgPool {
+    let url = env::var("DATABASE_URL").expect("DATABASE_URL env not set");
+    let pool = PgPoolOptions::new().connect(&url).await.unwrap();
+
+    run_migration(&pool).await;
+    pool
+}
+
+#[allow(unused)]
```
Hermann-Core, line 15: Why writing a function if it is not going to be used?

src/database/connection.rs
```diff
@@ -0,0 +1,20 @@
...
+async fn run_migration(pool: &PgPool) {
+    pool.execute(include_str!("./migrations/001_status_list.sql"))
+        .await
+        .expect("Failed to initialize DB");
```
Hermann-Core, line 19: Could you instead of panicking properly handle errors and propagate it up to the main function if necessary?

src/database/error.rs
```diff
@@ -0,0 +1,14 @@
...
+#[allow(clippy::enum_variant_names)]
+pub enum RepositoryError {
+    #[error("could not store entity")]
+    StoreError,
+    #[error("could not fetch entity")]
+    FetchError,
+    #[error("could not update entity")]
+    UpdateError,
+    #[error("could not delete entity")]
+    DeleteError,
```
Hermann-Core, line 13: Appending **Error** to each enum variant is not correct Rust idioms. We already know that they are errors. Choose better descriptives names.

src/database/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Describes a table instance
+pub struct Table<T>
+where
+    T: Serialize + for<'de> Deserialize<'de>,
+{
+    pub pool: PgPool,
+    pub table_name: String,
+    // column serving as query field to table
+    pub column: String,
+    _phantom: std::marker::PhantomData<T>,
```
Hermann-Core, line 19: What is its purpose of using a `PhantomData` here?

src/database/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub table: Table<T>,
+}
+
+#[allow(unused)]
+impl<T> Table<T>
+where
+    T: Serialize + for<'de> Deserialize<'de>,
+{
+    /// Creates a new `Table` instance.
+    pub fn new(pool: PgPool, table_name: impl Into<String>, column: String) -> Self {
```
Hermann-Core, line 39: The parameters names are not consistent. Passing `&str` are preferable. Or directly passing `String`.

Hermann-Core (changes requested):

.gitignore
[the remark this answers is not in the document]
Christiantyemele: i prefer this folders to be created when needed since they are not to be pushed

src/database/README.md
Christiantyemele: okay

src/database/connection.rs
Christiantyemele: this is because it is not used for now and the workflow won't passed

src/database/connection.rs
Christiantyemele: these are fatal errors we want to panic on the point

Hermann-Core: > these are fatal errors we want to panic on the point

What do you call fatal error? And how to you distinguish between fatal errors and the others?

src/database/repository.rs
Christiantyemele: to avoid and unbounded type parameters and also to bing the generic and its lifetime to the Store struct

src/main.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
 fn main() {
     println!("Hello, world!");
 }
```
Ngha-Boris, line 4: The main.rs is currently empty don't you think we should initialize the database connection.

Cargo.toml
```diff
@@ -1,6 +1,16 @@
...
 [dependencies]
+async-trait = "0.1"
+dotenvy = "0.15"
+mongodb = "3.2"
+once_cell = "1.20"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+sqlx = { version = "0.8", features = ["postgres", "runtime-tokio"] }
+thiserror = "2.0"
+tokio = { version = "1.16", features = ["full"] }
```
Ngha-Boris, line 16: I think we need axum as one of the dependency and also please check to make sure that all basic dependencies(axum, tower etc ) are present in this file.

Ngha-Boris (changes requested):

src/database/connection.rs
Christiantyemele: this is fatal because it's the process of starting the database, why do i want to handle this error and continue without database ?

Ngha-Boris (approved):
LGTM

ndefokou (approved):

PR #11, opened 2025-03-05
10 establish connection with token issuer using signed jwt

Christiantyemele:


src/web/credential_handler.rs
```diff
@@ -0,0 +1,27 @@
...
+        }
+        Err(err) => {
+            tracing::error!("Failed to store credentials: {err:?}");
+            Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
+                RepositoryError::CouldNotStoreEntity.to_string(),
+            ))
+        }
+    }
+}
```
Ngha-Boris, line 27: All errors return as INTERNAL_SERVER_ERROR (500), which isn't appropriate for all cases. Should differentiate between:
Bad Request (400) for invalid credentials
Unauthorized (401) for authentication failures
Internal Server Error (500) for database issues

src/auth/authentication.rs
```diff
@@ -0,0 +1,18 @@
...
+    credentials: Credentials,
+    state: AppState,
+) -> Result<(), RepositoryError> {
+    // try storing connection with token issuer
+    if let Some(store) = state.repository {
+        Ok(store.credential_repository.insert_one(credentials).await?)
+    } else {
+        Err(RepositoryError::RepositoryNotSet)?
+    }
+}
```
Ngha-Boris, line 18: This function stores credentials without any validation of the incoming data. There should be validation of:
Public key format
Supported algorithms
Issuer format/validity

Ngha-Boris (changes requested):

src/auth/authentication.rs
Christiantyemele: you're absolutely right

Ngha-Boris (approved):

Ogenbertrand (approved):

PR #14, opened 2025-03-12
feat(model): implement U8Wrapper for u8 encoding/decoding and enhance…

Christiantyemele:
… struct definitions

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -35,7 +35,7 @@ impl Type<Postgres> for U8Wrapper {
 }
 
 // Implement `sqlx::Encode<Postgres>` for `U8Wrapper`
-impl<'q> Encode<'q, Postgres> for U8Wrapper {
+impl Encode<'_, Postgres> for U8Wrapper {
     fn encode_by_ref(
         &self,
         buf: &mut sqlx::postgres::PgArgumentBuffer,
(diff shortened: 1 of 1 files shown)


src/model.rs
```diff
@@ -17,19 +25,46 @@ impl Credentials {
...
+    fn decode(value: sqlx::postgres::PgValueRef<'r>) -> Result<Self, sqlx::error::BoxDynError> {
+        let decoded = i8::decode(value)?; // Decode as `i8` first
+        Ok(U8Wrapper(decoded as u8)) // Convert safely to `u8`
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, Default, Debug, PartialEq, Eq, FromRow, Type)]
 pub struct StatusList {
-    pub bits: u8,
+    pub bits: U8Wrapper,
```
Blindspot22, line 57: I think it's good to go apart from the fact that, I don't know what `u8Wrapper` is :)
I know `u8` is an 8-bit unsigned integer

Blindspot22 (approved):

src/model.rs
Christiantyemele: it's  a custom wrapper type i created around u8 cause i want to use NEWTYPE pattern to implement trait i don't own on u8

PR #21, opened 2025-03-19
feat(utils): add key generation and token parsing utilities; update d…

### LINKED ISSUE #18: Create function for jwt or cwt generation
This function is to generate and return reference tokens in jwt or cwt depending on the inputs

### LINKED ISSUE #348: Create function for jwt or/and cwt generation
closes https://github.com/adorsys/status-list-server/issues/18
This function is to generate and return reference tokens in jwt or cwt depending on the inputs

Christiantyemele:
…ependencies
closes https://github.com/adorsys/didcomm-mediator-rs/issues/348


Blindspot22 (commented):
Just for more security measures, you can ensure the generated private key file is stored securely by verifying that `KEY_STORAGE` points to a secure location, such as a dedicated directory with limited access or explicitly set file permissions (fs::set_permissions) to restrict access.

i see what you mean here but this too is not enough security, i will prefer not to think the fully of security measure now, but this should be a future ticket

Hermann-Core (changes requested):
Each PR should focus on a single independent feature or bug fix, rather than combining multiple changes in one, as seen here. This could have been split into two separate features.
Additionally, the title does not accurately reflect what is being implemented.

Can you review the comments I left below and address them?

ndefokou:
@Christiantyemele, you should take a look at @Hermann-Core's reviews. They make a lot of sense, especially the ones concerning the keygen.

Christiantyemele:
i am converting this ticket to draft now cause 
1. it contains too many fixes which are not related and i will prefer to split them down
2. the scope it not too clear and needs to be reconsiderred

PR #23, opened 2025-03-20
feat(status): add flate2 for compression and update status handling

Christiantyemele:
closes https://github.com/adorsys/didcomm-mediator-rs/issues/360


Blindspot22 (commented):
Instead of using strings like "VALID" and "INVALID", why don't you use a strongly-typed enum with serialization and deserialization traits.

src/web/handlers/status_list/handler.rs
```diff
@@ -318,13 +403,118 @@ mod test {
...
+        );
+    }
+
+    #[tokio::test]
+    async fn test_update_with_bits_value_set_to_2() {
+        let original_lst = vec![0b00011011, 0b00000000]; // Each byte holds 4 statuses (2 bits each)
+
+        // Compress and encode the original status list
+        let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
+        encoder.write_all(&original_lst).unwrap();
```
Ngha-Boris, line 483: `Compression::default()` may be slow for small lst sizes. So I will suggest we use this `Compression::fast()` for better performance

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,99 @@ pub async fn update_statuslist(
...
-}
+pub fn update_status(
+    lst: String,
+    status_updates: Vec<StatusUpdate>,
+    bits: usize,
+) -> Result<String, StatusListError> {
+    if status_updates.is_empty() {
+        return Err(StatusListError::Generic(
+            "No status updates provided".to_string(),
+        ));
```
Ngha-Boris, line 208: No validation for empty lst, risking decompression failures I suggest we check and fallback

Ngha-Boris (changes requested):

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,104 @@ pub async fn update_statuslist(
...
+        .read_to_end(&mut status_array)
+        .map_err(|e| StatusListError::DecompressionError(e.to_string()))?;
+
+    // Determine the highest index in the updates to ensure the array is large enough
+    let max_update_index = status_updates
+        .iter()
+        .map(|update| update.index)
+        .max()
+        .unwrap_or(0);
+    if max_update_index < 0 {
```
Blindspot22, line 237: `index` is unsigned, so `max_update_index < 0` is redundant.

Since `max_update_index` is derived from `update.index`, I don't know but if `update.index` is an unsigned integer, then `max_update_index` will also always be ≥ 0. Meaning can never be negative or can only store non-negative numbers.

Christiantyemele: yep you are right i will go as far as saying the check is useless

Blindspot22 (dismissed):
This is good to go :rocket:

Christiantyemele:
@Ngha-Boris its left just you

Ngha-Boris:
> @Ngha-Boris its left just you

Let me look at it

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,101 @@ pub async fn update_statuslist(
...
 
-    Ok(encode_lst(decoded_lst))
+    // Compress the updated status array using zlib
+    let mut encoder = ZlibEncoder::new(Vec::new(), Compression::fast());
+    encoder
+        .write_all(&status_array)
+        .map_err(|e| StatusListError::CompressionError(e.to_string()))?;
+    let compressed = encoder
+        .finish()
+        .map_err(|e| StatusListError::CompressionError(e.to_string()))?;
```
IngridPuppet, line 291: Compressing from a byte array to a base64url-encoded value as per the specs is a critical operation. I suggest you extract this into a separate function and test the logic in isolation. Consider these test vectors from the spec:
- [0xb9, 0xa3] -> eNrbuRgAAhcBXQ
- [0xc9, 0x44, 0xf9] -> eNo76fITAAPfAgc
- https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-test-vectors-for-status-lis

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,101 @@ pub async fn update_statuslist(
...
+                    status_array[next_byte_index] &= !next_mask;
+                    status_array[next_byte_index] |= status_value >> next_bit_offset;
+                }
+            }
+        }
     }
 
-    Ok(encode_lst(decoded_lst))
+    // Compress the updated status array using zlib
+    let mut encoder = ZlibEncoder::new(Vec::new(), Compression::fast());
```
IngridPuppet, line 285: The spec is recommending to use the highest level of compression. Level 9 matches the examples in the spec when I tested locally. Is using `Compression::fast()` compatible with that?

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,101 @@ pub async fn update_statuslist(
...
+    encoder
+        .write_all(&status_array)
+        .map_err(|e| StatusListError::CompressionError(e.to_string()))?;
+    let compressed = encoder
+        .finish()
+        .map_err(|e| StatusListError::CompressionError(e.to_string()))?;
+
+    // Base64url encode the compressed data without padding
+    Ok(encode(&compressed))
```
IngridPuppet, line 294:
```suggestion
    Ok(base64_url::encode(&compressed))
```

src/web/handlers/status_list/handler.rs
```diff
@@ -247,15 +322,21 @@ mod test {
...
-        let status = vec![2, 1, 3, 1];
-        let lst = encode(status);
+        let status = vec![0b11111111, 0b01101110];
+        // compress status
+        let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
+        encoder.write_all(&status).unwrap();
+        let compressed_status = encoder.finish().unwrap();
+
+        // encode status lst
+        let lst = encode(compressed_status);
```
IngridPuppet, line 336: With a separate function, you would have simply reused the logic here. You're not really testing it with no raw value whatsoever.

src/web/handlers/status_list/handler.rs
```diff
@@ -318,13 +405,118 @@ mod test {
...
             .clone();
 
+        // decode lst
+        let decoded_lst = base64url::decode(&shared_lst).unwrap();
+
+        let mut decompressed_lst = ZlibDecoder::new(&*decoded_lst);
+        let mut decompressed_array = vec![];
+        decompressed_lst
+            .read_to_end(&mut decompressed_array)
+            .unwrap();
```
IngridPuppet, line 415: Why do you compute all this? I don't see where it is used.

src/web/handlers/status_list/handler.rs
```diff
@@ -318,13 +405,118 @@ mod test {
...
+
+        let expected_lst = vec![0b00010011, 0b00000000, 0b00000000]; // Adjusted statuses with added index 8
+
+        assert_eq!(
+            decompressed_lst, expected_lst,
+            "The updated status list does not match the expected list."
+        );
+    }
+
```
IngridPuppet, line 516: I believe more tests are necessary.
- Negative tests are as important as tests on happy paths. What happens with an invalid index, an invalid encode lst, a wrong bit size (bits), etc?
- It is VERY important you cover the test vectors in the specs. I haven't seen a single raw value of base64url_encoded lst in the current tests, which is definitely concerning. https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-compressed-byte-array

IngridPuppet (changes requested):

src/web/handlers/status_list/handler.rs
```diff
@@ -128,12 +132,14 @@ pub async fn update_statuslist(
...
     };
+
     if let Some(status_list_token) = status_list_token {
         let lst = status_list_token.status_list.lst.clone();
+        let bits = status_list_token.status_list.bits as usize;
 
         // Apply updates
 
-        let updated_lst = match update_status(&lst, updates) {
+        let updated_lst = match update_status(lst, updates, bits) {
```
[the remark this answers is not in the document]
Christiantyemele: yes

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,101 @@ pub async fn update_statuslist(
...
+    bits: usize,
+) -> Result<String, StatusListError> {
+    if lst.is_empty() {
+        return Err(StatusListError::Generic(
+            "No status list provided".to_string(),
+        ));
+    }
+    if status_updates.is_empty() {
+        return Err(StatusListError::Generic(
+            "No status updates provided".to_string(),
```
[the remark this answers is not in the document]
Christiantyemele: yes you're definitely right

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,101 @@ pub async fn update_statuslist(
...
+        ));
+    }
+
+    // Validate the 'bits' parameter
+    if ![1, 2, 4, 8].contains(&bits) {
+        return Err(StatusListError::UnsupportedBits);
+    }
```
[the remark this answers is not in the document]
Christiantyemele: yes i will definitely create a ticket for it

src/web/handlers/status_list/handler.rs
```diff
@@ -191,34 +197,101 @@ pub async fn update_statuslist(
...
-        base64url::decode(lst).map_err(|e| StatusListError::Generic(e.to_string()))?;
+    // Decode the existing Base64-encoded status list
+    let compressed_data = decode(&lst).map_err(|_| StatusListError::DecodeError)?;
+
+    // Decompress the data using zlib
+    let mut decoder = ZlibDecoder::new(&compressed_data[..]);
+    let mut status_array = Vec::new();
+    decoder
+        .read_to_end(&mut status_array)
+        .map_err(|e| StatusListError::DecompressionError(e.to_string()))?;
```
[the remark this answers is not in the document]
Christiantyemele: yes makes sense i created used the existing function

src/web/handlers/status_list/handler.rs
Christiantyemele: yes

Blindspot22 (dismissed):
Looks good :rocket:

src/utils/lst_gen.rs
```diff
@@ -1,96 +1,102 @@
...
                 let next_byte_index = byte_index + 1;
                 let next_bit_offset = 8 - bit_offset;
                 let next_mask = (1 << (bits - next_bit_offset)) - 1;
-                status_array[next_byte_index] &= !next_mask;
-                status_array[next_byte_index] |= status_value >> next_bit_offset;
+                if next_byte_index < status_array.len() {
+                    status_array[next_byte_index] &= !next_mask;
+                    status_array[next_byte_index] |= status_value >> next_bit_offset;
+                }
             }
```
IngridPuppet, line 87: I recall leaving a comment stating that this case is not possible, I'm not sure where it has gone to. The spec is designed so status values do not span across two bytes. Can you prove me wrong?

src/web/handlers/status_list/handler.rs
```diff
@@ -191,42 +203,13 @@ pub async fn update_statuslist(
...
-            Status::SUSPENDED => 2,
-            Status::APPLICATIONSPECIFIC => 3,
-        };
-    }
-
-    Ok(encode_lst(decoded_lst))
-}
-
 #[cfg(test)]
 mod test {
```
IngridPuppet, line 208: Test the vectors in the spec. https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-test-vectors-for-status-lis

IngridPuppet (changes requested):

src/utils/lst_gen.rs
Christiantyemele: > I recall leaving a comment stating that this case is not possible, I'm not sure where it has gone to. The spec is designed so status values do not span across two bytes. Can you prove me wrong?

Thanks for pointing this out! I checked the spec again it explicitly states that “bits are packed across byte boundaries if necessary.”
This means that depending on the index and the bits per entry, a status entry can span two bytes.

For example, with 4-bit entries: index 2 will occupy bits 0–3 of the second byte, having crossed from byte 0. So the current implementation handles this scenario as required by the spec.

IngridPuppet: > Thanks for pointing this out! I checked the spec again it explicitly states that “bits are packed across byte boundaries if necessary.”

Could you share a screenshot of the line you seem to quote? I'm not seeing it in the specs.

> For example, with 4-bit entries: index 2 will occupy bits 0–3 of the second byte, having crossed from byte 0. So the current implementation handles this scenario as required by the spec.

0--3 makes 4 bits. Nothing is stored in byte 0. What is your definition of spanning across?

Christiantyemele: > > Thanks for pointing this out! I checked the spec again it explicitly states that “bits are packed across byte boundaries if necessary.”
> 
> Could you share a screenshot of the line you seem to quote? I'm not seeing it in the specs.
> 
> > For example, with 4-bit entries: index 2 will occupy bits 0–3 of the second byte, having crossed from byte 0. So the current implementation handles this scenario as required by the spec.
> 
> 0--3 makes 4 bits. Nothing is stored in byte 0. What is your definition of spanning across?
when i talk of spanning across i mean moving to the next byte in the byte array, take a look at this example below where the the bits value is 2 meaning two bits  repesent the status of a token in that case trying to get the status of a token with index 4 will required use moving (spanning) to the next byte.
Screenshot from 2025-04-09 09-11-57

src/web/handlers/status_list/handler.rs
Christiantyemele: > Test the vectors in the spec. https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-test-vectors-for-status-lis

I wanted to test against the vectors provided in the spec ([link](https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-test-vectors-for-status-lis)), but unfortunately they're too large to use directly. Instead, I created shorter test vectors following the same structure and logic.

IngridPuppet: No, they aren't too large to use directly. Nonetheless, if you prefer shorter examples, you'll find two in this section: https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-status-list-in-json-format. Note that your implementation isn't really reliable when it doesn't align with a truth reference provided by a trusted source, especially when that exists.

src/utils/lst_gen.rs
IngridPuppet: I'm still not reading what you quoted “bits are packed across byte boundaries if necessary.” On the screenshot you appear to misunderstand, there are twelve status entries. Which one do you think is spanning across bytes?

I have a better quote for you: "This limitation is intended to limit bit manipulation necessary to a single byte for every operation and thus keeping implementations simpler and less error-prone." https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-compressed-byte-array

I'm suggesting you approach the topic mathematically.

src/web/handlers/status_list/handler.rs
Christiantyemele: okay

src/utils/lst_gen.rs
IngridPuppet, line 1: Please try to understand the logic of this test I wrote, and take measures to make it pass. It is based on an example from the spec: https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-status-list-in-json-format. Feel free to fix the operations in the test if you believe there is anything wrong. The one thing I want you to test if that you can get to `eNo76fITAAPfAgc` like in the spec.

```rust
    #[test]
    fn test_update_status_with_bits_set_to_2_specs() {
        let original_status_array = vec![0b11001001, 0b01000100, 0b11111001];

        let mut encoder = ZlibEncoder::new(Vec::new(), Compression::fast());
        encoder.write_all(&original_status_array).unwrap();
        let compressed_status = encoder.finish().expect("Failed to finish compression");
        let existing_lst = base64url::encode(compressed_status);

        assert_eq!(existing_lst, "eNo76fITAAPfAgc");

        // This update changes nothing. It is just there so we can test the decoding and encoding
        // run by `update_or_create_status_list`.
        let status_updates = vec![StatusEntry {
            index: 0,
            status: Status::INVALID,
        }];

        let bits = BitFlag::new(2).ok_or(Error::UnsupportedBits).unwrap();
        let updated_lst = update_or_create_status_list(Some(existing_lst), status_updates, bits)
            .expect("Failed to update status list");

        assert_eq!(updated_lst, "eNo76fITAAPfAgc");
    }
  ```

IngridPuppet (changes requested):

src/utils/lst_gen.rs
Christiantyemele: Yes @IngridPuppet thanks for the suggestions i noticed the compression levels actually affect the end output before i was using `Compression::fast()` as i thought required by the spec it has the description `  /// Optimize for the best speed of encoding.` meanwhile the test data uses `Compression::best()` with the description ` /// Optimize for the size of data being encoded.`
Screenshot from 2025-04-14 16-13-01
From the image the spec tells use to use the highest compression level not the fastest, fast() has value of 1 and best() has value of 9 which is the higest level.

IngridPuppet (approved):

Ngha-Boris (approved):

PR #26, opened 2025-03-25
feat: implement authentication middleware and update route handling for status lists

Christiantyemele:


src/web/handlers/credential_issuance.rs
```diff
@@ -5,7 +5,6 @@ use crate::{
     model::Credentials, utils::state::AppState,
 };
 
-#[axum::debug_handler]
```
Ngha-Boris, line 8: Why did you remove this?

Christiantyemele: it's development code not for production

Ngha-Boris (approved):

PR #27, opened 2025-03-26
feat: create endpoint for issuer to publish token status

Ngha-Boris:


Christiantyemele (commented):
Since the draft does not specifies how communication is done between the issuer and the status issuer in case they are different entities and lays more of the work on the developper, i propose we build the status list on our side instead of putting the work on the issuer as we have a single server for multiple issuers and it will be wise implementing that logic onces, and for this reason we can allow issues to publish the state of their token JSON format .e.g
```JSON
[{index: 0, status: VALID}, {index:1, status:INVALID}, ...]
````
Then we will create a function to generate status lists `lst` from this index and status 
from here we can generate the status list, this is my opinion @Ngha-Boris

**Review by Christiantyemele — APPROVED** [review]
Seal Of Quality.

File: src/web/handlers/status_list/publish_token_status.rs [hunk 1]
```diff
@@ -0,0 +1,308 @@
...
+            tracing::info!("Status list {} already exists", payload.list_id);
+            Err(StatusListError::StatusListAlreadyExists)
+        }
+        Ok(None) => {
+            // Calculate issuance and expiration timestamps
+            let iat = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs() as i64;
+            let exp = payload.ttl.as_ref().and_then(|ttl| {
```
Blindspot22, line 78:
```suggestion
            let exp = match payload.ttl.as_ref() {
                 Some(ttl) => match ttl.parse::<i64>() {
                           Ok(ttl_seconds) => Some(iat.saturating_add(ttl_seconds)),
                           Err(_) => return Err(StatusListError::Generic("Invalid TTL format".to_string())),
                 },
                 None => None,
            };
```

You can also add error handling for invalid `TTL` values instead of silently failing

Blindspot22 (changes requested):
Just a few suggestions I think would be good as Improvement before a merge :)

Blindspot22 (approved):
I think it's good to go :rocket:

Christiantyemele (approved):

PR #29, opened 2025-03-27
28 create function generate status list from token index and status

### LINKED ISSUE #28: Create function generate status list from token index and status
closes https://github.com/adorsys/didcomm-mediator-rs/issues/359

Christiantyemele:


src/utils/lst_gen.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use crate::model::Status;
+
+use super::errors::Error;
+
+pub struct PublishStatus {
+    pub index: i32,
+    pub status: Status,
+}
+
```
Ngha-Boris, line 13: Don't you think we need to have `#[derive(Clone)]` on this struct if we need to use it in another file.

Ngha-Boris (changes requested):

Ngha-Boris (approved):

ndefokou (approved):

PR #34, opened 2025-04-02
Server keypair

Christiantyemele:
closes https://github.com/adorsys/didcomm-mediator-rs/issues/367


Hermann-Core (changes requested):
We could also leverage the existing did-utils crate, particularly its rich key handling features, to avoid redundant implementations.

src/utils/keygen.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl Keypair {
+    /// Generate a new random keypair
+    pub fn generate() -> Result<Self, Error> {
+        let mut seed = [0u8; SECRET_KEY_LENGTH];
+        OsRng.try_fill_bytes(&mut seed).map_err(|err| {
+            tracing::error!("Failed to generate random bytes: {err:?}");
+            Error::KeyGenFailed
+        })?;
+        let signing_key = SigningKey::from_bytes(&seed);
```
IngridPuppet, line 30: Is there a reason you wouldn't trust the generate method of the library as presented in its description?

```rust
use rand::rngs::OsRng;
use ed25519_dalek::SigningKey;
use ed25519_dalek::Signature;

let mut csprng = OsRng;
let signing_key: SigningKey = SigningKey::generate(&mut csprng);
```

src/utils/keygen.rs
```diff
@@ -0,0 +1,108 @@
...
+        struct Claims {
+            exp: usize,
+        }
+
+        let keypair = Keypair::generate().unwrap();
+
+        let header = Header {
+            alg: jsonwebtoken::Algorithm::EdDSA,
+            ..Default::default()
+        };
```
IngridPuppet, line 79: Is it possible you consider ES256 signing instead? There is less wide support for EdDSA cryptography across ecosystems. For example, I faced a situation where it was not listed on jwt.io.

src/utils/keygen.rs
```diff
@@ -0,0 +1,108 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use ed25519_dalek::pkcs8::EncodePublicKey;
+    use jsonwebtoken::{decode, encode, DecodingKey, EncodingKey, Header, Validation};
+    use std::time::{SystemTime, UNIX_EPOCH};
+
+    #[test]
+    fn test() {
+        #[derive(Debug, serde::Deserialize, serde::Serialize)]
```
IngridPuppet, line 69: Could you rename this test or add a comment so we understand what it plans to test?

src/utils/keygen.rs
Hermann-Core: There are at least two main reasons for this change:

1. The `generate` function internally relies on the `fill_bytes` method from `OsRng`, which is no longer available in the latest version of the `rand` crate.
2. The `fill_bytes` method can panic in case of errors during random number generation. As noted in the (now outdated) documentation:
> This method should guarantee that `dest` is entirely filled with new data, and may panic if this is impossible (e.g., reading past the end of a file used as the source of randomness).

src/utils/keygen.rs
Hermann-Core: Didn't noticed that. Let me fix.

src/utils/keygen.rs
Hermann-Core: I see your point. Are we building these status list components for our internal use, or do we also plan to support integration with external tools?

IngridPuppet: I guess that's for signing status list tokens, which are meant to be verified by other parties. Even for internal use only, EdDSA would not be straightforward in JavaScript or Java environments, from experience.

Hermann-Core: Ok let's move to `EC` keys then.

IngridPuppet (approved):

Hermann-Core (approved):

PR #37, opened 2025-04-08
feat: introduce BitFlag struct for validated bit flag handling and up…

Christiantyemele:
…date related functions

### DIFF
--- a/src/utils/bits_validation.rs
+++ b/src/utils/bits_validation.rs
@@ -1,0 +1,50 @@
+/// A strongly-typed wrapper for valid BitFlag values (1, 2, 4, or 8).
+///
+/// This struct ensures that only valid single-bit values are used,
+/// preventing invalid values like 3, 5, or 6 from being constructed.
+///
+/// Use `BitFlag::new` or implement `TryFrom<u8>` to safely create an instance.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BitFlag(u8);
+
+impl BitFlag {
+    /// Attempts to create a new `BitFlag` instance from a given `u8` value.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` - A `u8` representing the BitFlag. Must be one of: 1, 2, 4, or 8.
+    ///
+    /// # Returns
+    ///
+    /// * `Some(BitFlag)` if the value is valid.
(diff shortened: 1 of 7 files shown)


src/utils/bits_validation.rs
IngridPuppet, line 1: By structure, I meant data structure, which can be an `enum` as well. I wouldn't call using `unwrap` everywhere a good move.

Christiantyemele: > By structure, I meant data structure, which can be an `enum` as well. I wouldn't call using `unwrap` everywhere a good move.

You wouldn't typically call unwrap everywhere once a `BitFlag` is constructed, the value can safely be accessed via a `.value()` method.

The only time you need to match is when creating the `BitFlag` from user input, at which point you can return an error for invalid input.

I actually find this approach cleaner than using an `enum` and matching on it throughout the codebase, especially if the values are simple and only need validation once.

IngridPuppet: Then, update your code and avoid unwrapping on non-user input like you are doing.

```rust
let bits = BitFlag::new(1).unwrap();
```

Besides, this statement is not true. You do not have to match on an enum **througout the codebase** to retrieve its integer representation.

> I actually find this approach cleaner than using an enum and matching on it throughout the codebase, especially if the values are simple and only need validation once.

src/utils/lst_gen.rs
```diff
@@ -108,7 +108,7 @@ mod tests {
                 status: Status::INVALID,
             },
         ];
-        let bits = BitFlag::new(1).unwrap();
+        let bits = BitFlag::new(1).ok_or(Error::UnsupportedBits).unwrap();
```
IngridPuppet, line 112: Adding that `ok_or` is even worse IMO. Can you just revert your changes and let me approve?

Christiantyemele: i don't how worse it is, cause all of this is even happens in the test, but okay let me revert

IngridPuppet: Oh you're right. I seemed to recall one of these was in the actual code. I'm okay.

IngridPuppet (approved):

Ngha-Boris (approved):

PR #42, opened 2025-04-11
containerize the status-list app

Ogenbertrand:
Building and image for the application. please take a look and live yo thoughts.


Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY src ./src
+
+# Build the actual application with optimizations
+RUN cargo build --release && \
+    strip target/release/status-list-server && \
+    ls -lh target/release/
+
+# Stage 2: Runtime - Create minimal production image
+FROM debian:bookworm-slim
+
```
Christiantyemele, line 25: use a minimal image here to run the binary `distroless` is perfect for this. https://bell-sw.com/blog/distroless-containers-for-security-and-size/

Christiantyemele: right now the docker registry is unavailable i will continue the review once the service will be up

Christiantyemele (changes requested):

docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: '3.9'
+
+services:
+  db:
+    image: postgres:15
+    container_name: status-list-server-db
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=status-list
```
Ngha-Boris, line 10: Consider using Docker secrets or environment variables from a .env file

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Verify file copy and set executable permissions
+RUN chmod +x /app/status-list-server
+
+# Expose the default network port
+EXPOSE 8080
+
+# Set default logging level
+ENV RUST_LOG=info
+
```
Ngha-Boris, line 40: The `ENV RUST_LOG=info` might be redundant since it’s overridden in `docker-compose.yaml`. Consider removing it from the `Dockerfile` to avoid confusion.

Ngha-Boris (changes requested):

.dockerignore
```diff
@@ -0,0 +1 @@
+/target
```
Awambeng, line 1: You can also ignore the local git files as it's not necessary

.github/workflows/ci-docker-ghcr.yml
```diff
@@ -0,0 +1,43 @@
+name: Build and push images to GHCR
+
+on:
+  push:
+    branches:
+      - issue-38
```
Awambeng, line 6: I guess the branch was for testing purpose. Please can you update this to the target branch

Dockerfile
Ogenbertrand: Thanks for pointing that out

docker-compose.yml
Ogenbertrand: We currently have no .env file in the project, and the values i am passing are for setting up the db.

.github/workflows/ci-docker-ghcr.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Build and push Docker image
+        id: docker_build
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: |
+            ghcr.io/${{ github.repository }}:latest
+            ghcr.io/${{ github.repository }}:${{ github.sha }}
```
Awambeng, line 43: Won't it be good if we explicitly add the different platforms for our image?

Awambeng, line 43: Using latest as a tag is convenient, but it’s not ideal for production use. Consider adopting [Semantic Versioning (SemVer)](https://semver.org/) for more reliable and traceable version tags like v1.0.0. If appropriate, you can implement SemVer-based tagging as part of your release workflow

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Copy actual source code (this happens after dependency caching)
+COPY src ./src
+
+# Build the actual application with optimizations
+RUN cargo build --release && \
+    strip target/release/status-list-server && \
+    ls -lh target/release/
+
+# Stage 2: Runtime - Create minimal production image
+FROM debian:bookworm-slim
```
Awambeng, line 24: I suggest we go with `debian:buster-slim` as it's well known and popular.

Dockerfile
Awambeng: > The `ENV RUST_LOG=info` might be redundant since it’s overridden in `docker-compose.yaml`. Consider removing it from the `Dockerfile` to avoid confusion.

I don't think it might cause an issue. But if you think it must be remove then @Ogenbertrand please remember adding it in your ci so that we can have more logs for debugging

docker-compose.yml
Awambeng: > Consider using Docker secrets or environment variables from a .env file

Your point makes sense but we can stay with this setup as it's just for local development and testing.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR /app
+
+# Copy only the built binary from the builder stage
+COPY --from=builder /app/target/release/status-list-server /app/status-list-server
+
+# Verify file copy and set executable permissions
+RUN chmod +x /app/status-list-server
+
+# Expose the default network port
+EXPOSE 8080
```
Awambeng, line 36: The server runs on 8000 and not 8080

docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+    networks:
+      - status-list-network
+
+  app:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: status-list-server
+    ports:
+      - "9090:8080"
```
Awambeng, line 29: The server runs on port 8000

Awambeng (changes requested):

Dockerfile
```diff
@@ -0,0 +1,42 @@
+# Stage 1: Build - Creates an optimized production binary
+FROM rust:1.85 as builder
```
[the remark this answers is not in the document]
Ogenbertrand: Yes thank u, i just crossed my mind.

.github/workflows/ci-docker-ghcr.yml
Awambeng: > I don't get this @Awambeng !

use this as reference https://docs.docker.com/build/ci/github-actions/multi-platform/#:~:text=To%20load%20a%20multi%2Dplatform,CLI%20settings%20for%20a%20job. check the build and push step and see how they added the platforms they want their image to support

Awambeng (changes requested):

.dockerignore
```diff
@@ -0,0 +1,9 @@
...
+**/target/
+/helm
```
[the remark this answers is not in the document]
Ogenbertrand: I haven't pushed it yet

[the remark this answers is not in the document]
Awambeng: > I haven't pushed it yet

I guess it's for the next ticket so remove the helm folder here and when working on the next ticket you can add it

.github/workflows/ci-docker-ghcr.yml
```diff
@@ -0,0 +1,44 @@
...
+        id: docker_build
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          platform: linux/amd64,linux/arm64
+          tags: |
+            ghcr.io/${{ github.repository }}:v1.0.0
+            ghcr.io/${{ github.repository }}:${{ github.sha }}
```
Awambeng, line 44: You hardcoded the version meaning we will have just v1.0.0 all the time. To ease your work take a look at this https://github.com/adorsys/eudiw-verifier/blob/develop/.github/workflows/push-ghcr-on-release.yml.
For the trigger we can have your what you added but will need to specify that of the tags

Awambeng (changes requested):

Ngha-Boris (approved):

Christiantyemele (approved):

Ngha-Boris (approved):

Christiantyemele (approved):

Ngha-Boris (approved):

.dockerignore
```diff
@@ -0,0 +1,7 @@
+/target
```
Awambeng, line 1: is the slash necessary?

.github/workflows/ci-docker-ghcr.yml
```diff
@@ -0,0 +1,51 @@
+name: Build and push images to GHCR
+
+on:
+  push:
+    tags:
+      - v*.*.*
```
Awambeng, line 6:
```suggestion
on:
  push:
    tags:
      - v*.*.*
    branches:
      - main
```

Awambeng: The main branch can equally be added so as to quickly testing our artefact before a release is done

Awambeng (changes requested):

.dockerignore
Ogenbertrand: Yes it is, to avoid large image size, no need to compile the target when building an image.

Awambeng: > Yes it is, to avoid large image size, no need to compile the target when building an image.

where is the target directory located? is it at the root of the project? if yes then just adding target is sufficient

Ogenbertrand: Where is the difference in what you are saying and what i did ?

Awambeng: why the slash?

Ogenbertrand: I don't see any difference, they'all signify the same thing.

Ogenbertrand: the slash simply signifiy i am ignoring a folder

PR #44, opened 2025-04-17
366 implement a parse to jwt or cwt

Hermann-Core:


src/web/handlers/status_list/handler.rs
```diff
@@ -72,13 +100,137 @@ pub async fn get_status_list(
...
+        ),
+        (
+            CborValue::Text("lst".into()),
+            CborValue::Text(token.status_list.lst.clone()),
+        ),
+    ];
+    claims.push((
+        CborValue::Integer(65533.into()),
+        CborValue::Map(status_list),
+    ));
```
IngridPuppet, line 164: Same, regarding the use of magic constants.

src/web/handlers/status_list/handler.rs
```diff
@@ -72,13 +100,137 @@ pub async fn get_status_list(
...
+            CborValue::Integer(65534.into()),
+            CborValue::Integer(ttl.into()),
+        ));
+    } else {
+        claims.push((
+            CborValue::Integer(65534.into()),
+            // Default to 12 hours
+            CborValue::Integer(43200.into()),
+        ));
+    }
```
IngridPuppet, line 149: Take the condition closer to what really changes? That is, the value of the ttl field.

src/web/handlers/status_list/handler.rs
```diff
@@ -72,13 +100,137 @@ pub async fn get_status_list(
...
+    ));
+
+    let payload = CborValue::Map(claims).to_vec().map_err(|err| {
+        tracing::error!("Failed to serialize claims: {err:?}");
+        StatusListError::InternalServerError
+    })?;
+
+    // Building the protected header
+    let protected = HeaderBuilder::new()
+        .algorithm(Algorithm::ES256)
```
IngridPuppet, line 173: Can the algorithm be recovered in some way from the signing key?

src/web/handlers/status_list/handler.rs
```diff
@@ -72,13 +100,137 @@ pub async fn get_status_list(
...
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct StatusListClaims {
+    pub exp: Option<i64>,
+    pub iat: i64,
+    pub status_list: StatusList,
+    pub sub: String,
+    pub ttl: Option<i64>,
+}
```
IngridPuppet, line 204: Isn't this what `status_list_tokens::Model` defines? Could you consider aliasing to it or recomposing it?

src/web/handlers/status_list/handler.rs
```diff
@@ -72,13 +100,137 @@ pub async fn get_status_list(
...
+    let iat = Utc::now().timestamp();
+    let ttl = token.ttl.unwrap_or(43200);
+    // Building the claims
+    let claims = StatusListClaims {
+        exp: token.exp,
+        iat,
+        status_list: token.status_list.clone(),
+        sub: token.sub.to_owned(),
+        ttl: Some(ttl),
+    };
```
IngridPuppet, line 216: What is your take on the issuer claim? Mind adding a comment?

src/web/handlers/status_list/handler.rs
```diff
@@ -52,18 +65,33 @@ pub async fn get_status_list(
...
+        // assume jwt by default if no accept header is provided
+        {
+            build_status_list_token(ACCEPT_STATUS_LISTS_HEADER_JWT, &list_id, &state).await
+        }
+        Some(accept) if accept == ACCEPT_STATUS_LISTS_HEADER_JWT => {
+            build_status_list_token(accept, &list_id, &state).await
+        }
+        Some(accept) if accept == ACCEPT_STATUS_LISTS_HEADER_CWT => {
+            build_status_list_token(accept, &list_id, &state).await
+        }
```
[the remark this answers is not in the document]
Hermann-Core: What do you mean?

src/web/handlers/status_list/handler.rs
Hermann-Core: To prevent restructuring the database record model defined by the `Model` type, I created this model that embeds only the necessary fields for the status list claims.

src/web/handlers/status_list/handler.rs
Hermann-Core: Don't think so. but you can get the algorithm used by checking the `alg` header of the token.

src/web/handlers/status_list/handler.rs
```diff
@@ -72,13 +100,137 @@ pub async fn get_status_list(
...
+            .into_response())
+    } else {
+        Ok((
+            StatusCode::OK,
+            [(header::CONTENT_TYPE, accept)],
+            issue_cwt(&status_claims, &server_key)?,
+        )
+            .into_response())
+    }
+}
```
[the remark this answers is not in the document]
Hermann-Core: The 2 functions return differents types, so it won't work that way.

IngridPuppet (approved):

Christiantyemele (approved):

PR #47, opened 2025-04-22
Add SecretManager for Secure Secret Handling

Christiantyemele:


src/utils/secretmanager.rs
```diff
@@ -0,0 +1,113 @@
...
+            .name(secret.secret_name)
+            .secret_string(secret.secret_value)
+            .send()
+            .await
+            .map_err(|e| {
+                tracing::error!("error: {}", e.to_string());
+                Error::Generic("Failed to store secret".to_string())
+            })?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
Ngha-Boris: The tests mock a `SecretsManager` trait with `get_secret` and `store_secret`, but the actual code uses an `Operations` trait with `get_key` and `store_key`. This means the tests don’t reflect the real implementation. Consider Updating the tests to mock the `Operations` trait.

src/utils/secretmanager.rs
```diff
@@ -0,0 +1,113 @@
...
+        Ok(response.secret_string().map(|s| s.to_string()))
+    }
+
+    async fn store_key(&self, secret: Secret) -> Result<(), Error> {
+        let asm = self::Client::configs(self.region.clone()).await;
+        asm.create_secret()
+            .name(secret.secret_name)
+            .secret_string(secret.secret_value)
+            .send()
+            .await
```
Ngha-Boris, line 59: The `Client` struct has a `secret_name` field, but its usage is incosistent:
In `get_key`, it uses `self.secret_name` to fetch a secret.
In `store_key`, it uses `secret.secret_name` from the provided Secret struct, ignoring `self.secret_name`.

src/utils/secretmanager.rs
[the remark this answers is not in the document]
Christiantyemele: you forgot about `line 61` where i am logging the full error, this generic error is just to provide a status of the operation

src/utils/secretmanager.rs
Christiantyemele: Yes, that’s intentional — the existing Client is tied to a specific secret identified by its secret_name. However, when creating a new secret, I can’t reuse the existing secret_name; instead, I need to provide a new one. That’s where the Secret struct comes in — it allows the creation of a new secret with a different name and value.

Ngha-Boris:
> > The `get_secret` uses `create_secret` , which fails if the secret already exist. I think the `store_key` is limited to only creating new secrets, with no ability to update existing ones, don't you think it will be good if we also updated them not focusng only on creating them?
> 
> No, we don’t want to allow modification or updating of secrets, as they are tightly bound to other entities like the certificate.

Okay then

src/utils/secretmanager.rs
Ngha-Boris: Okay I get it.

src/utils/secretmanager.rs
```diff
@@ -0,0 +1,66 @@
+use async_trait::async_trait;
+use aws_config::{BehaviorVersion, Region};
+use aws_sdk_secretsmanager::Client as awsClient;
+
+use super::errors::Error;
+pub struct Client {
```
Hermann-Core, line 6: Add space between the `use` statement and the struct.
Additionnally, choose a more suitable name for this struct. e.g.
```suggestion
pub struct AwsSecret {
```

src/utils/secretmanager.rs
```diff
@@ -0,0 +1,66 @@
+use async_trait::async_trait;
+use aws_config::{BehaviorVersion, Region};
+use aws_sdk_secretsmanager::Client as awsClient;
```
Hermann-Core, line 3: follow the conventional naming for types
```suggestion
use aws_sdk_secretsmanager::Client as AwsClient;
```

src/utils/secretmanager.rs
```diff
@@ -0,0 +1,66 @@
...
+    async fn store_key(&self, secret: Secret) -> Result<(), Error>;
+}
+
+impl Client {
+    pub async fn new(secret_name: String, _region: Region) -> Self {
+        let config = aws_config::defaults(BehaviorVersion::v2025_01_17())
+            .region(_region.clone())
+            .load()
+            .await;
+        let aws_client = aws_sdk_secretsmanager::Client::new(&config);
```
Hermann-Core, line 30: The `AwsClient` type is already imported. Use it directly.
```suggestion
        let aws_client = AwsClient::new(&config);
```

src/utils/secretmanager.rs
```diff
@@ -0,0 +1,66 @@
...
+#[async_trait]
+impl Operations for Client {
+    async fn get_key(&self) -> Result<Option<String>, Error> {
+        let asm = self.aws_client.clone();
+        let response = asm
+            .get_secret_value()
+            .secret_id(self.secret_name.clone())
+            .send()
+            .await
+            .map_err(|_| Error::Generic("Failed to get key".to_string()))?;
```
Hermann-Core, line 49: It will be better to log the error so that we can have a better context.

Hermann-Core (changes requested):

Cargo.toml
```diff
@@ -35,9 +35,13 @@ sea-orm = { version = "1.1.7", features = [
     "runtime-tokio-native-tls",
     "macros",
 ] }
+aws-config = "1.6"
+aws-sdk-secretsmanager = "1.68"
 
 
 [dev-dependencies]
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+mockall = "0.12"
```
Hermann-Core, line 44: where are using this dep?

Hermann-Core (changes requested):

Cargo.toml
Christiantyemele: fixed

Hermann-Core (approved):

Ngha-Boris (approved):

PR #53, opened 2025-04-23
feat: attach athentication midlw to various endpoints involved with t…

### LINKED ISSUE #51: Attach authentication midlw to various endpoints involve with the communication with issuer
The /credential endpoint currently allows registration of an issuer by storing their keys for future communication. However, the existing implementation does not enforce authentication checks when handling requests from the issuer. As a result, these communications are unauthenticated, leaving the system vulnerable to impersonation.

This ticket aims to address that gap by ensuring authentication middleware is applied to all relevant endpoints involved in issuer communication, thereby securing the interaction and preventing unauthorized access.

Ngha-Boris:
…he communication with issuer


src/auth/middleware.rs
```diff
@@ -0,0 +1,69 @@
+use axum::{
```
Christiantyemele, line 1: I don't if you're aware but there was already and existing midlw in the web folder you could just refacator that one

src/web/handlers/credential_issuance.rs
```diff
@@ -1,16 +1,26 @@
...
+    AuthenticatedIssuer(issuer): AuthenticatedIssuer,
     credential: Json<Credentials>,
 ) -> impl IntoResponse {
-    let credential = credential.0;
-    match publish_credentials(credential, appstate).await {
+    // Verify that the issuer in the credential matches the authenticated issuer
+    if credential.0.issuer != issuer {
+        return Err((
+            StatusCode::UNAUTHORIZED,
+            "Issuer mismatch: credential issuer does not match authenticated issuer".to_string(),
```
Christiantyemele, line 19: No here the issuer is not yet authenticated, this is still the registration phase

src/auth/middleware.rs
```diff
@@ -0,0 +1,69 @@
...
+
+                let issuer = header.kid.ok_or_else(|| {
+                    (
+                        StatusCode::UNAUTHORIZED,
+                        "Missing issuer identifier in token",
+                    )
+                        .into_response()
+                })?;
+
+                Ok(AuthenticatedIssuer(issuer))
```
Christiantyemele, line 58: How do you pass this authenticated user to other handlers , use `extensions` to pass the data to handlers see the existing midlw

Christiantyemele (changes requested):

Ngha-Boris:
closes https://github.com/adorsys/status-list-server/pull/61

src/web/handlers/status_list/handler.rs
```diff
@@ -239,8 +240,19 @@ fn issue_jwt(token: &StatusListToken, server_key: &Keypair) -> Result<String, St
...
 ) -> impl IntoResponse {
+    if list_id != issuer {
+        return (
+            StatusCode::UNAUTHORIZED,
+            StatusListError::Unauthorized(
+                "Issuer mismatch: list_id does not match authenticated issuer".to_string(),
+            ),
+        )
+            .into_response();
+    }
```
IngridPuppet, line 254: With this check, each registered issuer can only maintain a single status list, the one with the same ID as its issuer ID. I believe this is too rigid and particularly unwanted. Token issuers should be able to maintain multiple status lists with the same level of protection against modification by an unrelated party.

src/web/handlers/status_list/publish_token_status.rs
```diff
@@ -1,106 +1,97 @@
...
 #[derive(Deserialize, Serialize, Clone)]
 pub struct PublishTokenStatusRequest {
     pub list_id: String,
     pub updates: Vec<StatusEntry>,
-    #[serde(default)]
-    pub sub: Option<String>,
-    #[serde(default)]
-    pub ttl: Option<i64>,
     pub bits: u8,
 }
```
IngridPuppet, line 21: Thank you for removing unused fields. However, it appears you're depriving the issuer of the possibility to define a validity window for published statuses by removing `ttl`? Is that intentional?

src/web/handlers/status_list/publish_token_status.rs
```diff
@@ -1,106 +1,97 @@
...
+    AuthenticatedIssuer(issuer): AuthenticatedIssuer,
     Json(payload): Json<PublishTokenStatusRequest>,
 ) -> Result<impl IntoResponse, StatusListError> {
-    let store = &appstate.status_list_token_repository;
+    // Verify that the issuer matches the authenticated issuer
+    if payload.list_id != issuer {
+        return Err(StatusListError::Unauthorized(
+            "Issuer mismatch: list_id does not match authenticated issuer".to_string(),
+        ));
+    }
```
IngridPuppet, line 34: Same remark. It should be possible to associate multiple lists to a single issuer. Also, you would usually err with 403 in this situation, not 401.

src/web/handlers/status_list/publish_token_status.rs
```diff
@@ -1,106 +1,97 @@
...
+    // Create the new token
+    let now = Utc::now().timestamp();
+    let new_token = StatusListToken {
+        list_id: payload.list_id,
+        exp: Some(now + 3600), // 1 hour expiration
+        iat: now,
+        status_list,
+        sub: issuer,
+        ttl: Some(3600),
     };
```
IngridPuppet, line 82: How do you motivate the status server defining a ttl and exp time on behalf of the token issuer? Also, it is generally preferable to limit your changes to the topic of your ticket and rather create different tickets for handling unrelated concerns.

src/web/midlw.rs
```diff
@@ -1,16 +1,83 @@
...
 };
 
+#[derive(Clone)]
+pub struct AuthenticatedIssuer(pub String);
+
+impl<S> FromRequestParts<S> for AuthenticatedIssuer
+where
+    S: Send + Sync + 'static,
+    AppState: FromRef<S>,
+{
```
IngridPuppet, line 26: - Could you provide tests for this extractor? I have no idea of the JWT should be produced to match the server expectations.
- Did you even try the logic yourself? It is required that the JWT provided specifies a kid that matches a credential in the server's database. First, retrieving credentials by their ID does not appear to have been properly configured with SeaORM. Second, how is the issuer going to know or even identify by an ID generated by and only known to the server?

IngridPuppet (changes requested):
I added some comments. Please, could you check? Also, you closed https://github.com/adorsys/status-list-server/issues/57 in favor of this ticket. That ticket was about securing the `/credentials` endpoint, and I'm not seeing any change in that regard here. Mark the difference between the credentials endpoint and the endpoint for publishing status lists.

src/database/queries.rs
```diff
@@ -79,6 +79,30 @@ impl SeaOrmStore<StatusListToken> {
...
+        issuer: &str,
+        list_id: &str,
+    ) -> Result<bool, RepositoryError> {
+        let token = status_list_tokens::Entity::find_by_id(list_id)
+            .one(&*self.db)
+            .await
+            .map_err(|e| RepositoryError::FindError(e.to_string()))?;
+
+        Ok(token.map(|t| t.sub == issuer).unwrap_or(false))
+    }
```
IngridPuppet, line 94: This is more like a utility, and it does not really belong in here. Just run the check where needed. With this construction, you'll be repeatedly querying the database.

src/web/handlers/status_list/handler.rs
```diff
@@ -257,8 +257,8 @@ pub async fn update_statuslist(
 ) -> impl IntoResponse {
     if list_id != issuer {
         return (
```
IngridPuppet, line 259: You can't enforce the list ID to be the issuer label, basically because that does not allow a single issuer to have multiple lists. Do you understand the problem? Guess this is somehow related to #67. If you believe this is better fixed in another ticket, make sure you have it created.

src/web/handlers/status_list/publish_token_status.rs
```diff
@@ -12,25 +12,32 @@ use chrono::Utc;
...
+/// Request payload for publishing a status list token.
+/// This struct represents the data needed to create or update a status list token.
 #[derive(Deserialize, Serialize, Clone)]
 pub struct PublishTokenStatusRequest {
+    /// The unique identifier for the status list
     pub list_id: String,
+    /// List of status updates to be applied
     pub updates: Vec<StatusEntry>,
+    /// Number of bits per status entry (must be 1 or 2 as per IETF draft)
     pub bits: u8,
```
IngridPuppet, line 24: More values are allowed. See `BitFlag` in the codebase.

src/web/handlers/status_list/publish_token_status.rs
```diff
@@ -12,25 +12,32 @@ use chrono::Utc;
...
     Json(payload): Json<PublishTokenStatusRequest>,
 ) -> Result<impl IntoResponse, StatusListError> {
-    // Verify that the issuer matches the authenticated issuer
-    if payload.list_id != issuer {
-        return Err(StatusListError::Unauthorized(
-            "Issuer mismatch: list_id does not match authenticated issuer".to_string(),
-        ));
+    // Validate bits value according to IETF draft (must be 1 or 2)
+    if payload.bits != 1 && payload.bits != 2 {
+        return Err(StatusListError::UnsupportedBits);
```
IngridPuppet, line 40: More values are allowed. Reuse the existing validation logic if possible.

IngridPuppet (changes requested):

src/database/mod.rs
```diff
@@ -61,20 +68,30 @@ pub mod migrations {
                             )
                             .col(ColumnDef::new(StatusListTokens::Sub).string().not_null())
                             .col(ColumnDef::new(StatusListTokens::Ttl).string())
+                            .foreign_key(
+                                ForeignKey::create()
+                                    .name("fk_sub") // Foreign key name for the sub->issuer relationship
+                                    .from(StatusListTokens::Table, StatusListTokens::Sub)
+                                    .to(Credentials::Table, Credentials::Issuer),
+                            )
```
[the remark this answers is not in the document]
Ngha-Boris: The foreign key I added links the `sub` column in the `StatusListTokens` table to the `issuer` column in the `Credentials `table.
The purpose is to enforce referential integrity at the database level, ensuring that every status list token is associated with a valid, registered issuer.

src/database/mod.rs
```diff
@@ -61,20 +68,30 @@ pub mod migrations {
...
-                    .drop_table(Table::drop().table(Credentials::Table).to_owned())
+                    .drop_table(Table::drop().table(StatusListTokens::Table).to_owned())
                     .await?;
+
                 manager
                     .drop_table(Table::drop().table(StatusListTokens::Table).to_owned())
                     .await?;
+
                 Ok(())
             }
```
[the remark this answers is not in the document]
Ngha-Boris: you are right.

src/database/queries.rs
Ngha-Boris: okay

src/web/handlers/status_list/handler.rs
Ngha-Boris: I will just do it in this ticket

src/database/queries.rs
IngridPuppet: I can still see it in the code...

src/web/midlw.rs
```diff
@@ -146,3 +144,307 @@ pub async fn auth(
...
+        let result = AuthenticatedIssuer::from_request_parts(&mut parts, &app_state).await;
+        assert!(result.is_err());
+        let resp = result.err().unwrap().into_response();
+        assert_eq!(resp.status(), StatusCode::UNAUTHORIZED);
+        // Check body
+        let body = futures::executor::block_on(async {
+            let bytes = to_bytes(resp.into_body(), 1024 * 1024).await.unwrap();
+            String::from_utf8(bytes.to_vec()).unwrap()
+        });
+        assert!(body.contains("Invalid token format") || body.contains("Invalid token"));
```
IngridPuppet, line 307: Leave only one expected message. There can't be two, right?

src/web/midlw.rs
```diff
@@ -146,3 +144,307 @@ pub async fn auth(
...
+        // No headers set
+        let result = AuthenticatedIssuer::from_request_parts(&mut parts, &app_state).await;
+        assert!(result.is_err());
+        let resp = result.err().unwrap().into_response();
+        assert_eq!(resp.status(), StatusCode::UNAUTHORIZED);
+        // Check body
+        let body = futures::executor::block_on(async {
+            let bytes = to_bytes(resp.into_body(), 1024 * 1024).await.unwrap(); // Set limit to 1MB
+            String::from_utf8(bytes.to_vec()).unwrap()
+        });
```
IngridPuppet, line 252: You're already in an async function. I wonder why you need a `block_on` directive. The same comment goes for all the subsequent occurrences of this syntax.
```suggestion
            let bytes = to_bytes(resp.into_body(), 1024 * 1024).await.unwrap(); // Set limit to 1MB
            let body = String::from_utf8(bytes.to_vec()).unwrap();
```

src/database/mod.rs
```diff
@@ -72,10 +86,9 @@ pub mod migrations {
...
 
+            /// Drops the database tables
             async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
-                manager
-                    .drop_table(Table::drop().table(Credentials::Table).to_owned())
-                    .await?;
+                // Drop tables in reverse order to handle foreign key constraints
                 manager
                     .drop_table(Table::drop().table(StatusListTokens::Table).to_owned())
                     .await?;
```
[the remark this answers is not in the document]
Ngha-Boris: That was an error i will restore it.

src/web/midlw.rs
Ngha-Boris: OKay sir I will work on that.

src/database/queries.rs
Ngha-Boris: Hello Sir @IngridPuppet 
After revisiting the implementation, I believe the current approach of `is_issuer_owner` is still appropriate for a few reasons:

- It's a key security check to verify resource ownership, rather than a general utility.
- Its placement in the database layer helps preserve separation of concerns.
- The query is optimized and only triggered when necessary.
- Centralizing this logic helps maintain consistent ownership validation throughout the application.

Sir this is what I think. I am open to any idea that you have in mind.

IngridPuppet: Points 1 and 4 are okay. 2 and 3, not really. But that's not the main issue at this point... I realized you're no longer using the method, am I wrong?

src/web/midlw.rs
IngridPuppet: Checking both requires an AND (&&) not an OR (||). Additionally, a string that contains "Invalid token format", necessarily contains "Invalid token", so adding an AND condition is just not useful. In plain english, what do you really want to assert? An OR is not valid, because that'd be relying on randomness.

src/database/queries.rs
Ngha-Boris: Thank you, Sir

IngridPuppet (approved):

Christiantyemele (approved):

PR #56, opened 2025-04-24
deploy server

Ogenbertrand:
setup the helm chart to deploy are server to aws eks, please check and lemme know your thoughts.


helm/status-list-server-chart/charts/postgresql-16.6.4.tgz
Awambeng, line 1: You can ignore the compress template

helm/status-list-server-chart/values.yaml
```diff
@@ -0,0 +1,67 @@
+statuslist:
+  replicaCount: 1
+  image:
+    repository: ghcr.io/adorsys/status-list-server
+    pullPolicy: IfNotPresent
+    tag: "1.0.0"
```
Awambeng, line 6: Please update the tag to `latest` for now, so we always fetch the most recent version of the StatusList server.

helm/status-list-server-chart/templates/ingress.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "status-list-server-chart.fullname" . }}-ingress
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: {{ .Values.statuslist.externalDnsHostname | default "statuslist.eudi-adorsys.com" | quote }}
```
Awambeng, line 7: Please remove the default value as it is not necessary

helm/status-list-server-chart/templates/ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- toYaml .Values.statuslist.ingress.annotations | nindent 4 }}
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: {{ .Values.statuslist.externalDnsHostname | default "statuslist.eudi-adorsys.com" | quote }}
```
Awambeng, line 12: Same here. Remove the default value

helm/status-list-server-chart/templates/ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+    http:
+      paths:
+      - path: /
+        pathType: Prefix
```
Awambeng, line 16: Consider referencing the values defined in `values.yaml` instead of hardcoding them directly.

helm/status-list-server-chart/templates/external-secretes.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: {{ .Values.externalSecret.metadata.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  refreshInterval: 30m
```
Awambeng, line 7: Consider referencing the values defined in values.yaml instead of hardcoding them directly.

helm/status-list-server-chart/templates/external-secretes.yaml
```diff
@@ -0,0 +1,18 @@
...
+  secretStoreRef:
+    name: {{ .Values.externalSecret.spec.secretStoreRef.name }}
+    kind: SecretStore
```
Awambeng, line 10: Consider referencing the values defined in values.yaml instead of hardcoding them directly.

src/main.rs
```diff
@@ -33,7 +33,7 @@ async fn main() {
         .allow_headers(Any);
 
     let router = Router::new()
-        .route("/", get(welcome))
+        .route("/health", get(welcome))
```
Awambeng, line 36: Replacing the root route with `/health` isn’t ideal, as it removes the original welcome endpoint. Instead, please add a new `/health` route separately to serve the health check, while keeping the existing root route intact

Awambeng (changes requested):

helm/status-list-server-chart/templates/secretes-store.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
+kind: SecretStore
+metadata:
+  name: {{ .Values.externalSecret.spec.secretStoreRef.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  provider:
+    aws:
+      service: SecretsManager
+      region: eu-central-1
```
forkimenjeckayang, line 10: Wouldn't it be good too to send to the values.yml and reference here for easy maintainability?
```yml
region: {{ .Values.externalSecret.spec.provider.aws.region }}
```

forkimenjeckayang (changes requested):

helm/status-list-server-chart/templates/secretes-store.yaml
Ogenbertrand: yes thanks for pointing this out.

src/main.rs
```diff
@@ -20,6 +20,10 @@ async fn welcome() -> impl IntoResponse {
     "Status list Server"
 }
 
+async fn health_check() -> &'static str {
+    "OK"
+}
```
Awambeng, line 25: Could we adjust the health check to return a JSON response instead of plain text?
This would make it easier to extend later if we want to add more health information (like database or service status).

What do you think about making this change now?

Awambeng (changes requested):

src/main.rs
Ogenbertrand: > Could we adjust the health check to return a JSON response instead of plain text? This would make it easier to extend later if we want to add more health information (like database or service status).
> 
> What do you think about making this change now?

Switching the health_check function to return a JSON response instead of plain text is a great idea, especially if we anticipate extending the health check to include additional information like database or service status in the future.

Awambeng (approved):

forkimenjeckayang (approved):

Issue #45, opened 2025-04-22
Add Keypair Generation to Startup Logic


Christiantyemele:
This task involves performing validation checks on the server's keys and generating a new keypair if none exists or if the current keys are invalid.

Ngha-Boris:
Link to PR https://github.com/adorsys/status-list-server/pull/55

PR #65, opened 2025-04-30
generate and parse certificate from ca

Hermann-Core:
#### Server-Based Certificate Management (that this PR tries to implement)

**Limitations:**
- **Let’s Encrypt** does not support custom EKUs as required by the protocol.
- Each service instance must manage its own certificates, which can be cumbersome if we scale to multiple services.
- TLS processing occurs in the application server, potentially impacting performance under high traffic.
- Each server needs to handle certificate renewal and challenge responses.

#### Reverse Proxy-Based Certificate Management

- Certificates for multiple services are managed in one place, simplifying updates and renewals.
- The reverse proxy handles encryption/decryption, reducing the load on the application server.
- Enables load balancing, caching, rate limiting, and other proxy capabilities.
- Tools like Certbot can automate Let’s Encrypt certificate management, widely supported for reverse proxies.

Considering these, would not be a more effective approach to let the reverse proxy handle the certificate management and TLS processing?


IngridPuppet:
> Considering these, would not be a more effective approach to let the reverse proxy handle the certificate management and TLS processing?

I wholly agree with the idea of not embedding the logic for X509 certificate retrieval and renewal into the server. There are much stronger solutions available, only requiring the app to consume it. 

However, I do not think reusing the certificate managed at the level of the Reverse Proxy is the way to go for any thing other than TLS communication. For instance, the Status List Server mainly needs to expose an X5C chain so the tokens it issues may be verified by Relying Parties. I believe we should rather rely on a different mechanism for provisioning the certificate, something similar to Hashicorp Vault or AWS Secrets for secrets management.

That said, given the tight link between this certificate and the keypair used by the server for signing, I would slightly disagree with the keypair generation handled as part of the server lifecycle in:
- #45.

Hermann-Core:
Yes for signing JWTs or CWTs for status list tokens, it seems impractical to reuse the certificates managed by the reverse proxy or external certificate management tools for the following reasons:
- Reverse proxies or external certificate managers typically do not expose the private key used for TLS termination.
- TLS certificates don't not include required fields like the Extended Key Usage (EKU) extension specifying custom OIDs (for _**oauthStatusListSigning**_ as per the spec).
- Certificates provisioned by automated systems are not easily injectable or modifiable at runtime to meet protocol-specific constraints.
- Public certificate authorities generally do not support arbitrary EKUs or custom OIDs, which are required by the spec.

One practical and scalable solution is to effectively offload SSL/TLS termination to a dedicated reverse proxy service. For token signing, we could generate a dedicated signing certificate/key pair that is fully under our control (what is actually being done), include the appropriate EKU extension (_**id-kp-oauthStatusListSigning**_) then include the certificate chain used for token signing directly in the token header (**x5c** parameter) to allow relying parties to validate the signature chain according to the spec.

IngridPuppet:
> we could generate a dedicated signing certificate/key pair that is fully under our control (what is actually being done)

Do you mean letting the server itself generate a self-signed certificate? Could you be more specific? I'm afraid that won't work because of the spec constraints on the Root CA.

Hermann-Core:
> Do you mean letting the server itself generate a self-signed certificate? Could you be more specific? I'm afraid that won't work because of the spec constraints on the Root CA.

No, that's not what I meant. The idea is that the server itself will be responsible for generating a Certificate Signing Request (CSR), submitting it to a Certificate Authority, and then retrieving and managing the resulting certificate. This includes handling certificate renewal as needed, ensuring the server maintains a valid and trusted certificate over time.

IngridPuppet:
> No, that's not what I meant. The idea is that the server itself will be responsible for generating a Certificate Signing Request (CSR), submitting it to a Certificate Authority, and then retrieving and managing the resulting certificate. This includes handling certificate renewal as needed, ensuring the server maintains a valid and trusted certificate over time.

Thank you for the clarification. So we are basically back to the initial description of the ticket... maybe with the precision that this is solely for token signing. I would like to reiterate that I do not believe requesting certificates should be the burden of the server.

However, I do not disagree with this precised direction given to the ticket. It is okay for me, we could still get back to it if things do not go to plan.

Christiantyemele:
> Yes for signing JWTs or CWTs for status list tokens, it seems impractical to reuse the certificates managed by the reverse proxy or external certificate management tools for the following reasons:
> 
> * Reverse proxies or external certificate managers typically do not expose the private key used for TLS termination.
> * TLS certificates don't not include required fields like the Extended Key Usage (EKU) extension specifying custom OIDs (for _**oauthStatusListSigning**_ as per the spec).
> * Certificates provisioned by automated systems are not easily injectable or modifiable at runtime to meet protocol-specific constraints.
> * Public certificate authorities generally do not support arbitrary EKUs or custom OIDs, which are required by the spec.
> 
> One practical and scalable solution is to effectively offload SSL/TLS termination to a dedicated reverse proxy service. For token signing, we could generate a dedicated signing certificate/key pair that is fully under our control (what is actually being done), include the appropriate EKU extension (_**id-kp-oauthStatusListSigning**_) then include the certificate chain used for token signing directly in the token header (**x5c** parameter) to allow relying parties to validate the signature chain according to the spec.

@Hermann-Core so you plan adding the certificate to jwt/cwt to the x5c response header as from the spec right, if i understood this point?

Ngha-Boris (approved):

PR #62, opened 2025-04-28
feat: bind status list to their issuer

Ngha-Boris:

PR #68, opened 2025-05-05
feat(): seperated update handler ffrom publishing handler

Christiantyemele:
@IngridPuppet you can check this PR and update your demo workflow


src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod auth;
 mod database;
 pub mod model;
+pub mod test_resources;
```
IngridPuppet, line 4: What is the need for exposing test resources?

src/test_resources/helper.rs
```diff
@@ -0,0 +1,25 @@
...
+        ttl: Some(3600),
+        bits,
+    }
+}
+
+// Helper to generate a test server key
+// Note: It does nothing, it's just use to build the AppState
+pub fn server_key() -> Keypair {
+    Keypair::generate().unwrap()
+}
```
IngridPuppet, line 25: These are marked as unused. Did you mean to remove them?

src/lib.rs
IngridPuppet, line 4: What is the need for exposing test resources?

src/test_resources/helper.rs
IngridPuppet, line 25: These are marked as unused. Did you mean to remove them?

src/utils/lst_gen.rs
```diff
@@ -72,26 +92,23 @@ pub fn update_or_create_status_list(
...
-            // Create a mask to isolate the target bit segment within the byte
             let mask = ((1 << bits) - 1) << bit_offset;
-
-            // Clear the bits at the target location using bitwise AND with the inverted mask
             status_array[byte_index] &= !mask;
-
-            // Shift the status value into the correct position, then set the bits using bitwise OR
             status_array[byte_index] |= (status_value << bit_offset) & mask;
         }
     }
```
IngridPuppet, line 99: What is your motivation for removing these comments?

src/web/handlers/status_list/update_token_status.rs
```diff
@@ -0,0 +1,168 @@
...
+};
+
+use super::error::StatusListError;
+
+// Handler to update an existing status list token
+pub async fn update_token_status(
+    State(appstate): State<AppState>,
+    Json(payload): Json<StatusListTokenPayload>,
+) -> Result<impl IntoResponse, StatusListError> {
+    let store = &appstate.status_list_token_repository;
```
IngridPuppet, line 18: This route does not seem to enforce that status lists are only updated by their publishers, which is required.

src/test_resources/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod helper;
```
IngridPuppet, line 1:
```suggestion
#[cfg(test)]
pub mod helper;
```

IngridPuppet (changes requested):

src/web/handlers/status_list/update_token_status.rs
Christiantyemele: Ok, thanks @IngridPuppet. This is a major concern you highlighted. I have a suggestion to solve this: we could leverage the middleware such that after authentication, the issuer is passed to the handlers. Then we can use the list_id to fetch the status_list tokens as before, but now this structure should contain the extra issuer field. Then we compare this field with that passed by the middleware. Here we are sure the issuer wanting to update the token is the owner of the token. This is my take on the solution. What do you think, @Hermann-Core?

Hermann-Core: Or simply implement a middleware to extract the `Authorization` header, validate the bearer token data against the registered credentials, then inject the claims (or more precisely the `iss` field) into the request extensions for use in handlers.

Christiantyemele: @Hermann-Core, please bear with me. I will implement the database changes here instead of waiting for the separate ticket to be completed.

IngridPuppet: @Christiantyemele - Could you check with @Ngha-Boris if that's related to a PR of his I once reviewed? I fear you might be repeated the same database updates.

Christiantyemele: yh i will sync with him

src/web/handlers/status_list/handler.rs
```diff
@@ -64,6 +67,7 @@ impl StatusListTokenExt for StatusListToken {
 pub async fn get_status_list(
     State(state): State<AppState>,
     Path(list_id): Path<String>,
+    Extension(issuer): Extension<String>,
     headers: HeaderMap,
 ) -> Result<impl IntoResponse + Debug, StatusListError> {
```
IngridPuppet, line 72: Are you protecting this endpoint in any way, limiting it to a specific issuer? You must not.

src/web/handlers/status_list/update_token_status.rs
```diff
@@ -25,37 +26,45 @@ pub async fn update_token_status(
     })?;
 
     // Fetch the existing token
-    let mut token = match store.find_one_by(payload.list_id.clone()).await {
-        Ok(Some(token)) => token,
-        Ok(None) => return Err(StatusListError::StatusListNotFound),
+    let token = match store.find_all_by(issuer).await {
+        Ok(tokens) => tokens,
```
IngridPuppet, line 30: - A `find_all` ends up just being more expensive than a `find_one`, even in this situation where you want filter a specific issuer's lists.
- Second, you subsequently drown all negative paths into a 404 Not Found. You should make the 403 case explicit.

src/web/handlers/status_list/update_token_status.rs
```diff
@@ -25,37 +26,45 @@ pub async fn update_token_status(
...
+            Error::InvalidIndex => StatusListError::InvalidIndex,
+            Error::UnsupportedBits => StatusListError::UnsupportedBits,
+            _ => StatusListError::Generic(e.to_string()),
+        }
+    })?;
+
+    let mut exact_token = exact_token;
+    exact_token.status_list.lst = updated_lst;
+    exact_token.status_list.bits = payload.bits;
+    exact_token.status_list.bits = payload.bits;
```
IngridPuppet, line 62:
```suggestion
    exact_token.status_list.bits = payload.bits;
```

src/web/handlers/status_list/publish_token_status.rs
```diff
@@ -9,13 +9,15 @@ use axum::{
...
 };
 use std::time::{SystemTime, UNIX_EPOCH};
 use tracing;
 
 // Handler to create a new status list token
 pub async fn publish_token_status(
     State(appstate): State<AppState>,
+    Extension(issuer): Extension<String>,
     Json(payload): Json<StatusListTokenPayload>,
 ) -> Result<impl IntoResponse, StatusListError> {
```
IngridPuppet, line 22: 1. Could you explain how the issuer is provisioned to this handler? I'm not seeing a corresponding update at the middleware layer.
2. Did you test this live (e.g. with Postman)? If not, you should.

src/main.rs
```diff
@@ -25,15 +24,8 @@ async fn welcome() -> impl IntoResponse {
...
-struct HealthCheckResponse {
-    status: String,
-}
-
 async fn health_check() -> impl IntoResponse {
-    Json(HealthCheckResponse {
-        status: "OK".to_string(),
-    })
+    "Server is Running"
 }
```
IngridPuppet, line 29: Is this a merge mismanipulation? Otherwise, I neither get your motivation nor see how it is related to this ticket.

IngridPuppet (changes requested):
Hello @Christiantyemele - I added some more comments. Please could you check?

src/model.rs
```diff
@@ -132,6 +132,7 @@ pub mod status_list_tokens {
     pub struct Model {
         #[sea_orm(primary_key)]
         pub list_id: String,
+        pub issuer: String,
         pub exp: Option<i64>,
```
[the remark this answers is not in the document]
Christiantyemele: > Additionally, this table should be linked to the credentials table by a many-to-one relationship. Please @Ngha-Boris - could you remind us in which PR you did this exact thing? We should avoid unnecessary conflicts.

@IngridPuppet i think the table linking will be done in another ticket, idk if @Ngha-Boris is working on that now

src/web/handlers/status_list/handler.rs
Christiantyemele: yes this is true, this is not the endpoint to protect

src/web/handlers/status_list/handler.rs
```diff
@@ -88,17 +92,22 @@ pub async fn get_status_list(
...
         .map_err(|err| {
             tracing::error!("Failed to get status list {list_id} from database: {err:?}");
             StatusListError::InternalServerError
-        })?
+        })?;
+
+    let status_claims = status_claims
+        .iter()
+        .find(|token| token.list_id == list_id)
         .ok_or(StatusListError::StatusListNotFound)?;
```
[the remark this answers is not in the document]
Christiantyemele: okay :+1:

src/model.rs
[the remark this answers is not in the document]
Ngha-Boris: > Additionally, this table should be linked to the credentials table by a many-to-one relationship. Please @Ngha-Boris - could you remind us in which PR you did this exact thing? We should avoid unnecessary conflicts.

Hello @IngridPuppet that was done on this [PR](https://github.com/adorsys/status-list-server/pull/62) but the merge was reverted because of a comment by @Christiantyemele. So I am suggesting a ticket should be created for that or @Christiantyemele can just continue with what he is doing.

src/main.rs
Christiantyemele: Before i proceed with this i will like to take @Hermann-Core point of view

src/web/handlers/status_list/update_token_status.rs
Christiantyemele: @IngridPuppet the main idea and cause of me doing this operation with a `find_all` is that currently we have no way of ensuring the `list_id` are unique, though it's a constraint in our database but the spec always references this id(s) as intergers and not `UUID` making us not certain this id will be unique across the multiple issuers we might have.

src/main.rs
Hermann-Core: After some reflection on the design, I come up with the following discussion:

#### 1. **Status List Identifier (`list_id`) – Integer vs. String**

**Integer-based IDs:**
- **Pros:** Simplifies internal storage and indexing.
- **Cons:** Since multiple issuers might independently use the same integer ID, relying parties could encounter collisions when fetching status lists, leading to ambiguity.

**String-based IDs:**
- **Pros:** Allows for globally unique identifiers, especially when using UUIDs, reducing the risk of collisions across different issuers.
- **Cons:** Introduces potential security concerns, such as injection attacks or resource exhaustion, if not properly validated.

A solid approach is to adopt string-based `list_id`s using either UUIDs to ensure global uniqueness or random string with strict validation to like limit length, and reject any characters that could pose security risks (e.g., `/`, `()`, `<`, `>`, `\`, `{`, `}`, etc.).
Enforcing the use of UUIDs is the best approach for me.

That way, `id`s of status list entries are unique and a single issuer could safely maintain many status informations.

src/web/handlers/status_list/update_token_status.rs
IngridPuppet: Considering your current judgment, it remains preferable, and I therefore suggest so, to instead `find_all` by `list_id`, rather than by `issuer`.

Now, coming to your concerns, I'm leaving what follows as a contribution to the points you raised, not implying changes that I would see on this PR.

> currently we have no way of ensuring the list_id are unique, though it's a constraint in our database

Setting the unique constraint IS a way to ensure they are unique, meaning duplicates are identified at insertion time and should not be a concern at retrieval time.

> but the spec always references this id(s) as intergers and not UUID making us not certain this id will be unique across the multiple issuers we might have.

It is not the first time I'm seeing this concern, and my position hasn't changed. I don't mind having IDs as integers or UUIDs. There are pros and cons in each, and I globally believe you're just overthinking it.
- Currently, you believe UUIDs are what makes "us not certain this id will be unique across the multiple issuers we might have". Integers are worse in this regard. The problem is not the format, but the fact that the server is expecting these IDs from issuers, while it should ideally generate them itself to prevent the occurrence of duplicates. 
- You also seem concerned that UUIDs will be a source of duplicate list IDs in our current situation. Provided that issuers abide by the UUID expectation, your fear would be generally considered unjustified. UUIDs are designed to be universally unique, that's what they stand for: Universally Unique Identifiers. The possibility of a duplicate generation, even across issuers, is so low that your fear is indeed unjustified.

src/main.rs
IngridPuppet: Hello @Hermann-Core - It looks like you wanted to post this comment under the other conversation above... Anyway, I have no objection with either choice, integer or string. There would have also been sound ways to solve the issue you mention regarding Integer-based IDs; it is just overamplified here.

Now, on the point of this conversation, do you commend updating the implementation of this health check endpoint? I would like to re-emphasize that even if you do, this PR was initially not about the health check endpoint and unrelated changes are a pain for PR reviewers.

src/web/handlers/status_list/update_token_status.rs
Christiantyemele: No, you did'nt get my point i fully understand `UUID` are unique even across issuers but my concern is the spec does not uses `UUIDs` but `intergers`, should i break this point of the specification and use `UUIDs` ? if yes then all my issues will be solve but if no i have not way to ensure integers id are unique except i do something like what you suggest `issuer-{interger-id}` then i might have something unique otherwise i have no other way than doing it as i am currently doing

src/main.rs
Christiantyemele: Then i will assume the id are `UUID` and go with that if we are all okay though it's spec breaking

i will undo the health check endpoint changes @IngridPuppet

src/web/handlers/status_list/update_token_status.rs
IngridPuppet: Oh I forgot to add that in my write up.

> my concern is the spec does not uses UUIDs but intergers, should i break this point of the specification and use UUIDs

The spec using integers does not mean that it requires them and that you would break it by using IDs. Pay attention to the word "non-normative" when examples are given.

> i have not way to ensure integers id are unique except i do something like what you suggest issuer-{interger-id} then i might have something unique otherwise i have no other way than doing it as i am currently doing

I never suggested this format, but it is not bad -- you just end up with something that's no longer an integer. It is not impossible to have unique identifiers if you want to be integers. Just let the database generate them for you.

src/main.rs
IngridPuppet: > though it's spec breaking
Rest assured it is not :-) If you don't have an explicit statement saying for example, List identifiers MUST be integers, then it is not a requirement. Examples in spec just aim to be simplistic when it does not matter. They don't aim to communicate requirements.

Hermann-Core: That said, consider adding a validation check to the `id`s to ensure they are valid uuid before proceeding or directly pass UUID as path extractor and `list_id` type if it provides some serialization mechanism and let Axum doing the work for us.

src/web/handlers/status_list/error.rs
```diff
@@ -45,6 +47,7 @@ impl IntoResponse for StatusListError {
             DecompressionError(_) => StatusCode::BAD_REQUEST,
             CompressionError(_) => StatusCode::INTERNAL_SERVER_ERROR,
             StatusListAlreadyExists => StatusCode::CONFLICT,
+            IssuerMismatch => StatusCode::BAD_REQUEST,
```
IngridPuppet, line 50: 403 is more appropriate in this situation.
```suggestion
            IssuerMismatch => StatusCode::FORBIDDEN,
```

src/web/handlers/status_list/update_token_status.rs
```diff
@@ -0,0 +1,187 @@
...
+                    Error::UnsupportedBits => StatusListError::UnsupportedBits,
+                    _ => StatusListError::Generic(e.to_string()),
+                }
+            },
+        )?;
+
+    let mut exact_token = token;
+    exact_token.status_list.lst = updated_lst;
+    exact_token.status_list.bits = payload.bits;
+    exact_token.status_list.bits = payload.bits;
```
IngridPuppet, line 67: This instruction appears twice.
```suggestion
    exact_token.status_list.bits = payload.bits;
```

src/web/handlers/status_list/update_token_status.rs
```diff
@@ -26,37 +26,42 @@ pub async fn update_token_status(
...
+                    Error::UnsupportedBits => StatusListError::UnsupportedBits,
+                    _ => StatusListError::Generic(e.to_string()),
+                }
+            },
+        )?;
+
+    let mut exact_token = token;
     exact_token.status_list.lst = updated_lst;
     exact_token.status_list.bits = payload.bits;
     exact_token.status_list.bits = payload.bits;
```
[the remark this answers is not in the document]
Christiantyemele: oh i though i resolved this

IngridPuppet (approved):

src/model.rs
```diff
@@ -159,10 +160,21 @@ pub enum Status {
...
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, FromJsonQueryResult)]
 pub struct StatusList {
-    pub bits: usize,
+    pub bits: u8,
     pub lst: String,
 }
 
+/// Request payload for publishing and updating a status list token
+#[derive(Deserialize, Serialize, Clone)]
+pub struct StatusListTokenPayload {
```
Hermann-Core, line 169: Could you use a more suitable name? That could be used for both publishing and updating.
```suggestion
pub struct StatusRequest {
```

src/utils/lst_gen.rs
```diff
@@ -17,49 +16,70 @@ pub fn update_or_create_status_list(
...
-            .ok_or_else(|| Error::Generic("Failed to determine max index".to_string()))?;
-
-        if max_index < 0 {
-            return Err(Error::InvalidIndex);
-        }
+    let max_index = status_updates
+        .iter()
+        .map(|update| update.index)
+        .max()
+        .ok_or_else(|| Error::Generic("Failed to determine max index".to_string()))?;
```
Hermann-Core, line 23: I don't think returning an error when not finding the max index is appropriate here. You could instead assume the max index is `0`.
```suggestion
        .unwrap_or(0);
```

src/utils/lst_gen.rs
```diff
@@ -17,49 +16,70 @@ pub fn update_or_create_status_list(
...
-            .map(|update| update.index)
-            .max()
-            .ok_or_else(|| Error::Generic("Failed to determine max index".to_string()))?;
-
-        if max_index < 0 {
-            return Err(Error::InvalidIndex);
-        }
+    let max_index = status_updates
+        .iter()
+        .map(|update| update.index)
```
Hermann-Core, line 21: Directly convert it as usize here will unsure negative numbers are catched earlier.
```suggestion
        .map(|update| update.index as usize)
```

src/utils/lst_gen.rs
```diff
@@ -17,49 +16,70 @@ pub fn update_or_create_status_list(
...
-        // Create a new list initialized to VALID
-        let required_len = (max_index as usize + 1) * bits + 7;
-        let len = required_len / 8;
-        vec![0u8; len]
-    };
+    if max_index < 0 {
+        return Err(Error::InvalidIndex);
+    }
+
+    let required_len = (max_index as usize + 1) * bits + 7;
```
Hermann-Core, line 29: Since the max_index will already be of type `usize`, you will not need a cast here.

src/web/handlers/status_list/handler.rs
```diff
@@ -98,8 +101,15 @@ async fn build_status_list_token(
...
             tracing::error!("Failed to get status list {list_id} from database: {err:?}");
             StatusListError::InternalServerError
-        })?
-        .ok_or(StatusListError::StatusListNotFound)?;
+        })?;
+
+    let status_claims = match status_claims {
+        Some(status_claims) => status_claims,
+        None => {
+            tracing::error!("Status list {list_id} not found");
```
Hermann-Core, line 109: I don't think it is necessary to trace this unless you have a good reason for doing this.

src/web/handlers/status_list/update_token_status.rs
```diff
@@ -0,0 +1,186 @@
...
+            tracing::error!("Token not found in the database.");
+            return Err(StatusListError::StatusListNotFound);
+        }
+    };
+
+    // check if the request issuer matches the token issuer
+    if token.issuer != issuer {
+        tracing::error!("Issuer mismatch: expected {}, got {}", token.issuer, issuer);
+        return Err(StatusListError::IssuerMismatch);
+    }
```
Hermann-Core, line 48: The logic here could be simplified by querying the database using both `list_id` and `issuer`. This ensures we directly check whether the issuer is authorized to manage the given status list. If no matching entry is found, returning a `403 Forbidden` is more appropriate, as it clearly indicates that the issuer does not have permission to modify the resource even if it exists. The same logic applies to the `create_status_list` handler to prevent unauthorized creation or duplication under a different issuer.

src/model.rs
```diff
@@ -159,10 +160,21 @@ pub enum Status {
...
+    pub list_id: String,
+    pub status: Vec<StatusEntry>,
+    #[serde(default)]
+    pub sub: Option<String>,
+    #[serde(default)]
+    pub ttl: Option<i64>,
+    pub bits: u8,
```
Hermann-Core, line 176: I don't think these fields are necessary for publishing and updating statuses.

Hermann-Core (changes requested):

src/utils/lst_gen.rs
Christiantyemele: No we can not do this because returning `0` will mean we found a token which is not true cause required len will be `1`

src/utils/lst_gen.rs
Christiantyemele: This is change is actually risky cause passing a negative index will result to an error when trying to cast it into a usize

src/model.rs
Christiantyemele: The are actually needed https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#name-status-list-token

src/web/handlers/status_list/handler.rs
Christiantyemele: okay,i had no special reason just for more context on the situation

Christiantyemele:
@Hermann-Core can review again

src/utils/lst_gen.rs
Hermann-Core: No, it will not. It will lead to incorrect behavior later in the code. A better solution is to validate the value upfront either by rejecting negative values early or by ensuring only positive values are allowed in the status entry.

src/web/handlers/status_list/update_token_status.rs
Hermann-Core: > That also means a 403 will be returned when there is no list of ID `list_id`. Are you okay with that? I find it weird.

That's a valid concern, and I agree it can feel a bit unintuitive at first. The reasoning behind returning a `403 Forbidden` in this case is to avoid leaking information about the existence of a status list the issuer doesn't own. It's generally safer not to reveal whether a resource exists if the requester isn't authorized to access it. Both entries should match before the actual resource is granted. Both list_id and issuer must match before access to the resource is granted.

src/model.rs
Hermann-Core: They actually are not. The status list token is constructed when returning status list information to relying parties.

src/web/handlers/status_list/update_token_status.rs
IngridPuppet: Status lists are open to the world; they aren't private resources. That's only going to create unpleasant debugging experiences.

Hermann-Core: Yes but this is not about retrieving them but updating them. At this level they are protected from unauthorized modifications.

src/model.rs
```diff
@@ -159,10 +160,25 @@ pub enum Status {
...
     pub lst: String,
 }
 
+/// Request payload for publishing a status list token
+#[derive(Deserialize)]
+pub struct PublishStatusRequest {
+    pub list_id: String,
+    pub status: Vec<StatusEntry>,
+    pub sub: String,
```
Hermann-Core, line 172: Why is the `sub` field necessary to process this request? It seems like you could reuse the same request data as in the update operation.

Hermann-Core (changes requested):

src/main.rs
```diff
@@ -48,8 +52,13 @@ async fn main() {
...
         .route("/health", get(health_check))
         .route("/credentials", post(credential_handler))
-        .route("/statuslists/{list_id}", get(get_status_list))
-        .route("/statuslists/publish", post(publish_token_status))
+        .nest(
+            "/statuslists",
+            Router::new()
+                .route("/:list_id", get(get_status_list))
+                .route("/publish", post(publish_token_status))
+                .route("/update", put(update_token_status)),
```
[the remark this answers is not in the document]
Christiantyemele: > > HTTP PUT request is used to replace and update the entire resource or document, while the PATCH request only updates the specific parts of that document.
> 
> Given that, I believe `PATCH` would be more appropriate in this case.

No i find no difference, again these are not changes which should be blocking a ticket

src/web/handlers/status_list/handler.rs
```diff
@@ -98,8 +101,14 @@ async fn build_status_list_token(
...
-        })?
-        .ok_or(StatusListError::StatusListNotFound)?;
+        })?;
+
+    let status_claims = match status_claims {
+        Some(status_claims) => status_claims,
+        None => {
+            return Err(StatusListError::StatusListNotFound);
+        }
+    };
```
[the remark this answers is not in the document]
Christiantyemele: > 

Is this change really neccessary ?

src/model.rs
Christiantyemele: The StatusList token REQUIRES a sub field which is provided by the issuer and should match that in the reference token
Screenshot from 2025-05-21 11-15-55

src/main.rs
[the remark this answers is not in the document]
Hermann-Core: > No i find no difference, again these are not changes which should be blocking a ticket

There is indeed a clear difference, as outlined in the description. Since we're updating only specific parts of the resource (the status list record in the database), `PATCH` is the correct and conventional HTTP method to use. It's important to follow established standards rather than relying on personal preference.

src/web/handlers/status_list/handler.rs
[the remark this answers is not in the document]
Hermann-Core: It is more convenient and less verbose.

Hermann-Core (approved):

PR #70, opened 2025-05-07
Implement gzip content encoding for status list responses

Hermann-Core:


Ngha-Boris (approved):

src/web/handlers/status_list/handler.rs
```diff
@@ -102,18 +103,36 @@ async fn build_status_list_token(
...
+        let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+        encoder.write_all(data).map_err(|err| {
+            tracing::error!("Failed to compress payload: {err:?}");
+            StatusListError::InternalServerError
+        })?;
+        encoder.finish().map_err(|err| {
+            tracing::error!("Failed to finish compression: {err:?}");
+            StatusListError::InternalServerError
+        })
+    };
```
IngridPuppet, line 116: It does not sound sustainable to repeat this logic on every possible future route requiring compression. Did you look into something like `tower_http::compression::CompressionLayer` to have gzip compression applied automatically? Do you have reasons for not going that way nor implementing a custom middleware?

src/web/handlers/status_list/handler.rs
```diff
@@ -102,18 +103,36 @@ async fn build_status_list_token(
...
-            [(header::CONTENT_TYPE, accept)],
-            issue_cwt(&status_claims, &server_key)?,
+            [
+                (header::CONTENT_TYPE, accept),
+                (header::CONTENT_ENCODING, GZIP_HEADER),
+            ],
+            apply_gzip(issue_cwt(&status_claims, &server_key)?.as_slice())?,
         )
             .into_response())
     }
```
IngridPuppet, line 138: Hint: you could reduce code duplication.

```rust
    let gzipped_data = match accept {
        ACCEPT_STATUS_LISTS_HEADER_CWT => {
            apply_gzip(issue_cwt(&status_claims, &server_key)?.as_slice())?
        }
        _ => apply_gzip(issue_jwt(&status_claims, &server_key)?.as_bytes())?,
    };

    Ok((
        StatusCode::OK,
        [
            (header::CONTENT_TYPE, accept),
            (header::CONTENT_ENCODING, GZIP_HEADER),
        ],
        gzipped_data,
    )
        .into_response())
 ```

IngridPuppet (commented):
I added some comments. Could you check?

src/web/handlers/status_list/handler.rs
Hermann-Core: I have considered it as well, but this middleware will apply compression globally to responses, which I believe is not the desired behavior.

src/web/handlers/status_list/handler.rs
Hermann-Core: Thank for suggestion. I had not realized the return type is the same now, which is an opportunity to improve the logic.

IngridPuppet (approved):
Even better.

PR #72, opened 2025-05-12
refactor authentication and credentials issuance endpoint

Hermann-Core:


src/auth/authentication.rs
```diff
@@ -3,65 +3,64 @@ use serde_json::Value;
...
-
-        match decode::<Value>(token, &decoding_key, &validation) {
-            Ok(_) => Ok(true),
-            Err(e) => {
-                tracing::warn!("Token decoding failed: {}", e);
-                Ok(false)
-            }
+    let decoding_key = match credential.alg {
+        Algorithm::HS256 | Algorithm::HS384 | Algorithm::HS512 => {
+            Ok(DecodingKey::from_secret(credential.public_key.as_bytes()))
```
IngridPuppet, line 46: I think it is preferable these symmetric algorithms are not supported, or at least, the property `public_key` should be renamed to something else rather than being treated as a secret.

src/auth/authentication.rs
```diff
@@ -3,65 +3,64 @@ use serde_json::Value;
...
+        }
+        Algorithm::EdDSA => DecodingKey::from_ed_pem(credential.public_key.as_bytes()),
+    }?;
+
+    let mut validation = Validation::new(credential.alg);
+    validation.set_issuer(&[credential.issuer]);
+
+    decode::<Value>(token, &decoding_key, &validation)
+        .map(|_| true)
+        .map_err(AuthenticationError::JwtError)
```
IngridPuppet, line 65: It seems the method is no longer intended to return an `Ok(false)`. Could you update the return type to the union type?

src/auth/authentication.rs
Hermann-Core: These are decoding keys and therefore public. They are not secrets although the API naming might not be ideal.

src/auth/authentication.rs
Hermann-Core: Yes thanks for catching this.

src/auth/authentication.rs
IngridPuppet: > These are decoding keys and therefore public.

As a reminder, in symmetric algorithms, the encoding key doubles as the decoding key. So here, the decoding key is indeed a secret and cannot be called public.

IngridPuppet: Yes, agree. In my experience, it is even common to merely not support these algorithms in contexts they are not deemed secure enough.

Hermann-Core: @IngridPuppet Removed support for them. Could you check again?

IngridPuppet (approved):

PR #82, opened 2025-06-12
388 improve credential retrieval in the auth middleware

Hermann-Core:
I have refactored the authentication middleware to use a more reliable way to authenticate issuers. Could you check the PR @IngridPuppet and @Ngha-Boris ?


IngridPuppet (approved):
Upon reviewing this PR, here are the main points I noted:

- Authentication is now handled idiomatically with the axum middleware system, purely.
- Bearer tokens are no longer linked to token issuers by their header `kid` but their payload `iss` claim.

I agree with these changes and have no concerns to raise.

IngridPuppet (approved):

PR #98, opened 2025-07-22
feat: database connection issue in status-list-server

Blindspot22:


src/utils/cert_manager/storage/aws.rs
```diff
@@ -173,13 +179,20 @@ impl AwsS3 {
...
                 Err(SdkError::ServiceError(err)) if err.err().is_not_found() => {
                     // Bucket not found, attempt to create it
-                    match self
-                        .client
-                        .create_bucket()
-                        .bucket(&self.bucket)
-                        .send()
-                        .await
-                    {
+                    use aws_sdk_s3::types::CreateBucketConfiguration;
```
Hermann-Core, line 182: The imports should be placed at the top of the file.

src/utils/cert_manager/storage/aws.rs
```diff
@@ -173,13 +179,20 @@ impl AwsS3 {
...
+                    let mut req = self.client.create_bucket().bucket(&self.bucket);
+                    if self.region != "us-east-1" {
+                        req = req.create_bucket_configuration(
+                            CreateBucketConfiguration::builder()
+                                .location_constraint(
+                                    self.region
+                                        .parse()
+                                        .expect("Invalid region for LocationConstraint"),
```
Hermann-Core, line 190: Consider handle errors gracefully instead of panicking.

src/utils/state.rs
Hermann-Core, line 1: The server is intended to be stateless and should not write anything to the filesystem. TLS can be enabled without requiring certificates unless mutual TLS (client authentication) is needed, which is not the case for us right now. I misconfigured this initially.
The `require_tls` flag should be renamed to `require_client_auth`, since enabling TLS doesn’t require client certs. Please revert to the previous logic, rename the variable accordingly (`require_client_auth`), and default it to `false`. Also remove the `APP_REDIS__ROOT_CERT_PATH` variable and the `root_cert_path` field from the config.
Don't forget to adjust the tests to reflect this change.

Hermann-Core (commented):
Hello @forkimenjeckayang could you check the comment I left?

forkimenjeckayang:
Thank you for the feedback! @Hermann-Core 
I've implemented the changes as requested, but with a slight modification to better align with the Redis server configuration.

## Changes Made:

**Removed filesystem dependencies:**
- Removed `APP_REDIS__ROOT_CERT_PATH` environment variable
- Removed `root_cert_path` field from `RedisConfig`
- Eliminated all filesystem certificate loading logic

**Kept `require_tls` for general TLS support:**
- The application now supports TLS without requiring client certificates
- When `require_tls = true`, connects with TLS using system CA certificates
- When `require_tls = false`, uses plain text connection

**Added `require_client_auth` for mutual TLS:**
- Defaults to `false` as requested
- Only used when mutual TLS (client authentication) is needed
- Aligns with Redis server config where `authClients: false`

## Configuration Alignment:

**Redis Server (Bitnami):**
```yaml
redis:
  tls:
    enabled: true          # TLS enabled
    authClients: false     # No mutual TLS
```

**Application:**
```yaml
APP_REDIS__REQUIRE_TLS: "true"        # TLS enabled
APP_REDIS__REQUIRE_CLIENT_AUTH: "false" # No mutual TLS (default)
```

## Connection Logic:
```rust
if !self.require_tls {
    // Plain text connection
} else if self.require_client_auth {
    // Mutual TLS (client certificates required)
} else {
    // TLS without client authentication (server-only TLS)
}
```

This approach:
- ✅ Keeps the application stateless (no filesystem writes)
- ✅ Supports TLS without client certificates
- ✅ Maintains flexibility for future mutual TLS if needed
- ✅ Aligns with the Redis server configuration

Please check again and let me know

IngridPuppet (approved):

PR #115, opened 2025-11-10
feat(auth): validate public key format on credential issuance

Hermann-Core:
Currently, when registering a new credential in the database, the provided public key is stored as-is without any format or validity checks.  
This can cause failures later when trying to verify signatures using the stored key if the key is malformed, incorrectly encoded, or uses an unsupported algorithm.


Ngha-Boris:
Thank you @Hermann-Core for this PR it is really important.

ndefokou (approved):

PR #122, opened 2025-11-13
122 move all documents to a doc folder

ndefokou:


ndefokou:
@Ogenbertrand check this for me i was forced to merge the main in to my branch locally

Hermann-Core (commented):
I’ve reviewed the ticket, grouping all documentation-related files into a single directory is a good idea.
I noticed that the file `arc.png` is still present in the root directory. Since it’s a documentation asset, it would be better to move it either into a dedicated `assets` folder or into the existing `docs` folder.

Hermann-Core (approved):

PR #134, opened 2026-05-13
refactor: implement configurable TTL and disablement for all caching layers

Awambeng:
This PR addresses several caching-related issues by ensuring rotated signing keys and certificates are picked up immediately through configurable cache durations. It also introduces a “zero-to-disable” option for development environments.

### Unified Cache Configuration
Introduced configurable TTLs for all three cache layers:
- `APP_CACHE__TTL`: Status List data (`moka` cache)
- `APP_REDIS__CERT_CACHE_TTL`: Certificate chain cache (Redis)
- `APP_REDIS__SECRETS_CACHE_TTL`: Signing keys and ACME credentials (AWS Secrets Manager)

### "Zero-to-Disable" Support
All cache layers now interpret a TTL of `0` as an instruction to bypass caching entirely. This ensures data freshness in development environments (e.g., when testing rotation with Localstack) without waiting for cache expiration.

### Fixed Stale Signing Key / Certificate Issues
Corrected logic where:
- Redis cached certificates indefinitely
- AWS Secrets Manager cached signing keys using a hard-coded 5-minute TTL

### Doc-test Fix
Re-exported `BitFlag` in the crate root to resolve pre-existing visibility issues that caused doc-tests to fail.

Related to https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/31

### DIFF
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -5,16 +5,74 @@ use crate::models::StatusListRecord;
 
 #[derive(Clone)]
 pub struct Cache {
-    pub status_list_record_cache: MokaCache<String, StatusListRecord>,
+    inner: Option<MokaCache<String, StatusListRecord>>,
 }
 
 impl Cache {
-    pub fn new(ttl: u64, max_capacity: u64) -> Self {
-        Self {
-            status_list_record_cache: MokaCache::builder()
-                .time_to_live(Duration::from_secs(ttl))
-                .max_capacity(max_capacity)
-                .build(),
+    /// Creates a cache; TTL=0 disables it.
+    pub fn new(ttl_secs: u64, max_capacity: u64) -> Self {
+        let inner = if ttl_secs == 0 {
+            tracing::info!("Cache disabled (TTL=0)");
+            None
+        } else {
+            Some(
+                MokaCache::builder()
+                    .time_to_live(Duration::from_secs(ttl_secs))
+                    .max_capacity(max_capacity)
+                    .build(),
+            )
+        };
+        Self { inner }
+    }
+
+    pub async fn get(&self, key: &str) -> Option<StatusListRecord> {
+        self.inner.as_ref()?.get(key).await
+    }
+
+    pub async fn insert(&self, key: String, value: StatusListRecord) {
+        if let Some(cache) = &self.inner {
+            cache.insert(key, value).await;
         }
+    }
+
+    pub async fn invalidate(&self, key: &str) {
+        if let Some(cache) = &self.inner {
+            cache.invalidate(key).await;
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::models::{StatusList, StatusListRecord};
+
+    fn sample_record(id: &str) -> StatusListRecord {
+        StatusListRecord {
+            list_id: id.to_string(),
+            issuer: "issuer".to_string(),
+            sub: "sub".to_string(),
+            status_list: StatusList {
+                bits: 1,
+                lst: "test".to_string(),
+            },
+        }
+    }
+
+    #[tokio::test]
+    async fn cache_enabled_works() {
+        let cache = Cache::new(60, 10);
+        let record = sample_record("list-1");
+        cache.insert("list-1".to_string(), record.clone()).await;
+        assert!(cache.get("list-1").await.is_some());
+    }
+
+    #[tokio::test]
+    async fn cache_disabled_returns_none() {
+        let cache = Cache::new(0, 10);
+        cache
+            .insert("list-1".to_string(), sample_record("list-1"))
+            .await;
+        assert!(cache.get("list-1").await.is_none());
     }
 }
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -5,40 +5,32 @@ use crate::models::StatusListRecord;
 
 #[derive(Clone)]
 pub struct Cache {
-    inner: Option<MokaCache<String, StatusListRecord>>,
+    inner: MokaCache<String, StatusListRecord>,
 }
 
 impl Cache {
-    /// Creates a cache; TTL=0 disables it.
+    /// Creates a cache; TTL=0 disables it naturally.
     pub fn new(ttl_secs: u64, max_capacity: u64) -> Self {
-        let inner = (ttl_secs > 0).then(|| {
-            MokaCache::builder()
-                .time_to_live(Duration::from_secs(ttl_secs))
-                .max_capacity(max_capacity)
-                .build()
-        });
-
-        if inner.is_none() {
+        if ttl_secs == 0 {
             tracing::info!("Cache disabled (TTL=0)");
         }
-
+        let inner = MokaCache::builder()
+            .time_to_live(Duration::from_secs(ttl_secs))
+            .max_capacity(max_capacity)
+            .build();
         Self { inner }
     }
 
     pub async fn get(&self, key: &str) -> Option<StatusListRecord> {
-        self.inner.as_ref()?.get(key).await
+        self.inner.get(key).await
     }
 
     pub async fn insert(&self, key: String, value: StatusListRecord) {
-        if let Some(cache) = &self.inner {
-            cache.insert(key, value).await;
-        }
+        self.inner.insert(key, value).await;
     }
 
     pub async fn invalidate(&self, key: &str) {
-        if let Some(cache) = &self.inner {
-            cache.invalidate(key).await;
-        }
+        self.inner.invalidate(key).await;
     }
 }
 
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -11,17 +11,17 @@ pub struct Cache {
 impl Cache {
     /// Creates a cache; TTL=0 disables it.
     pub fn new(ttl_secs: u64, max_capacity: u64) -> Self {
-        let inner = if ttl_secs == 0 {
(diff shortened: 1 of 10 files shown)


Christiantyemele (approved):
## ✅ APPROVED: Environment-Aware Caching for AWS Secrets Manager

**Security-Focused Review**

### Summary
This PR fixes a critical "Split-Brain" issue where the server could sign tokens with stale cached keys after rotation. It adds environment-aware caching: disabled in development (immediate updates), 5-minute TTL in production (performance + AWS rate limits).

### Code Review

**README.md** ✅
- Clear documentation of caching behavior
- Accurately describes production vs development modes
- Good security context provided

**src/utils/cert_manager/storage/aws.rs** ✅
- Clean `Option<SecretsCacheClient>` pattern for optional caching
- Proper fallback to direct client when cache disabled
- Correct error handling for ResourceNotFoundException in both paths
- Cache invalidation on update/delete operations preserved
- Good use of `if let Some(cache)` pattern

**src/utils/state.rs** ✅
- Clean environment-based TTL configuration
- `ENV_PRODUCTION` check is clear and correct
- 300-second (5 min) TTL appropriate for production
- `None` for development disables caching as intended

### Security Analysis ✅
- **No caching in dev:** Prevents stale key usage during rapid testing/rotation
- **5-min cache in prod:** Balances performance with key freshness
- **Cache invalidation:** Properly handled on secret updates/deletes
- **Error handling:** ResourceNotFoundException correctly caught in both cache and direct paths
- **No secret leakage:** Cache only stores in memory, no persistence concerns

### Recommendations (Minor)
- Consider making cache size (100) configurable via env var
- Add metrics/logging for cache hits/misses in production

### Verdict
**Production-ready security fix.** Clean implementation, proper error handling, good documentation. No blocking issues.

LGTM! 🚀

src/utils/cert_manager/storage/aws.rs
```diff
@@ -22,24 +22,37 @@ use crate::{cert_manager::storage::StorageError, utils::cert_manager::Storage};
 pub struct AwsSecretsManager {
     client: SecretsClient,
     cache: SecretsCacheClient,
+    secrets_cache_ttl: Duration,
 }
```
IngridPuppet: Any need to add this field? The cache is already initialized. This creates multiple sources of truth.

src/utils/cert_manager/storage/aws.rs
```diff
@@ -22,24 +22,37 @@ use crate::{cert_manager::storage::StorageError, utils::cert_manager::Storage};
...
 
 impl AwsSecretsManager {
     /// Create a new instance of [AwsSecretsManager] with the given AWS SDK config
-    pub async fn new(config: &SdkConfig) -> Result<Self, StorageError> {
+    pub async fn new(config: &SdkConfig, ttl: Duration) -> Result<Self, StorageError> {
```
IngridPuppet: Could you use a more descriptive arg name? It is not clear what would the ttl apply to at this point.

src/utils/cert_manager/storage/aws.rs
```diff
@@ -22,24 +22,37 @@ use crate::{cert_manager::storage::StorageError, utils::cert_manager::Storage};
...
+            ttl
+        };
+
         let cache = SecretsCacheClient::from_builder(
             asm_builder,
             NonZeroUsize::new(100).unwrap(),
-            Duration::from_secs(300),
+            internal_ttl,
             true,
         )
```
IngridPuppet: This is definitely a "hack" solution we can avoid. I'd prefer you make `cache` an `Option<SecretsCacheClient>`. WDYT?

src/utils/cert_manager/storage/redis.rs
```diff
@@ -30,15 +30,23 @@ impl Redis {
...
     async fn store(&self, key: &str, value: &str) -> Result<(), StorageError> {
         let mut conn = self.conn.clone();
-        if let Some(ttl) = self.ttl {
-            let _: () = conn.set_ex(key, value, ttl).await?;
-        } else {
-            let _: () = conn.set(key, value).await?;
+        match self.ttl {
+            Some(0) => Ok(()), // Cache disabled
+            Some(ttl) => {
+                let _: () = conn.set_ex(key, value, ttl).await?;
```
IngridPuppet: Is it required to treat "0" explicitly? I'd rather we have a silent solution, or a log at most.

src/utils/cache.rs
```diff
@@ -5,16 +5,74 @@ use crate::models::StatusListRecord;
...
+        } else {
+            Some(
+                MokaCache::builder()
+                    .time_to_live(Duration::from_secs(ttl_secs))
+                    .max_capacity(max_capacity)
+                    .build(),
+            )
+        };
+        Self { inner }
+    }
```
IngridPuppet, line 22: Is it also required here to treat 0 explicitly? Most cache solutions will process a 0 TTL natively. This looks okay to me though, but would prefer a more concise solution.

IngridPuppet (changes requested):
Hello @Awambeng - I reviewed the changes and left some comments. Please could you check? I'd globally feel there are too many changes than necessary and recommend to use a more concise syntax whenever possible.

src/utils/cert_manager/storage/aws.rs
Awambeng: Removed the `secrets_cache_ttl` field. I refactored the cache to be an `Option<SecretsCacheClient>`, which naturally handles the enabled/disabled state without needing to store the TTL twice.

src/utils/cert_manager/storage/aws.rs
Awambeng: Renamed the `ttl` argument to `secrets_cache_ttl` for better clarity.

src/utils/cert_manager/storage/aws.rs
Awambeng: Thank you! I've applied the suggestion.

src/utils/cert_manager/storage/redis.rs
Awambeng: Simplified the TTL handling in Redis storage by removing redundant logic and making the implementation more concise

src/utils/cache.rs
Awambeng: Refactored the utility cache initialization to use the more concise `.then()` pattern for optional initialization

IngridPuppet (approved):
LGTM

Hermann-Core:
Hello @Awambeng, I noticed that the implementation relies on `Option` to disable the TTL when it is set to zero, which introduces a bit of additional complexity and verbosity in the TTL handling logic. Did you check whether directly passing the TTL value without wrapping it in an `Option` already disables the cache behavior naturally when set to `0`? I believe this might already work as expected without requiring explicit optional handling.

Awambeng:
> Hello @Awambeng, I noticed that the implementation relies on `Option` to disable the TTL when it is set to zero, which introduces a bit of additional complexity and verbosity in the TTL handling logic. Did you check whether directly passing the TTL value without wrapping it in an `Option` already disables the cache behavior naturally when set to `0`? I believe this might already work as expected without requiring explicit optional handling.

Hi @Hermann-Core, thank you for the suggestion! 

I've verified that both `moka` and `aws_secretsmanager_caching` handle a 0 TTL naturally (items expire immediately or aren't cached). I've refactored the code to remove the `Option` wrappers in those layers, which significantly simplifies the logic while maintaining the 'disabled' behavior. 

I also streamlined the Redis implementation, keeping an explicit check for 0 to ensure it remains a 'silent' fallback

src/utils/cert_manager/storage/redis.rs
```diff
@@ -7,38 +7,36 @@ use crate::cert_manager::storage::{Storage, StorageError};
...
 
 #[async_trait]
 impl Storage for Redis {
     async fn store(&self, key: &str, value: &str) -> Result<(), StorageError> {
-        let mut conn = self.conn.clone();
-        if let Some(ttl) = self.ttl {
-            let _: () = conn.set_ex(key, value, ttl).await?;
-        } else {
-            let _: () = conn.set(key, value).await?;
+        if self.ttl == 0 {
```
Hermann-Core: We may want to persist a data in Redis without TTL. Here you are exiting directly without setting the value if the TTL is 0. I suggest letting the previous configuration as is with TTL explicitly set via the with_ttl method and not enforcing the TTL in the new construction so we can have choice to set a value with TTL or not. WDYT?

src/utils/state.rs
```diff
@@ -67,10 +67,14 @@ pub async fn build_state(config: &AppConfig) -> EyeResult<AppState> {
...
 
     // Initialize the storage backends for the certificate manager
-    let cache = Redis::new(redis_conn.clone());
+    let cache = Redis::new(redis_conn.clone()).with_ttl(config.redis.cert_cache_ttl);
     let cert_storage =
         AwsS3::new(&aws_config, BUCKET_NAME, config.aws.region.clone()).with_cache(cache);
-    let secrets_storage = AwsSecretsManager::new(&aws_config).await?;
+    let secrets_storage = AwsSecretsManager::new(
+        &aws_config,
+        Duration::from_secs(config.redis.secrets_cache_ttl),
```
Hermann-Core: Isn’t it better to add this TTL config to AWS configuration instead? Since it is not coupled to Redis but AWS

Hermann-Core (approved):

Hermann-Core:
Thank you for providing this configuration.

PR #136, opened 2026-05-18
Fix/fix certificate provisioning failures caused by acme account serialization issue and localstack auth changes

### LINKED ISSUE #135: Fix certificate provisioning failures caused by ACME account serialization issue and LocalStack auth changes
This ticket tracks dependency and local infrastructure updates required to fix:

1. Certificate provisioning failures caused by an ACME account serialization/deserialization issue that was patched in newer versions of the `instant-acme` crate.
2. Local integration test failures caused by recent `LocalStack` Docker images now requiring an authentication token by default.

The main goal of this ticket is to stabilize certificate issuance/renewal flows and restore reliable local testing environments through dependency upgrades and explicit service version pinning.

### 1. ACME account serialization issue

Some environments fail to provision or renew certificates because the serialized ACME account state becomes incompatible or invalid during persistence/loading.

This issue was patched upstream in a newer version of the `instant-acme` crate.

Current consequences:
- ACME account restoration may fail
- Certificate provisioning may not start correctly
- Renewal flows may break after restart/redeployment
- Persistent storage backends may contain unusable serialized account state

### 2. LocalStack authentication requirement

Recent `LocalStack` Docker Hub images require an authentication token by default.

This breaks local integration tests and Docker Compose environments that previously worked without authentication.

Current consequences:
- Local AWS emulation fails to start correctly
- Route53/S3/Secrets Manager integration tests fail
- Local development setup becomes unreliable
- CI/local parity issues appear depending on cached image versions

### Infrastructure / Docker updates

- Pin `LocalStack` Docker image version to `4.14`
- Avoid floating `latest` tags in Docker Compose and CI environments
- Update all local development and test environments to use the pinned version consistently

## Tasks

 - [x] Upgrade project dependencies
 - [x] Run certificate issuance tests
 - [x] Run certificate renewal tests
 - [x] Pin LocalStack and other services versions
 - [x] Update Docker Compose file

## Acceptance Criteria

- Certificate provisioning works correctly after restart/redeployment
- ACME account persistence/loading succeeds reliably
- Certificate renewal flows continue to function correctly
- Local integration tests pass without requiring LocalStack authentication tokens
- Docker Compose environments are reproducible across machines
- CI uses deterministic LocalStack versions
- No regressions in existing ACME challenge handling logic

## Notes

This ticket intentionally focuses on dependency management and environment reproducibility rather than feature development. The LocalStack pinning is intended to stabilize local development until authenticated LocalStack usage is explicitly supported/configured.

Estimated time: 0.5 day

Hermann-Core:
This ticket tracks dependency and local infrastructure updates required to fix:

1. Certificate provisioning failures caused by an ACME account serialization/deserialization issue that was patched in newer versions of the `instant-acme` crate.
2. Local integration test failures caused by recent `LocalStack` Docker images now requiring an authentication token by default.

The main goal of this ticket is to stabilize certificate issuance/renewal flows and restore reliable local testing environments through dependency upgrades and explicit service version pinning.

### Infrastructure / Docker updates

- Pin `LocalStack` Docker image version to `4.14`
- Avoid floating `latest` tags in Docker Compose and CI environments
- Update all local development and test environments to use the pinned version consistently

## Tasks

 - [x] Upgrade project dependencies
 - [x] Run certificate issuance tests
 - [x] Run certificate renewal tests
 - [x] Pin LocalStack and other services versions
 - [x] Update Docker Compose file

## Acceptance Criteria

- Certificate provisioning works correctly after restart/redeployment
- ACME account persistence/loading succeeds reliably
- Certificate renewal flows continue to function correctly
- Local integration tests pass without requiring LocalStack authentication tokens
- Docker Compose environments are reproducible across machines
- No regressions in existing ACME challenge handling logic

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   db:
-    image: postgres:17-alpine
+    image: postgres:18.4
     restart: always
     container_name: status-list-db
     ports:
@@ -20,7 +20,7 @@ services:
       - status-list-network
 
   redis:
-    image: redis:8.0-alpine
+    image: redis:8.6
     container_name: redis
     ports:
       - 6379:6379
@@ -29,7 +29,7 @@ services:
 
   localstack:
     container_name: localstack
-    image: localstack/localstack
+    image: localstack/localstack:4.14
     ports:
       - 4566:4566
     environment:
@@ -43,7 +43,7 @@ services:
       - status-list-network
 
   challtestsrv:
-    image: ghcr.io/letsencrypt/pebble-challtestsrv:latest
+    image: ghcr.io/letsencrypt/pebble-challtestsrv:2.10
     command: -http01 "" -https01 "" -tlsalpn01 ""
     ports:
       - 8055:8055
@@ -51,7 +51,7 @@ services:
       - status-list-network
 
   pebble:
-    image: ghcr.io/letsencrypt/pebble:latest
+    image: ghcr.io/letsencrypt/pebble:2.10
     command: -config /test/pebble_config.json -strict -dnsserver challtestsrv:8053
     ports:
       - 14000:14000
@@ -90,7 +90,7 @@ services:
       - status-list-network
 
   prometheus:
-    image: prom/prometheus
+    image: prom/prometheus:v3.11.3
     container_name: prometheus
     ports:
       - 9090:9090
@@ -106,7 +106,7 @@ services:
       - '--web.console.templates=/usr/share/prometheus/consoles'
 
   pushgateway:
-    image: prom/pushgateway
+    image: prom/pushgateway:v3.11.2
     container_name: pushgateway
     ports:
       - 9091:9091
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
       POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-postgres}
       POSTGRES_DB: ${POSTGRES_DB:-status-list}
     volumes:
-      - pgdata:/var/lib/postgresql/data
+      - pgdata:/var/lib/postgresql
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U postgres"]
       interval: 5s
@@ -106,7 +106,7 @@ services:
       - '--web.console.templates=/usr/share/prometheus/consoles'
 
   pushgateway:
-    image: prom/pushgateway:v3.11.2
+    image: prom/pushgateway:v1.11.2
     container_name: pushgateway
     ports:
       - 9091:9091
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,18 +29,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.8.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
-dependencies = [
- "cfg-if",
- "once_cell",
- "version_check",
- "zerocopy",
-]
-
-[[package]]
 name = "aho-corasick"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -72,9 +60,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -87,15 +75,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.7"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
+checksum = "52ce7f38b242319f7cabaa6813055467063ecdc9d355bbb4ce0c68908cd8130e"
 dependencies = [
  "utf8parse",
 ]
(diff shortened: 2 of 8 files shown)


Awambeng (commented):
Thanks for these improvements @Hermann-Core. 

Since the ACME serialization issue was resolved mainly through the library upgrades, it might be worth adding a small 'save and load' regression test for the ACME account.

This would ensure that if a future library update breaks serialization again, we catch it during testing rather than hitting rate limits in production

Awambeng:
I also encountered some failing tests locally. Could you please take a look as well?

image

docker-compose.yml
```diff
@@ -106,7 +106,7 @@ services:
       - '--web.console.templates=/usr/share/prometheus/consoles'
 
   pushgateway:
-    image: prom/pushgateway
+    image: prom/pushgateway:v3.11.2
```
IngridPuppet:
```suggestion
    image: prom/pushgateway:v1.11.2
```

I can't run `docker compose up --build` atm.

```
 ✘ Image prom/pushgateway:v3.11.2                     Error failed to resolve reference "docker.io/prom/pushgateway:v3.11.2": docker.io/prom/pushgateway:v3.11.2: not found
```

docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   db:
-    image: postgres:17-alpine
+    image: postgres:18.4
     restart: always
```
IngridPuppet, line 4: I keep running into the error below despite removing volumes. It looks like an issue with my local environment, but it is worth confirming whether or not you're running into the same.

```
status-list-db  | Error: in 18+, these Docker images are configured to store database data in a
status-list-db  |        format which is compatible with "pg_ctlcluster" (specifically, using
status-list-db  |        major-version-specific directory names).  This better reflects how
status-list-db  |        PostgreSQL itself works, and how upgrades are to be performed.
status-list-db  | 
status-list-db  |        See also https://github.com/docker-library/postgres/pull/1259
status-list-db  | 
status-list-db  |        Counter to that, there appears to be PostgreSQL data in:
status-list-db  |          /var/lib/postgresql/data (unused mount/volume)
status-list-db  | 
status-list-db  |        This is usually the result of upgrading the Docker image without
status-list-db  |        upgrading the underlying database using "pg_upgrade" (which requires both
status-list-db  |        versions).
status-list-db  | 
status-list-db  |        The suggested container configuration for 18+ is to place a single mount
status-list-db  |        at /var/lib/postgresql which will then place PostgreSQL data in a
status-list-db  |        subdirectory, allowing usage of "pg_upgrade --link" without mount point
status-list-db  |        boundary issues.
status-list-db  | 
status-list-db  |        See https://github.com/docker-library/postgres/issues/37 for a (long)
status-list-db  |        discussion around this process, and suggestions for how to do so.
status-list-db exited with code 1 (restarting)
```

IngridPuppet (changes requested):
I tried running `docker compose up --build` and left some comments. Please could you check?

IngridPuppet:
> I also encountered some failing tests locally. Could you please take a look as well?
> 
> image

Same. `cargo test` is treating the example code in a documentation block in `bits_validation.rs` as a test to run. This works though: 

```bash
cargo nextest run --workspace --all-targets --all-features
```

Not sure if that's intended. There should be a way to document code without it being treated a test if not wanted.

docker-compose.yml
Awambeng: This error looks familiar. It usually happens when Postgres 18+ is started with an old volume that was created using the previous data path `/var/lib/postgresql/data`, while the new version expects data under `/var/lib/postgresql` with a versioned structure. When it detects the old layout, it refuses to start to avoid potential data corruption from an incompatible upgrade.

IngridPuppet: Thank you @Awambeng - I can confirm being able to docker compose up after changing `/var/lib/postgresql/data` to `/var/lib/postgresql` in the compose yaml.

docker-compose.yml
Hermann-Core: Thank you for that. That was changed by the agent without me noticing 🤦🏻‍♂️

docker-compose.yml
Hermann-Core: Exactly @Awambeng, that location has been updated.
> Important Change: [the PGDATA environment variable of the image was changed to be version specific in PostgreSQL 18 and above](https://github.com/docker-library/postgres/pull/1259). For 18 it is /var/lib/postgresql/18/docker. Later versions will replace 18 with their respective major version (e.g., /var/lib/postgresql/19/docker for PostgreSQL 19.x). The defined VOLUME was changed in 18 and above to /var/lib/postgresql. Mounts and volumes should be targeted at the updated location. This will allow users upgrading between PostgreSQL major releases to use the faster --link when running pg_upgrade and mounting /var/lib/postgresql.

Hermann-Core:
> Same. `cargo test` is treating the example code in a documentation block in `bits_validation.rs` as a test to run. This works though:
> 
> ```shell
> cargo nextest run --workspace --all-targets --all-features
> ```
> 
> Not sure if that's intended. There should be a way to document code without it being treated a test if not wanted.

We are running the full test suite using `cargo-nextest`, which by default does not execute doctest examples. I have fixed the issue though. The CI pipeline will be improved as part of the upcoming review ticket we are preparing, following the same foundation and standards already established in the cloud wallet project.

Hermann-Core:
Hello @Awambeng @IngridPuppet thank you for the reviews. The issues have been fixed, could you have another look?

Screenshot 2026-05-19 at 12 55 22

docker-compose.yml
IngridPuppet, line 1: Do you plan to add the integration test raised by @Awambeng ?

IngridPuppet (approved):
I can `docker compose up` and `cargo test` with the update.

docker-compose.yml
Hermann-Core: Yes I will add that.

Awambeng (commented):
Hi @Hermann-Core, thank you for adding the new test.

The current integration tests validate the full provisioning and renewal flows, but they do not explicitly verify ACME account serialization/deserialization.

Since the serialization issue was mainly resolved through dependency upgrades, I think it could still be valuable to add a dedicated regression test for the ACME account persistence flow to catch future serialization issues early. WDYT?

Hermann-Core:
> Since the serialization issue was mainly resolved through dependency upgrades, I think it could still be valuable to add a dedicated regression test for the ACME account persistence flow to catch future serialization issues early. WDYT?

Instead of creating isolated tests for each individual component of the ACME flow, I decided to validate the entire certificate provisioning flow end-to-end. This allows us to test not only the account serialization and persistence logic which are themselves part of the provisioning process, but also the interaction between all components involved in the flow. As a result, if an issue occurs at any stage of the provisioning pipeline, it can be detected through the integration test as well.

Awambeng (approved):

IngridPuppet (approved):

Awambeng:
> > Since the serialization issue was mainly resolved through dependency upgrades, I think it could still be valuable to add a dedicated regression test for the ACME account persistence flow to catch future serialization issues early. WDYT?
> 
> Instead of creating isolated tests for each individual component of the ACME flow, I decided to validate the entire certificate provisioning flow end-to-end. This allows us to test not only the account serialization and persistence logic which are themselves part of the provisioning process, but also the interaction between all components involved in the flow. As a result, if an issue occurs at any stage of the provisioning pipeline, it can be detected through the integration test as well.

Okay, It's fine then. Thank you!

Hermann-Core:
Hello @IngridPuppet @Awambeng I have fixed the merge conflicts, could you have another look or approve again so we can merge?

IngridPuppet (approved):

PR #146, opened 2026-07-02
feat: allow external certificate management

### LINKED ISSUE #140: Support external certificate management
In order to support deployments in environments, where certificate management is already handled(f.e. using cert-manager), it should be possible to load signing material into the server from an external source, instead of relying on the internal ACME implementation.

wistefan:
closes #140 

- certificate and signing key can be mounted as files
- ACME and file-mount are mutually exclusive - only one mechanism can be used


Hermann-Core:
Hello @wistefan sorry for the late reply. Thank you for the contrinution, could you resolve the conflicts?

wistefan:
> Hello @wistefan sorry for the late reply. Thank you for the contrinution, could you resolve the conflicts?

Hi  @Hermann-Core, sorry that it took a while, I was on vacation. Pls check if its ok now

Hermann-Core:
> Hi @Hermann-Core, sorry that it took a while, I was on vacation. Pls check if its ok now

Hello @wistefan, welcome back from your vacation! It looks like the CI pipeline is currently failing. Could you take a look and fix the issues?

Issue #152, opened 2026-07-02
Fix Status enum ApplicationSpecific variant to conform with spec


Hermann-Core:
## Objective

Fix the `Status` enum's `ApplicationSpecific` variant to correctly conform to the [OAuth Status List spec draft 21](https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/), which defines status values as integers where `0 = VALID`, `1 = INVALID`, `2 = SUSPENDED` and values ≥ `256` are reserved for application-specific use.

## Context

The current `Status` enum in `src/models.rs` defines an `ApplicationSpecific` variant that may not correctly enforce the range constraint (values must be ≥ 256) or properly serialize/deserialize per the spec.

## Current State

The `Status` enum (in `src/models.rs`) likely looks something like:

```rust
pub enum Status {
    Valid,
    Invalid,
    Suspended,
    Custom(u32),
}
```

The `ApplicationSpecific` variant needs validation that the value is ≥ 256 and proper JSON serialization matching the spec's integer representation.

## Deliverables

- [ ] Review `src/models.rs` for the `Status` enum definition.
- [ ] Ensure `ApplicationSpecific` variant validates that values are ≥ 256.
- [ ] Implement proper `Serialize`/`Deserialize` that produces/consumes plain integers (not objects).
- [ ] Add unit tests for all status values:
  - `0` → `Valid`
  - `1` → `Invalid`
  - `2` → `Suspended`
  - `256` → `Custom(256)`
  - Values 2–255 should be rejected or handled as errors.
- [ ] Ensure the status bits encoding/decoding in `src/utils/lst_gen.rs` is compatible.
- [ ] Run `cargo test` to verify.

## Acceptance Criteria

- `Status` enum correctly maps to and from the integer representation per spec draft 21.
- `Custom` values < 256 are rejected.
- All existing tests pass.
- New unit tests cover edge cases.

## References

- Spec: https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/

PR #182, opened 2026-07-06
Fix two draft-21 §5.2 CWT compliance bugs (type header + tag-18); add spec test vectors + compliance matrix

martcpp:
Fixes two MUST-level draft-ietf-oauth-status-list-21 §5.2 deviations in the CWT issuance path, hardens the CWT test, adds authoritative spec test vectors, and lands the compliance matrix from the audit.

## Bugs fixed (both in `issue_cwt()`, `get_status_list.rs`)
1. **CWT type header (label 16)** now emits the full media type  `application/statuslist+cwt` instead of the abbreviated `statuslist+cwt`. §5.2 requires the full media type for the COSE header (the abbreviated form is correct only for the JWT `typ`). New constant `STATUS_LIST_CWT_TYPE_VALUE` in `constants.rs`; the now-unused `STATUS_LISTS_HEADER_CWT` was removed rather than left dead.
2. **CWT tagging** now emits `COSE_Sign1_Tagged` (CBOR tag 18) via `to_tagged_vec()`. §5.2 MUSTs the COSE_Sign1/Mac0 tag and forbids the CWT tag 61); the code previously emitted untagged bytes. Matches the spec's worked
   example (`d2 84 …`) and this repo's `docs/architecture.md` hex dump.

## Test changes
- `test_get_status_list_success_cwt`: decodes via `from_tagged_slice()`, asserts claim keys against **literal** spec integers (`2`/`6`/`4`/`65534`/`65533`) instead of the encoder's own constants (previous assertion was tautological), and asserts the label-16 value directly off the protected header.
 New `test_spec_vector_1_bit` / `test_spec_vector_2_bit` in `lst_gen.rs` using  draft-21 §4.1/§4.2 worked vectors (`[0xB9,0xA3]`→`eNrbuRgAAhcBXQ`, `[0xC9,0x44,0xF9]`→`eNo76fITAAPfAgc`). Hard assertion is the backend-independent **decompress** direction; encode direction kept as a secondary canonical-output pin. Reuses `encode_compressed`, no new abstractions.

## Docs
- New `docs/compliance_matrix.md`: all 18 audited rows (Fixed / Compliant / Deviation / Out-of-scope / Hardening-note). 

## Deferred (tracked separately)
- `Status::ApplicationSpecific` to #152 
-  cert EKU to #136.
- `aggregation_uri` (§9, OPTIONAL)  #183
- Historical resolution `?time=` (§8.4, OPTIONAL)  #184
- gzip: respect `Accept-Encoding` + don't compress CWT (§8.2, SHOULD)  #185


src/web/handlers/status_list/constants.rs
```diff
@@ -1,7 +1,11 @@
 pub(super) const ACCEPT_STATUS_LISTS_HEADER_JWT: &str = "application/statuslist+jwt";
 pub(super) const ACCEPT_STATUS_LISTS_HEADER_CWT: &str = "application/statuslist+cwt";
 pub(super) const STATUS_LISTS_HEADER_JWT: &str = "statuslist+jwt";
-pub(super) const STATUS_LISTS_HEADER_CWT: &str = "statuslist+cwt";
+
+/// COSE header label 16 ("type") value for the CWT protected header (draft-ietf-oauth-status-list-21 §5.2).
+/// Unlike the JWT `typ` header, which uses the abbreviated `statuslist+jwt` form, the CWT
+/// type header MUST carry the full media type.
+pub(super) const STATUS_LISTS_CWT_TYPE_VALUE: &str = "application/statuslist+cwt";
```
ndefokou, line 6: Verified correct per §5.2: the COSE type label (16) MUST carry the full media type `application/statuslist+cwt`, distinct from the JWT `typ` header which §5.1 mandates as the abbreviated `statuslist+jwt`. Matches the spec's worked CWT header (`78 1a application/statuslist+cwt`) and this repo's `docs/architecture.md:76`. Removing the dead `STATUS_LISTS_HEADER_CWT` rather than leaving it was the right call.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -210,7 +210,10 @@ fn issue_cwt(
     let protected = HeaderBuilder::new()
         .algorithm(Algorithm::ES256)
         .value(HeaderParameter::X5Chain.to_i64(), x5chain_value)
-        .value(CWT_TYPE, CborValue::Text(STATUS_LISTS_HEADER_CWT.into()))
+        .value(
+            CWT_TYPE,
+            CborValue::Text(STATUS_LISTS_CWT_TYPE_VALUE.into()),
```
ndefokou, line 233: Blocking — separate §4.3 violation in this same `issue_cwt` function. §5.2 requires the `status_list` claim (65533) to conform to §4.3, and §4.3 mandates `lst` be a **CBOR byte string (major type 2)** holding the raw zlib-compressed bytes — not a base64url text string. But the `status_list` map built above (lines ~188-201) emits `CborValue::Text(status_record.status_list.lst.clone())`, where `lst` is the base64url string from `encode_compressed` → `base64url::encode`.

The spec's worked CWT example confirms the byte-string form: `63 6c7374 4a 78dadbb918000217015d` (`4a` = bstr len 10, the literal `78 da db b9 18 00 02 17 01 5d` behind `eNrbuRgAAhcBXQ`).

Suggested fix — decode the stored base64url `lst` to raw bytes and emit `CborValue::Bytes`:
```rust
let lst_bytes = base64url::decode(&status_record.status_list.lst)
    .map_err(|e| { tracing::error!("invalid lst: {e:?}"); StatusListError::InternalServerError })?;
// ...
(CborValue::Text("lst".into()), CborValue::Bytes(lst_bytes)),
```
`bits` (Integer, major type 0) and the text keys are already correct per §4.3 — only the `lst` value type is wrong.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -225,7 +228,8 @@ fn issue_cwt(
         })
         .build();
 
-    let cwt_bytes = sign1.to_vec().map_err(|err| {
+    // Tagged as COSE_Sign1_Tagged (CBOR tag 18), per draft-ietf-oauth-status-list-21 §5.2.
+    let cwt_bytes = sign1.to_tagged_vec().map_err(|err| {
```
ndefokou, line 250: Verified correct: `to_tagged_vec()` produces CBOR tag 18 (`d2 84 …`), satisfying §5.2's MUST to emit `COSE_Sign1_Tagged` (18) and the MUST NOT to use the CWT tag (61). The prior `to_vec()` emitted untagged bytes, which was non-compliant. Heads-up: this changes the CWT endpoint's wire bytes for any external consumer (they were already non-conformant) — worth a CHANGELOG note.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -500,11 +507,26 @@ mod tests {
...
         let ttl = claims
             .iter()
-            .find(|(k, _)| k == &CborValue::Integer(TTL.into()))
+            .find(|(k, _)| k == &CborValue::Integer(65534i32.into()))
             .unwrap()
             .1
             .clone();
         assert_eq!(ttl, CborValue::Integer(300.into()));
+
+        // Verify the protected header's type (label 16) is the full media type, per §5.2 -
```
ndefokou: Follow-up to the §4.3 fix above: the `lst` assertion currently does `assert_eq!(lst, CborValue::Text(encode_compressed(&[0, 0, 0]).unwrap()));` (just above this block). That codifies the non-compliant text form rather than the spec. Once the producer emits `lst` as `CborValue::Bytes(...)`, this should expect a byte string (major type 2), the raw compressed bytes:
```rust
let raw = base64url::decode(&encode_compressed(&[0, 0, 0]).unwrap()).unwrap();
assert_eq!(lst, CborValue::Bytes(raw));
```
The key lookup stays text `"lst"` (correct per §4.3); only the expected *value* type changes from `Text` to `Bytes`. The claim-key hardening you added here (literal integers, label-16 assertion) is great — please extend the same spec-literal discipline to the `lst` value type.

src/utils/lst_gen.rs
```diff
@@ -468,4 +468,70 @@ mod tests {
...
+                    Status::INVALID
+                } else {
+                    Status::VALID
+                },
+            })
+            .collect();
+
+        let result = create_status_list(updates).unwrap();
+        assert_eq!(result.bits, 1);
+        assert_eq!(result.lst, "eNrbuRgAAhcBXQ");
```
ndefokou, line 792: Nit: this test relies solely on encode-direction equality (`result.lst == "eNrbuRgAAhcBXQ"`). Per the header comment above, that equality is coupled to flate2's zlib backend (miniz_oxide vs system zlib via the `zlib` feature). Unlike `test_spec_vector_1_bit`/`_2_bit`, there's no decompress fallback here, so a backend change would break this test without proving anything about production correctness. Consider asserting the decompressed content too (decode + `ZlibDecoder` → `[0xB9, 0xA3]`) so the production-path guarantee holds independent of the backend.

ndefokou (changes requested):
Spec review of PR #182 against draft-ietf-oauth-status-list-21.

## What's correct (nice work)
- **§5.2 type header (label 16) → `application/statuslist+cwt`** (full media type). Confirmed against §5.2 and the spec's worked example header `…781a6170706c69636174696f6e2f7374617475736c6973742b637774` (`78 1a` = text(26)). JWT `typ` is correctly left as the abbreviated `statuslist+jwt` per §5.1 — good that the two weren't collapsed. This also reconciles the code with `docs/architecture.md:76`, which already documented the compliant form.
- **§5.2 COSE tagging → `to_tagged_vec()`** emits CBOR tag 18 (`COSE_Sign1_Tagged`), matching `d2 84 …` and the MUST NOT to use the CWT tag (61). The previous untagged `to_vec()` was non-compliant.
- **Hardened CWT test** (claim keys asserted as literal integers `2/6/4/65534/65533`, plus a label-16 assertion) — the prior constants-keyed assertions were tautological.
- **§4.1/§4.2 spec test vectors** — good interop pin (the decompress direction is the backend-independent guarantee).

## Blocking — one MUST-level CWT issue remains within this PR's scope
The CWT `status_list.lst` is emitted as a **CBOR text string** (base64url), but §5.2 routes the `status_list` claim (65533) to **§4.3**, which mandates `lst` be a **CBOR byte string (major type 2)** containing the *raw* zlib-compressed byte array — not base64url text. The spec's worked CWT example encodes it as a byte string: `63 6c7374 4a 78dadbb918000217015d` (`4a` = bstr len 10, literal `78 da …` zlib stream), i.e. the bytes behind `eNrbuRgAAhcBXQ`, not the base64url text itself.

Production emits `CborValue::Text(status_record.status_list.lst.clone())` (where `lst` is the base64url string from `encode_compressed` → `base64url::encode`), and the test asserts that same `CborValue::Text(...)`. The test thus codifies the non-compliant behavior; a conformant external CWT verifier would expect a byte string. Both producer and test need to change (see inline). The matrix's CWT rows should then move to "Fixed" rather than "Compliant". `bits` (Integer/major type 0) and the text keys `bits`/`lst` are already correct per §4.3 — only the `lst` value type is wrong.

## Non-blocking / nits
- `lst_gen.rs` `test_spec_vector_1_bit_through_create_status_list` asserts only encode-direction equality; add a decompress fallback so it survives a flate2 zlib-backend switch (see inline).
- `keygen.rs` `LineEnding::default()` → `LF` is an unrelated drive-by (cross-platform PEM determinism) — harmless, worth noting it's outside the §5.2 scope.
- The `to_tagged_vec()` change alters the CWT endpoint's wire bytes for any external consumer (they were non-conformant) — a CHANGELOG note would help.
- Deferred items (gzip `Accept-Encoding`/CWT scoping §8.2 → #185, `aggregation_uri` §9 → #183, `?time=` §8.4 → #184, `ApplicationSpecific` → #152, EKU → #136) are well-scoped and referenced.

Requesting changes for the §4.3 `lst` byte-string item, a MUST within the CWT path this PR explicitly audits and claims compliant.

martcpp:
@ndefokou @Christiantyemele @Ngha-Boris pls check this

ndefokou (approved):

ndefokou (approved):

docs/compliance_matrix.md
Hermann-Core, line 1: This seems to be out of scope and tracked by a separate ticket.

src/utils/keygen.rs
```diff
@@ -78,7 +78,7 @@ impl Keypair {
     pub fn to_pkcs8_pem(&self) -> Result<String, Error> {
         self.repr
             .key
-            .to_pkcs8_pem(LineEnding::default())
+            .to_pkcs8_pem(LineEnding::LF)
```
Hermann-Core, line 81: Do you have any valid reason for this change?

src/utils/lst_gen.rs
```diff
@@ -468,4 +468,81 @@ mod tests {
...
+    // The following two tests use the worked examples from
+    // draft-ietf-oauth-status-list-21 §4.1 / §4.2, rather than hand-rolled bit
+    // patterns, as an authoritative interop check.
+    //
+    // The decompress-direction assertion is the real interop guarantee: DEFLATE
+    // permits multiple valid encodings of the same input, so it must hold
+    // regardless of the zlib backend flate2 is built against. The encode-direction
+    // assertion additionally pins flate2's exact byte-for-byte output (and thus
+    // implicitly `Compression::best()`), but is coupled to the backend rather than
+    // being a pure spec requirement.
```
Hermann-Core: We should avoid verbose comments.

src/web/handlers/status_list/constants.rs
```diff
@@ -1,7 +1,11 @@
 pub(super) const ACCEPT_STATUS_LISTS_HEADER_JWT: &str = "application/statuslist+jwt";
 pub(super) const ACCEPT_STATUS_LISTS_HEADER_CWT: &str = "application/statuslist+cwt";
 pub(super) const STATUS_LISTS_HEADER_JWT: &str = "statuslist+jwt";
-pub(super) const STATUS_LISTS_HEADER_CWT: &str = "statuslist+cwt";
+
+/// COSE header label 16 ("type") value for the CWT protected header (draft-ietf-oauth-status-list-21 §5.2).
+/// Unlike the JWT `typ` header, which uses the abbreviated `statuslist+jwt` form, the CWT
+/// type header MUST carry the full media type.
```
Hermann-Core: We should avoid overcommenting. Comments should be brief and focus on important invariants, architectural details, or other long-term relevant information.

src/web/handlers/status_list/constants.rs
```diff
@@ -1,7 +1,11 @@
 pub(super) const ACCEPT_STATUS_LISTS_HEADER_JWT: &str = "application/statuslist+jwt";
 pub(super) const ACCEPT_STATUS_LISTS_HEADER_CWT: &str = "application/statuslist+cwt";
 pub(super) const STATUS_LISTS_HEADER_JWT: &str = "statuslist+jwt";
-pub(super) const STATUS_LISTS_HEADER_CWT: &str = "statuslist+cwt";
+
```
Hermann-Core, line 4: I would have recommended simply updating the value of the existing constant rather than introducing a new one and replacing its usage throughout the codebase. That is precisely the purpose of constants: to provide a single source of truth so that changing the underlying value only requires updating it in one place.

Hermann-Core (commented):
Hello @martcpp, I left somme comments for your attention, could you check and address them?

docs/compliance_matrix.md
martcpp: It in the tickets scope

Hermann-Core: why do we need a compliance matrix in the repo first? and why do we need it in this ticket?

src/utils/keygen.rs
martcpp: SO it not platform depend thanks it have error of my end those about this

docs/compliance_matrix.md
martcpp: it to keep track of what the new update come with it a no harm down that track

src/web/handlers/status_list/constants.rs
martcpp: STATUS_LISTS_HEADER_CWT sat next to STATUS_LISTS_HEADER_JWT, which correctly holds the abbreviated form ("statuslist+jwt"). If I'd kept the CWT constant's name and just changed its value to the full media type, the two would still read as parallel siblings while secretly differing in kind one abbreviated, one full media type. That symmetry is plausibly why the original bug existed: the code assumed CWT's type header should mirror JWT's abbreviated convention, which COSE's label 16 doesn't actually follow. Keeping the old name doesn't fix that trap for the next person, it just relocates it. The new name puts the asymmetry in the identifier itself rather than only in a comment someone can skip.

src/utils/keygen.rs
Hermann-Core: Could you revert the change done here? the `Default` implementation will choose the correct line ending depending on the platform.

src/web/handlers/status_list/constants.rs
Hermann-Core: The naming suggests that both of these are intended to represent the values of the `typ` header, so I do not quite see the connection with the symmetry you are referring to. But still

src/utils/keygen.rs
martcpp: it will fail for people  using other machine  hat are not unix  likes keeping to use a fix lf for linux make more sense thank

Hermann-Core: This is not portable. As stated in the `LineEnding` documentation:
> Use `LineEnding::default` to get an appropriate line ending for the current operating system.

Windows uses `CRLF`, whereas Unix-like operating systems use `LF`. To keep the implementation portable across platforms, we should use `LineEnding::default()` instead of hardcoding `LF`.

martcpp: it fail on window usig default

martcpp: harding coding it force it to state one place rather then dancing around

Hermann-Core:
@martcpp you have conflicts and a failling pipeline.

martcpp:
> @martcpp you have conflicts and a failling pipeline.

On it

martcpp:
@Hermann-Core pls recheck let merge today thanks this ticket has been here for too long

Hermann-Core (approved):

PR #188, opened 2026-07-07
Refactor/redesign endpoints to follow best restful api practices

### LINKED ISSUE #145: Redesign endpoints to follow best RESTful API practices
## Objective

Redesign the server's HTTP endpoints to follow best RESTful API practices. The current endpoint structure (defined in `src/web/handlers/` and `src/web/mod.rs`) uses action-based path segments that deviate from REST conventions. This issue ensures the API is intuitive, resource-oriented, and aligned with the [OAuth Status List spec](https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/).

## Current State

- `src/web/mod.rs` — defines routes such as `/publish`, `/update`, `/status-list/{id}`, etc.
- `src/web/handlers/status_list/` — handlers for `publish_status`, `update_status`, `get_status_list`, `issue_credential`.
- Endpoints mix noun-based and verb-based naming (e.g., `publish-status-list` is verb-based, should be a `POST /status-lists` resource creation).
- HTTP methods are not used consistently (e.g., updates may not use `PATCH`/`PUT`).
- Status codes may not fully align with REST conventions.

## Deliverables

- [x] Audit all current endpoints and document them (method, path, request/response body, status codes).
- [x] Propose a new RESTful endpoint scheme
- [x] Ensure proper use of HTTP methods: `GET` (idempotent read), `POST` (create), `PUT`/`PATCH` (update).
- [x] Use plural nouns for resource collections, kebab-case for multi-word segments.
- [x] Return appropriate HTTP status codes: `201 Created` for new resources, `200 OK` for successful reads/updates, `404 Not Found`, `400 Bad Request`, `401 Unauthorized`, `403 Forbidden`, `409 Conflict`.
- [x] Support content negotiation (`Accept` / `Content-Type` headers) where applicable.
- [x] Implement pagination, filtering, and sorting for list endpoints if/if they return collections.
- [x] Update all integration tests and the artillery load tests to match new paths.
- [x] Update `demo/` notebooks to use new endpoints.
- [x] Update Helm chart and any client tooling references.

## References

- REST API best practices: https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design

Hermann-Core:
Redesigned the server's HTTP endpoints to follow best RESTful API practices. The current endpoint structure (defined in `src/web/handlers/` and `src/web/mod.rs`) uses action-based path segments that deviate from REST conventions. This PR ensures the API is intuitive, resource-oriented, and aligned with the [RESTful API best practices](https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design).


artillery-tests/tests/spike-test.yml
martcpp, line 1: create a postman collection for this project it will be very helpful and create since api base even alot test can be done the and load e.t.c @Hermann-Core @IngridPuppet what you take about this

IngridPuppet: I'm open to trying that out. But can we handle it separately? It may unnecessarily make this PR harder to review.

martcpp: not saying t should be done in same pr a tracking issue will do

Hermann-Core: Hello @martcpp, we will have to discuss about it. Let's have an internal discussion.

artillery-tests/tests/auth-test.yml
```diff
@@ -53,12 +53,11 @@ scenarios:
...
-          url: "/statuslists/publish"
+      - put:
+          url: "/api/v1/status-lists/{{ listId }}/statuses"
           headers:
             Authorization: "Bearer {{ token }}"
           json:
-            list_id: "{{ listId }}"
             status:
               - index: 1
                 status: "VALID"
```
IngridPuppet, line 63: - Could you share your reasons for using PUT? For a new list, POST would still make sense to me. 
- There should no longer be a nested `status` in the payload, only the status array directly. In the worst case, it should be made plural: `statuses`.

docs/blog/status-list-server.adoc
```diff
@@ -86,9 +86,9 @@ curl -s http://localhost:8000/
 
 - `GET /` → "Status list Server"
 - `GET /health` → liveness probe
-- `POST /credentials` → register an issuer’s public key (JWK)
-- `POST /statuslists/publish` (auth) → create a status list
-- `PATCH /statuslists/update` (auth) → update a status list
+- `POST /api/v1/credentials` → register an issuer’s public key (JWK)
+- `PUT /api/v1/status-lists/{list_id}/statuses` (auth) → create a status list
+- `PATCH /api/v1/status-lists/{list_id}/statuses` (auth) → update a status list
 - `GET /statuslists/{list_id}` (public) → gzip-compressed JWT or CWT
```
IngridPuppet: For consistency
```suggestion
- `GET /api/v1/status-lists/{list_id}` (public) → gzip-compressed JWT or CWT
```

docs/blog/status-list-server.adoc
```diff
@@ -86,9 +86,9 @@ curl -s http://localhost:8000/
 
 - `GET /` → "Status list Server"
 - `GET /health` → liveness probe
-- `POST /credentials` → register an issuer’s public key (JWK)
-- `POST /statuslists/publish` (auth) → create a status list
-- `PATCH /statuslists/update` (auth) → update a status list
+- `POST /api/v1/credentials` → register an issuer’s public key (JWK)
+- `PUT /api/v1/status-lists/{list_id}/statuses` (auth) → create a status list
+- `PATCH /api/v1/status-lists/{list_id}/statuses` (auth) → update a status list
```
IngridPuppet, line 91: The general reason for having a POST than PUT/PATCH endpoint is that initially, the resource has no identifier, and that's acquired through POST upon resource creation. In our case, the resource management identifier is "outsourced". Do you think we could have just one endpoint for both creation and update? We currently do an extra call in the plugin just because of this.

src/web/handlers/status_list/publish_status.rs
```diff
@@ -46,14 +47,11 @@ pub async fn publish_status(
                 lst: stl.lst,
             };
 
-            let sub = format!(
-                "https://{}/statuslists/{}",
-                appstate.server_domain, payload.list_id
-            );
+            let sub = format!("https://{}/statuslists/{}", appstate.server_domain, list_id);
```
IngridPuppet:
```suggestion
            let sub = format!("https://{}/api/v1/status-lists/{}", appstate.server_domain, list_id);
```

src/web/handlers/status_list/publish_status.rs
```diff
@@ -182,25 +178,26 @@ mod tests {
...
             mock_db
                 .append_query_results::<status_lists::Model, Vec<_>, _>(vec![
-                    vec![],                  // find_one_by in handler returns None
-                    vec![new_token.clone()], // insert_one return
-                    vec![new_token.clone()], // find_one_by in test verification
+                    vec![],
+                    vec![new_token.clone()],
+                    vec![new_token.clone()],
                 ])
                 .into_connection(),
```
IngridPuppet, line 188: Could you restore the comments? They look meaningful in this case, to understand when they will be yielded.

src/models.rs
```diff
@@ -106,10 +106,15 @@ pub struct StatusList {
...
-    pub list_id: String,
+pub struct CreateStatusRequest {
+    pub status: Vec<StatusEntry>,
+}
+
+/// Request payload for updating status entries in an existing status list
+#[derive(Deserialize)]
+pub struct UpdateStatusRequest {
     pub status: Vec<StatusEntry>,
 }
```
IngridPuppet: I see no harm keeping the same structure. They are identical. The intent can be understood from context.

src/startup.rs
```diff
@@ -71,15 +77,31 @@ impl HttpServer {
...
+/// Protocol endpoint for fetching status list tokens.
+///
+/// `GET /statuslists/{list_id}` — returns a status list token in JWT or CWT format.
+///
+/// This endpoint follows the OAuth Status List specification and is NOT prefixed
+/// with `/api/v1` because it is a protocol-level endpoint that must match the `uri`
+/// embedded in referenced tokens.
+fn protocol_routes() -> Router<AppState> {
+    Router::new().route("/{list_id}", get(get_status_list))
+}
```
IngridPuppet: The spec does not mandate a specific structure for the URI apart from being a URI. Examples are non-normative. It's best to keep consistent with one's implementation.

IngridPuppet (commented):
Hello @Hermann-Core - I've reviewed the new API contract and left some comments. Please could you check?

artillery-tests/tests/auth-test.yml
Hermann-Core: Thanks. On the method choice: I would keep `PUT` here because the client supplies `{list_id}`, so the create request targets a known resource URI. `POST` would fit better if the server generated the identifier. For the payload shape, agreed: the singular `status` wrapper is confusing. I changed the request body to use `statuses`. I kept the JSON object wrapper instead of a raw top-level array so the request remains extensible if we later need metadata around the status updates.

docs/blog/status-list-server.adoc
Hermann-Core: Agreed. I updated the public retrieval endpoint to `GET /api/v1/status-lists/{list_id}` for consistency with the API namespace.

docs/blog/status-list-server.adoc
Hermann-Core: I would keep create and update separate for now. `PUT` creates the known resource and rejects an existing list, while `PATCH` applies partial status changes to an existing list. Combining them into one endpoint would introduce upsert semantics, which could be useful for the plugin, but I think that should be an explicit API design change with dedicated tests rather than folded into this refactor.

src/web/handlers/status_list/publish_status.rs
Hermann-Core: Agreed. I changed `sub` to use the canonical public status-list URL: `https://{domain}/api/v1/status-lists/{list_id}`.

src/web/handlers/status_list/publish_status.rs
Hermann-Core: Agreed. I restored the comments explaining the `MockDatabase` query result ordering; they are useful context for why each mocked result is yielded.

src/models.rs
Hermann-Core: Agreed. Since the create and update payloads are identical, I replaced the separate request structs with one shared `StatusesRequest` containing `statuses: Vec<StatusEntry>`.

src/startup.rs
Hermann-Core: Agreed. The spec does not require the legacy `/statuslists/{list_id}` shape, so I changed the implementation to use `GET /api/v1/status-lists/{list_id}` as the public retrieval endpoint and removed the `/statuslists/{list_id}` route rather than keeping it as an alias.

Hermann-Core:
Hello @IngridPuppet thanl you for the review. I have addressed the comments, could you take another look?

demo/workflows/01-an-issuer-maintains-a-status-list.ipynb
IngridPuppet, line 1: Nitpick. Did you run these notebooks again? I'm surprised not to see new output changes.

IngridPuppet (approved):
Thank you for the update. All my concerns were addressed.

Ngha-Boris (approved):

demo/workflows/01-an-issuer-maintains-a-status-list.ipynb
Hermann-Core: I had this error when trying to run the demo, do you have an idea of what happened?
```sh
(base) ➜  demo git:(refactor/redesign-endpoints-to-follow-best-restful-api-practices) conda env create -f environment.yml
2 channel Terms of Service accepted
Retrieving notices: done
Channels:
 - conda-forge
 - defaults
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - _libgcc_mutex==0.1=main

Current channels:

  - https://conda.anaconda.org/conda-forge
  - https://repo.anaconda.com/pkgs/main
  - https://repo.anaconda.com/pkgs/r

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

Ngha-Boris (approved):

PR #201, opened 2026-07-09
Feat/cache certificate chain with invalidation on re provisioning

### LINKED ISSUE #158: Cache certificate chain with invalidation on re-provisioning
## Objective

Implement a caching mechanism for the certificate chain extracted by the certificate manager, with cache invalidation triggered only when a new certificate is provisioned. This reduces redundant certificate parsing/validation overhead on every request.

## Context

Currently, the certificate chain may be re-extracted or re-parsed on each request or on each certificate manager operation. This is inefficient.

A well-designed cache should:
- Cache the parsed certificate chain in memory.
- Invalidate only when a new certificate is provisioned (via ACME or store).
- Be thread-safe (usable from async handlers).

## Deliverables

- [x] Design the certificate chain cache.
- [x] Implement cache invalidation hook: when a new certificate is provisioned, the cache is invalidated.
- [x] Add cache hit/miss metrics (prometheus counters).
- [x] Add configurable cache TTL as a safety net (optional, in addition to invalidation).
- [x] Add tests for cache hit, miss, and invalidation scenarios.

## Acceptance Criteria

- Certificate chain is cached and reused across requests.
- Cache is invalidated when a new certificate is provisioned.
- Thread-safe and async-compatible.
- Metrics expose cache hit/miss rates.

Estimated time: 1 day

Hermann-Core:
Implemented a caching mechanism for the certificate chain extracted by the certificate manager, with cache invalidation triggered only when a new certificate is provisioned. This reduces redundant certificate parsing/validation overhead on every request.

## Metrics

The cache exposes the following Prometheus counters (all annotated with a `domain` label for multi-domain observability):

- `certificate_chain_cache_hits_total` — counter incremented on every read that resolves from the cache.
- `certificate_chain_cache_misses_total` — counter incremented on every read that falls through to storage.
- `certificate_chain_cache_replacements_total` — counter incremented whenever a freshly provisioned chain replaces the cached entry (post-provisioning hook).

All three counters are described via `describe_counter!` and zero-initialised at startup so they show up in Prometheus scrapes immediately, before first use.

## Public API changes (breaking for direct consumers of `cert_manager` / `cache`)

- `cert_manager::CertManager::cert_chain_parts` now returns `Result<Option<CertificateChain>, CertError>` where `CertificateChain = Arc<[String]>` (was `Vec<String>`). This removes an allocation on the hot path and makes cache hits cheaply shareable by reference.
- `cache::Cache` (the status-list cache) was renamed to `cache::StatusListCache` and now stores `Arc<StatusListRecord>` instead of `StatusListRecord` to avoid per-read cloning.
- `cache::CertChainCache` is a new, separate cache type for certificate chains. Its constructor now takes a TTL and a domain label: `CertChainCache::new(ttl: Duration, domain: impl AsRef<str>)`.
- The previously public `web::handlers::status_list::get_status_list::StatusListToken` struct (JWT claims envelope) is now private; JWT response construction uses a borrow-only `StatusListClaims` serializer and `StatusListToken` is only reconstructed in tests.
- `cert_manager::DEFAULT_CHAIN_CACHE_TTL` is a new `pub const Duration` (1 hour) used as the single source of truth for the default cache TTL across `CertManager` and the config layer.

## Behaviour changes worth noting

- On re-provisioning of a certificate (`request_certificate` succeeds), the cached chain is eagerly replaced via the `cache_provisioned_chain` hook rather than simply invalidated, so the next read skips the storage load and parse entirely.
- In multi-replica deployments, only the replica that actually provisions the certificate replaces its in-memory cache. Non-provisioning replicas rely on the configured TTL as the only refresh mechanism. The default is 1 hour; setting it to `0` disables expiry entirely (only safe when the process re-provisions or restarts on every rotation).


Ngha-Boris (commented):
.

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,53 @@
...
+        };
+        metrics::counter!(metric).increment(1);
+        cached
+    }
+
+    pub async fn insert(&self, key: String, value: CertificateChain) {
+        self.inner.insert(key, value).await;
+    }
+
+    pub async fn update(&self, key: String, value: CertificateChain) {
```
Ngha-Boris: **Issue:** The `update` method calls `insert` under the hood, which replaces the entry rather than invalidating it. The metric name (`certificate_chain_cache_invalidations_total`) and the method name are misleading.

**Suggestion:** Rename the method to `replace_on_provision` or `invalidate_and_insert`, and update the metric name to `certificate_chain_cache_replacements_total` to accurately reflect what happens. If you want a true invalidation metric, add a separate `invalidate` method that calls `self.inner.invalidate(key)` and records that metric.

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,53 @@
...
+        metrics::counter!(INVALIDATION_METRIC).increment(1);
+        self.inner.insert(key, value).await;
+    }
+
+    #[cfg(test)]
```
Ngha-Boris, line 106: **Issue:** This `invalidate` method is gated behind `#[cfg(test)]`, which means production code cannot manually invalidate the cache. More importantly, the acceptance criteria state *“Cache is invalidated when a new certificate is provisioned.”* The test suite only exercises this test-only method directly; it does not verify that the actual provisioning path (`request_certificate`) triggers the cache invalidation/replacement.

**Suggestion:** Make `invalidate` a regular `pub` method for operational flexibility, but more importantly, add a test that verifies the cache is replaced when a new certificate is provisioned (either by mocking the full ACME flow or by extracting the post-provision cache hook into a small, testable helper).

src/utils/cert_manager.rs
```diff
@@ -338,23 +358,20 @@ impl CertManager {
...
+    /// Extract individual certificates from the certificate chain and return shared base64-encoded chain parts.
     ///
     /// This function will return `None` if the server certificate was not found.
     ///
     /// # Errors
     /// Returns an error if the certificate chain cannot be parsed or if there was an issue when trying to retrieve the server certificate
-    pub async fn cert_chain_parts(&self) -> Result<Option<Vec<String>>, CertError> {
-        use base64::prelude::{BASE64_STANDARD, Engine as _};
-
+    pub async fn cert_chain_parts(&self) -> Result<Option<CertificateChain>, CertError> {
```
Ngha-Boris, line 396: **Issue:** This is a manual cache-aside pattern. Two concurrent requests can both miss the cache, both read from storage, and both parse the certificate chain. This is harmless for correctness but defeats the purpose of avoiding redundant parsing on the first few concurrent requests.

**Suggestion:** Consider using `moka::future::Cache::try_get_with` so that concurrent misses are coalesced into a single storage load and parse:
```rust
let result = self.cert_chain_cache.inner
    .try_get_with(cert_key.clone(), async { ... })
    .await;
```
If you prefer to keep the manual pattern, add a brief comment documenting the intentional optimistic approach.

src/utils/cert_manager/tests.rs
```diff
@@ -322,3 +332,76 @@ async fn test_cert_chain_parts() {
...
+    let first = cert_manager.cert_chain_parts().await.unwrap().unwrap();
+    let second = cert_manager.cert_chain_parts().await.unwrap().unwrap();
+
+    assert_eq!(first.len(), 2);
+    assert!(Arc::ptr_eq(&first, &second));
+    assert_eq!(cert_storage.load_count(), 1);
+}
+
+#[tokio::test]
+async fn test_cert_chain_cache_invalidation_reloads_chain() {
```
Ngha-Boris, line 403: **Issue:** This test validates the test-only `invalidate` method, not the real invalidation hook triggered by provisioning. It does not prove that `request_certificate` (or any provisioning path) actually replaces the cached certificate chain.

**Suggestion:** Add a test that verifies the cache is replaced when a new certificate is provisioned. If mocking the full ACME flow is too heavy, extract the post-provision cache update into a small, testable helper method and unit-test that. At minimum, document the test gap in a `TODO` comment.

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,53 @@
...
+const INVALIDATION_METRIC: &str = "certificate_chain_cache_invalidations_total";
+
+#[derive(Clone)]
+pub struct CertChainCache {
+    inner: Cache<String, CertificateChain>,
+}
+
+impl CertChainCache {
+    pub fn new(ttl: Duration) -> Self {
+        let builder = Cache::builder().max_capacity(CACHE_CAPACITY);
```
Ngha-Boris, line 62: **Issue:** `CACHE_CAPACITY` is set to `16`, but this cache is owned by a single `CertManager` and only ever stores one key (`cert_key`). The magic number is undocumented and confusing.

**Suggestion:** Either replace it with `1` (since one entry is sufficient) or add a comment explaining why headroom is reserved (e.g., for future multi-domain support).

src/utils/cert_manager/tests.rs
```diff
@@ -322,3 +332,76 @@ async fn test_cert_chain_parts() {
     let parts = cert_manager.cert_chain_parts().await.unwrap().unwrap();
     assert_eq!(parts.len(), 2);
 }
+
+#[tokio::test]
```
Ngha-Boris, line 336: **Issue:** The test verifies storage load count and `Arc::ptr_eq`, which is good. However, there is no test verifying the Prometheus hit/miss counters (`certificate_chain_cache_hits_total` / `certificate_chain_cache_misses_total`) are actually incremented.

**Suggestion:** Add a lightweight unit test for `CertChainCache` itself that asserts the hit and miss counters are incremented. You can use `metrics::with_local_recorder` in a test to capture the counter values.

src/web/handlers/status_list/publish_status.rs
```diff
@@ -17,14 +17,16 @@ pub async fn publish_status(
     Extension(issuer): Extension<String>,
     Json(payload): Json<StatusRequest>,
 ) -> Result<impl IntoResponse, StatusListError> {
+    let StatusRequest { list_id, status } = payload;
```
Ngha-Boris: **Issue:** This file contains refactoring (payload destructuring, formatting changes) that is unrelated to the ticket scope: *“Cache certificate chain with invalidation on re-provisioning.”*

**Suggestion:** These are harmless changes, but they increase review surface area and risk unrelated regressions. In the future, move unrelated refactors to a separate cleanup PR to keep the focus strictly on the ticket deliverables.

src/web/handlers/status_list/update_status.rs
```diff
@@ -16,16 +16,18 @@ pub async fn update_status(
     Extension(issuer): Extension<String>,
     Json(payload): Json<StatusRequest>,
 ) -> Result<impl IntoResponse, StatusListError> {
+    let StatusRequest { list_id, status } = payload;
```
Ngha-Boris: **Issue:** Similar to `publish_status.rs`, this file contains unrelated refactoring (payload destructuring, `std::mem::take` optimizations) that is outside the scope of the ticket.

**Suggestion:** Keep unrelated refactors in a separate PR to maintain a clean review boundary.

Ngha-Boris (changes requested):
Thanks for the PR! The implementation is on the right track, but there are a few issues that need to be addressed before merging most notably a compilation error, some misleading naming around cache invalidation, and a test gap regarding the actual provisioning hook.

Please see the comments below.

src/utils/cache/cert_chain.rs
Hermann-Core: The production provisioning path uses the `replace` helper (`cert_chain_cache.replace`), not a true evict. After provisioning a new certificate the cache is eagerly repopulated so the next read never hits storage. A manual `invalidate` (evict-only) would regress to a storage load + parse on next read. Keeping it test-only is intentional.

The post-provision cache-replace logic is now extracted into `cache_provisioned_chain(&cert_pem)` (the exact hook `request_certificate` calls after storing the new cert), and a new test `test_cache_provisioned_chain_replaces_cached_entry` exercises that real code path without mocking ACME. It asserts the cached entry is replaced with the new chain (ptr-inequal) and that no additional storage load occurs.

src/utils/cert_manager.rs
Hermann-Core: `try_get_with` is a poor fit here because the `Ok(None)` case (certificate not yet provisioned) must remain a cache miss. The cache value type is `CertificateChain` with no valid sentinel for absence. Using `try_get_with` would require restructuring the cache to store an `Option<CertificateChain>` or introducing a sentinel value, both adding complexity for no real-world win.

The thundering-herd concern is negligible in practice: concurrent cold misses only fire after TTL expiry (default 1h) and each parse is ~3 base64-encoded PEM chunks — microseconds of work.

src/utils/cert_manager/tests.rs
Hermann-Core: The hit/miss counter logic is a one-line `is_some()` branch — trivially correct by inspection. Testing it requires introducing a `metrics-util` dev-dependency and a `with_local_recorder` harness, but there is no existing metrics-testing convention in this codebase (no `metrics-util` dependency, no `with_local_recorder` usage anywhere in `src/`). The existing `Arc::ptr_eq` + `load_count` tests already validate the caching behavior end-to-end. The metric is pure side-effect instrumentation that does not affect correctness. Adding a metrics-test harness would be a worthwhile repo-wide investment, but it is out of scope for this PR and should be established as a project-wide convention rather than bolted on here.

src/web/handlers/status_list/publish_status.rs
Hermann-Core: This comment is stale — `publish_status.rs` is **not** in the final PR diff. The refactoring existed in an intermediate commit (`85d74bf`) but was removed during the merge with `main` (`b97db3f`). No unrelated changes remain in that file.

The scoping point is taken for future PRs.

src/web/handlers/status_list/update_status.rs
Hermann-Core: Keeping these changes — they are genuine efficiency improvements:

- `std::mem::take(&mut exact_status_list.status_list.lst)` avoids cloning the `String` (the status list bitstring) before passing it to `update_status_list`. The old code cloned the entire record unnecessarily.
- Using `list_id` (already extracted from the `Path`) instead of `exact_status_list.list_id` avoids an extra field access after the move.
- `store.update_one(&list_id, exact_status_list)` moves the record by value instead of `exact_status_list.clone()`.

These are correct, low-risk, and eliminate unnecessary allocations on every status update request. The scoping point is fair — I will keep unrelated refactors in a separate PR going forward.

Hermann-Core:
Hello @Ngha-Boris thanks for the reviews, the comments have been addressed, could you check again?

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    pub async fn get(&self, key: &str) -> Option<CertificateChain> {
+        let cached = self.inner.get(key).await;
+        let metric = if cached.is_some() {
+            HIT_METRIC
+        } else {
+            MISS_METRIC
+        };
+        metrics::counter!(metric).increment(1);
```
martcpp: (describe_counter! + zero-init) and add at least one test asserting hit/miss counters move. Otherwise "expose hit/miss rates" is only partially met.

src/utils/cert_manager.rs
```diff
@@ -150,6 +157,15 @@ impl CertManager {
         self
     }
 
+    /// Override the in-memory parsed certificate chain cache TTL.
+    ///
+    /// A zero duration keeps the cache active with no TTL safety expiry; cache
```
martcpp, line 169: document that in multi-replica deployments TTL is the only refresh mechanism for non-provisioning replicas, and that 0 disables it.

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,57 @@
...
+pub type CertificateChain = Arc<[String]>;
+
+// One CertManager serves a single TLD+1, so only one chain is ever cached.
+const CACHE_CAPACITY: u64 = 1;
+const HIT_METRIC: &str = "certificate_chain_cache_hits_total";
+const MISS_METRIC: &str = "certificate_chain_cache_misses_total";
+const REPLACEMENT_METRIC: &str = "certificate_chain_cache_replacements_total";
+
+#[derive(Clone)]
+pub struct CertChainCache {
```
martcpp: Replace moka with ArcSwapOption/RwLock<Option<…>> for this single-entry cache, or at least switch the read path to try_get_with to close the stampede window and make get+insert atomic.

src/utils/cert_manager.rs
```diff
@@ -26,10 +27,13 @@ use tracing::{error, info, instrument, warn};
 use x509_parser::pem::Pem;
 
 use crate::{
+    cache::CertChainCache,
     cert_manager::{challenge::ChallengeHandler, http_client::DefaultHttpClient, storage::Storage},
     utils::keygen::Keypair,
 };
 
+const DEFAULT_CACHE_TTL: Duration = Duration::from_hours(1);
```
martcpp: Collapse the duplicated TTL default to one source of truth.

src/utils/cert_manager.rs
```diff
@@ -338,23 +355,20 @@ impl CertManager {
...
+    /// Extract individual certificates from the certificate chain and return shared base64-encoded chain parts.
     ///
     /// This function will return `None` if the server certificate was not found.
     ///
     /// # Errors
     /// Returns an error if the certificate chain cannot be parsed or if there was an issue when trying to retrieve the server certificate
-    pub async fn cert_chain_parts(&self) -> Result<Option<Vec<String>>, CertError> {
-        use base64::prelude::{BASE64_STANDARD, Engine as _};
-
+    pub async fn cert_chain_parts(&self) -> Result<Option<CertificateChain>, CertError> {
```
martcpp, line 396: the cert_chain_parts public return-type change and StatusListToken visibility change in the PR description.

martcpp (changes requested):

martcpp (commented):
Consider a metric label (e.g. cert_key/domain) so multi-domain deployments don't aggregate blindly.

src/utils/cache/cert_chain.rs
Hermann-Core: **`ArcSwapOption`/`RwLock<Option<…>>`:** This would remove moka and lose the TTL safety net entirely. In a multi-replica deployment, a non-provisioning replica relies on TTL as the only refresh mechanism for picking up a freshly provisioned chain — a `RwLock<Option<…>>` has no built-in expiry, so a long-lived replica would serve the stale chain indefinitely. moka also handles concurrent insert/evict atomically; a hand-rolled `RwLock` would need explicit care to avoid a torn replacement between a `get` (read lock) and a `replace` (write lock). moka is already pulled in by `StatusListCache` (same dependency, no new cost) and transparently handles TTL, eviction and concurrency for free.

**`try_get_with`:** Same reasoning as the earlier thread on `cert_manager.rs`: the `Ok(None)` case (certificate not yet provisioned) cannot be represented in a `Cache<String, CertificateChain>` — `try_get_with` would force caching absence via a sentinel, which is the wrong semantic (a later `request_certificate` then has to remember to evict the sentinel rather than just `replace` the real chain). The stampede window for a single-entry cache with a 1-hour default TTL is at most one extra parse per expiry cycle, and parsing three base64 PEM chunks is microseconds of work — not worth redesigning the cache around.

Hermann-Core:
Hello @martcpp, could you check again?

src/utils/state.rs
```diff
@@ -103,6 +103,7 @@ pub async fn build_state(config: &AppConfig) -> EyeResult<AppState> {
     .with_cert_storage(cert_storage)
     .with_secrets_storage(secrets_storage)
     .with_challenge_handler(challenge_handler)
+    .with_cert_chain_cache_ttl(Duration::from_secs(config.server.cert.chain_cache_ttl))
```
martcpp, line 102: recorder installed aafter handle resolution during build_state  Acceptance criterion "metrics expose cache hit/miss rates" is not met. Fix by resolving counters per-call (recommended) or installing the recorder before build_state, and fix the unit test so it would have caught this.

Cargo.toml
martcpp, line 1: Pin rust-version in Cargo.toml or avoid Duration::from_hours

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,153 @@
...
+        self.inner.invalidate(key).await;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_hit_miss_counters_increment() {
```
martcpp: This test masks the bug above: it installs the recorder before constructing the cache, which is the opposite of production ordering. Rewrite it to construct the cache first, then assert counters still land.

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,153 @@
...
+            (h, m, r)
+        };
+        // Zero-init so the counters appear in Prometheus before first use.
+        hit_counter.increment(0);
+        miss_counter.increment(0);
+        replacement_counter.increment(0);
+
+        let builder = Cache::builder().max_capacity(CACHE_CAPACITY);
+        let inner = if ttl.is_zero() {
+            builder.build()
```
martcpp, line 64: ttl == 0 = cache forever here, but StatusListCache::new (status_list.rs:14) treats 0 = disabled. Same config file, same sentinel, opposite meaning — align or document loudly

src/utils/cert_manager.rs
```diff
@@ -27,9 +28,15 @@ use x509_parser::pem::Pem;
...
     cert_manager::{challenge::ChallengeHandler, http_client::DefaultHttpClient, storage::Storage},
-    utils::keygen::Keypair,
+    utils::{cache::CertChainCache, keygen::Keypair},
 };
 
+/// Default cache TTL when no override is supplied.
+///
+/// Exported as a single source of truth: `Config::load` references this
+/// constant so the runtime default and the code fallback always agree.
+pub const DEFAULT_CHAIN_CACHE_TTL: Duration = Duration::from_hours(1);
```
martcpp: Duration::from_hours needs a very recent stable and Cargo.toml has no rust-version pin. Pin rust-version or use from_secs(3600).

src/utils/cert_manager.rs
```diff
@@ -150,6 +164,24 @@ impl CertManager {
         self
     }
 
+    /// Override the in-memory parsed certificate chain cache TTL.
```
martcpp, line 167: Add to this doc comment: the whole staleness story is only safe because the signing key is stable across renewals (signing_key_pem reuses the stored secret). If renewal ever rotates the key, every staleness window becomes a token-validation outage.

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,153 @@
...
+        self.inner.insert(key, value).await;
+    }
+
+    #[cfg(test)]
+    pub(crate) async fn invalidate(&self, key: &str) {
+        self.inner.invalidate(key).await;
+    }
+}
+
+#[cfg(test)]
```
martcpp, line 138: Missing tests: TTL expiry, ttl == 0 never-expire branch, replacement-counter increment under a recorder.

martcpp (changes requested):
Document server.cert.chain_cache_ttl (README/env sample), including the 0 = never expire semantics and the multi-replica staleness bound; consider reconciling the 0 semantics with cache.ttl.

src/utils/cache/cert_chain.rs
martcpp: get - load - insert isn't atomic. A miss can load the old cert, get overtaken by the renewal task's replace(), then insert the stale chain back over the fresh one  stale for up to the TTL, which is the exact failure this PR exists to prevent. try_get_with is a poor fit (it'd cache the Ok(None) not-yet-provisioned case, which must stay a miss). Instead, guard the insert: have replace() bump a generation counter, capture it before the storage load, and skip the insert if it changed. Or make the read path non-writing and let only the provisioning hook populate the cache. If you consider the staleness bounded-and-acceptable (stable key, old cert still time-valid), that's a legitimate call  but then document it as accepted rather than leave it implicit

src/utils/cache/cert_chain.rs
Hermann-Core: The replacement-counter test has been added as part of fixing the recorder-initialization bug (see updated `test_counters_work_when_cache_constructed_before_recorder`). Regarding TTL expiry and the `ttl=0` never-expire branch: these behaviors are entirely delegated to the `moka` crate, which is heavily tested upstream. Testing them here would just be testing `moka`'s internal timer wheels rather than our logic, so I've opted to rely on their guarantees and instead heavily documented the semantics.

Hermann-Core:
@martcpp I've addressed all the review comments and pushed the fixes. Could you please take another look when you have a moment? Thanks!

src/utils/cache/cert_chain.rs
```diff
@@ -0,0 +1,193 @@
...
+/// key (`cert_key`). One [`CertManager`](crate::cert_manager::CertManager)
+/// instance owns exactly one `CertChainCache`, so the cache holds at most one
+/// entry per running process.
+///
+/// # TTL semantics
+///
+/// A **zero** TTL (`Duration::ZERO`) keeps the cache active with **no expiry**;
+/// entries persist until explicitly replaced by the provisioning hook. This
+/// differs from [`StatusListCache`](super::StatusListCache), where `ttl = 0`
+/// **disables** the cache entirely (inserts expire immediately).
```
Christiantyemele, line 23: **Important — Stale doc comment contradicts `StatusListCache` behavior in this PR**

The doc comment says:
> This differs from `StatusListCache`, where `ttl = 0` **disables** the cache entirely (inserts expire immediately).

But the new `StatusListCache::new` (line 17–27 of `status_list.rs`) does the opposite: when `ttl_secs == 0` it calls `builder.build()` with **no** `time_to_live`, so entries **never expire**. The explicit new test `cache_ttl_zero_means_no_expiry` confirms this (the old test was `cache_disabled_returns_none` which asserted the opposite).

This comment should be corrected to match the actual code:
```rust
/// A **zero** TTL (`Duration::ZERO`) keeps the cache active with **no expiry**;
/// entries persist until explicitly replaced by the provisioning hook. This is
/// now **consistent** with [`StatusListCache`](super::StatusListCache), where
/// `ttl = 0` also means entries never expire.
```

src/utils/cache/status_list.rs
```diff
@@ -0,0 +1,77 @@
...
+    inner: MokaCache<String, Arc<StatusListRecord>>,
+}
+
+impl StatusListCache {
+    /// Creates a cache.
+    ///
+    /// A **zero** TTL (`ttl_secs = 0`) keeps the cache active with **no
+    /// expiry**; entries persist until explicitly invalidated. This is
+    /// consistent with [`CertChainCache`](super::CertChainCache).
+    pub fn new(ttl_secs: u64, max_capacity: u64) -> Self {
```
Christiantyemele, line 18: **Important — Undocumented behavioral change to `cache.ttl` config**

The old `Cache::new(0, ...)` called `time_to_live(Duration::from_secs(0))`, which caused entries to expire immediately — the old test `cache_disabled_returns_none` asserted `get` returns `None` after `insert`. The new `StatusListCache::new(0, ...)` calls `builder.build()` without `time_to_live`, so entries **never expire**.

This is a **breaking semantic change** to the pre-existing `cache.ttl` config (default 300s). Any deployment that sets `APP_CACHE__TTL=0` to disable the status-list cache will, after this PR, serve cached entries indefinitely.

**Multi-replica staleness risk:** The `update_status` handler (`update_status.rs:89`) only invalidates the cache entry on the replica that processed the update. Non-provisioning replicas with TTL=0 will keep the stale status-list entry forever (until process restart). With the old TTL=0→disabled behavior, those replicas always read from the DB.

**Recommendation:** Call this out explicitly in the PR description and in README/helm documentation. If the intent is to standardize TTL=0 semantics with the cert-chain cache, that is a reasonable design choice — but the behavioral change to `cache.ttl` must be surfaced as a breaking change rather than introduced silently alongside the rename.

Christiantyemele (changes requested):
## Summary

The cert-chain caching mechanism is well-implemented: lazy counter resolution, eager post-provisioning replacement, `Arc<[String]>` sharing to eliminate hot-path allocations, and good test coverage (cache-hit/miss/replacement semantics with `Arc::ptr_eq` + storage load counts). `cargo check`, `cargo clippy --all-targets -- -D warnings`, and the full test suite (24 unit tests including the new `test_counters_work_when_cache_constructed_before_recorder` and provisioning-hook tests) all pass clean.

**Findings: 0 Bug, 2 Important, 0 Suggestion, 0 Nit, 0 Question.**

Key themes: (1) The `CertChainCache` doc comment contradicts the new `StatusListCache` behavior in this same PR — it claims TTL=0 "disables" the status-list cache, but the rewritten `StatusListCache` does the opposite (permanent entries). (2) The `StatusListCache` TTL=0 semantics silently changed from the old "disabled" behavior to "never expire" — an undocumented behavioral change to the pre-existing `cache.ttl` config that risks stale reads in multi-replica deployments.

## Prior review notes (resolved/agreed)

- **Ngha-Boris** — misleading `update` name → **resolved**: renamed to `replace` with `REPLACEMENT_METRIC`. Cache-aside vs `try_get_with` → author declined; reasoning (cache absence can't be represented as `CertificateChain` sentinel) is sound. Counters test → **resolved**: added `test_counters_work_when_cache_constructed_before_recorder`. Unrelated refactors in `publish_status.rs`/`update_status.rs` → publish_status changes removed; update_status changes kept with allocation-avoidance justification.
- **martcpp** — domain label on metrics → **resolved**: `domain_label` added. `describe_counter!` + zero-init → **resolved**: implemented with `init_cert_chain_cache_counters()` in `main.rs`. TTL constant dedup → **resolved**: `DEFAULT_CHAIN_CACHE_TTL`. Document `chain_cache_ttl` in README/helm and reconcile 0-semantics with `cache.ttl` → **still partially open**: code comments document semantics well, but README and helm `values.yaml`/`values-local.yaml` do not include `APP_SERVER__CERT__CHAIN_CACHE_TTL`. `ArcSwap`/`RwLock` instead of moka → author declined; TTL safety net argument is reasonable.

Ngha-Boris (approved):
LGTM please address the comments of @Christiantyemele before merging

martcpp (approved):

src/utils/cache/cert_chain.rs
Hermann-Core: Good catch. I restored `StatusListCache` to its previous `ttl=0` behavior instead of changing the cert-chain docs to match the accidental new behavior. `StatusListCache::new(0, ...)` now disables the status-list cache again, so the cert-chain comment is accurate: cert-chain `Duration::ZERO` means no expiry, while status-list `ttl=0` means immediate expiry/disabled.

src/utils/cache/status_list.rs
Hermann-Core: Agreed. I restored the previous `APP_CACHE__TTL=0` behavior so this PR no longer silently changes status-list cache semantics. `StatusListCache::new(0, ...)` now expires inserts immediately again, and the test is back to asserting `get` returns `None`. I also documented the separate cert-chain TTL override in `.env.template`, Helm values, and README, including `0 = never expire` and the multi-replica staleness bound.

Christiantyemele (approved):

Issue #162, opened 2026-07-02
Support for widely used DNS providers beyond AWS Route53


Hermann-Core:
## Objective

Add support for widely used DNS providers beyond AWS Route53 to the ACME DNS-01 challenge solver. This enables operators to use their preferred DNS provider for automated certificate provisioning.

## Context

Currently, the certificate manager's DNS-01 challenge only supports AWS Route53 for completing ACME DNS-01 challenges. Many deployments use other DNS providers (Cloudflare, Google Cloud DNS, Azure DNS, DigitalOcean, etc.) and cannot use the automated ACME provisioning without Route53.

Making the DNS provider pluggable aligns with the broader goal of making the certificate manager more flexible (#159, #161).

## Deliverables

- [ ] Define a `DnsProvider` trait with methods for:
  - Creating/deleting DNS TXT records.
  - Polling for record propagation.
  ```rust
  #[async_trait]
  pub trait DnsProvider: Send + Sync {
      async fn create_txt_record(&self, domain: &str, value: &str) -> Result<()>;
      async fn delete_txt_record(&self, domain: &str) -> Result<()>;
      async fn verify_propagation(&self, domain: &str, value: &str) -> Result<()>;
  }
  ```
- [ ] Implement providers for:
  - **Cloudflare** (via API token).
  - **Google Cloud DNS** (via service account JSON).
  - **Azure DNS** (via service principal).
  - **ACME-DNS** (self-hosted, provider-agnostic).
  - Keep **AWS Route53** as one provider implementation.
- [ ] Add configuration to select the DNS provider
- [ ] Integrate the trait into the DNS-01 challenge solver.
- [ ] Add tests for each provider (using mocks or test containers where feasible).
- [ ] Update `.env.template` and Helm values.
- [ ] Document supported providers and their configuration.

## Acceptance Criteria

- DNS provider is selectable via configuration.
- At least Cloudflare and Google Cloud DNS are implemented in addition to Route53.
- The `DnsProvider` trait is generic enough for future providers.
- Route53 implementation is refactored to use the trait.
- All tests pass.
- Documentation lists all supported providers with setup instructions.

## References

- Related: #159 (cert provisioning config), #161 (builder pattern for cert manager)
- Lego (Go ACME library) DNS providers: https://go-acme.github.io/lego/

martcpp:
est time 2 days

Issue #155, opened 2026-07-02
Improve server-to-server authentication with key rotation support


Hermann-Core:
## Objective

Replace the current authentication mechanism with a robust server-to-server (S2S) authentication model that supports key rotation, multiple signing keys, and graceful key transitions without downtime.

## Context

The server currently uses a simple bearer-token / static-key authentication mechanism (see `src/web/auth/`). For production server-to-server communication, a more robust mechanism is needed that supports:
- Multiple valid keys simultaneously (for rotation).
- Key identifier (kid) in JWT headers.
- Graceful key rollover (old key remains valid until expiry after new key is provisioned).
- JWKS endpoint for key distribution.

## Deliverables

- [ ] Design the S2S auth model (JWT with kid, JWKS endpoint, key rotation strategy).
- [ ] Implement JWKS endpoint (`/.well-known/jwks.json`) exposing public keys.
- [ ] Support multiple concurrent signing keys with `kid` identification.
- [ ] Implement key rotation: new key can be provisioned while old key remains valid until configurable expiry.
- [ ] Add configurable key rotation interval and key overlap period.
- [ ] Update auth middleware to validate JWTs against any valid key in the JWKS.
- [ ] Add tests for key rotation scenarios (old token still valid after rotation, new token valid, expired key rejected).
- [ ] Document the auth model and key rotation process.
- [ ] Update `src/config.rs` with auth-related configuration options.

## Acceptance Criteria

- JWKS endpoint is available and returns all valid public keys.
- Key rotation occurs without downtime or rejected tokens during overlap period.
- Auth middleware validates tokens against any key in the JWKS.
- All tests pass including rotation scenario tests.
- Documentation covers key rotation and S2S auth setup.

Hermann-Core:
Reopening as part of #141. This issue captures the requirement to move issuer/server-to-server authentication to a key-rotation-ready model with `kid`, JWKS, and graceful rollover.

Issue #141, opened 2026-07-02
Status List Server Hardening & Production Readiness


Hermann-Core:
## Overview

This EPIC tracks all work needed to harden the status-list-server for production: ensuring spec compliance with [OAuth Status List draft 21](https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/), improving infrastructure (CI, Docker, dependencies), enhancing security (auth, key management, caching), increasing flexibility (configurable database, cert provisioning, DNS providers), and improving operability (observability, error handling, API design).

The EPIC also incorporates action items from the architecture review discussion in [#139](https://github.com/adorsys/status-list-server/issues/139), including caching strategy re-evaluation, log collection, least-privilege DB access, and HA considerations.

## References

- Spec: https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/
- Discussion: https://github.com/adorsys/status-list-server/issues/139

Issue #160, opened 2026-07-02
Add OpenTelemetry collector and make tracing configurable (dev/prod)


Hermann-Core:
## Objective

Add an OpenTelemetry collector to the deployment stack and make the tracing, metrics, and logs collector configurable depending on the environment (dev or prod). In dev, they should output to stdout/console; in prod, it should export to an OpenTelemetry collector via OTLP.

## Context

Currently, the server uses Prometheus for metrics (`src/utils/metrics.rs`) and likely has basic tracing/logging. For production observability, we need structured tracing exported via OpenTelemetry, with environment-specific configuration.

## Deliverables

- [x] Create a collector initialization module (`telemetry`) that switches based on `ENVIRONMENT` env var:
  - `ENVIRONMENT=dev` (or not set): Output to stdout/console
  - `ENVIRONMENT=prod`: Export via OTLP to collector
- [x] Add an OpenTelemetry collector service to `docker-compose.yml`:
  - Jaeger (or Tempo) for traces
  - Prometheus for metrics (via Otel collector receiver)
  - Loki (or equivalent) for logs
- [x] Instrument key spans:
  - HTTP request handling (auto-instrument via `tower-http` tracing middleware)
  - Database queries (auto-instrument via `sqlx` if using tracing integration)
  - Certificate provisioning (manual spans around cert renewal logic)
  - Status list generation (manual spans around list creation/update)
- [x] Update Helm chart to optionally deploy an Otel collector sidecar or standalone service (controlled by `.Values.otelCollector.enabled`)
- [x] **Fix the `values.yaml` inconsistency** (see parent EPIC #452): Currently `otelCollector.enabled: false` but `APP_TELEMETRY__ENABLED: "true"`. Either:
  - Enable the collector by default in `values.yaml`, OR
  - Default `APP_TELEMETRY__ENABLED: "false"` until a collector is configured
  - Document the relationship between these settings

## Acceptance Criteria

- [x] Telemetry output is configurable based on `ENVIRONMENT` variable
- [x] Dev mode logs to stdout (no collector required for local development)
- [x] Prod mode exports via OTLP to the collector
- [x] `docker-compose up` brings up the Otel collector alongside the server
- [x] Key operations are instrumented with spans (HTTP requests, DB queries, cert provisioning, list generation)
- [x] Helm chart's `otelCollector.enabled` and `statuslist.env.APP_TELEMETRY__ENABLED` are consistent (no silent telemetry failures)
[...]

PR #212, opened 2026-07-14
feat: Refactor server to hexagonal architecture and implement in-memory and PostgreSQL adapters

### LINKED ISSUE #186: Refactor server to hexagonal architecture with domain models, ports, and dependency injection
## Objective

Restructure the codebase around a hexagonal (ports-and-adapters) architecture, similar to the approach used in the [cloud-identity-wallet](https://github.com/ADORSYS-GIS/cloud-identity-wallet) project, to clearly separate domain logic from infrastructure and to make upcoming changes (multiple backends, key rotation, spec evolution) easier to implement and test.

## Context

The current code mixes HTTP handlers, database access, external service clients, and domain logic in a way that makes it hard to swap implementations or unit-test components in isolation. Introducing a hexagonal architecture will provide:

- A stable **domain layer** containing entities, value objects, and use-case invariants.
- **Inbound ports** exposed by the domain (e.g., publish status list, update statuses, retrieve status list token).
- **Outbound ports** for infrastructure concerns (database, cache, certificate store, secrets store, DNS, metrics).
- **Adapters** implementing each outbound port (Postgres, Redis, S3, AWS Secrets Manager, in-memory mocks, etc.).
- Dependency injection of adapters into the application services, rather than constructing them in `startup.rs`.

This directly supports the upcoming changes tracked under #141: backend feature flags (#168), certificate provisioning strategies (#159), status-list cache (#167), DNS providers (#162), and S2S auth (#155).

## Deliverables

- [ ] Audit the current directory structure and identify entities, repositories, external clients, and handlers.
- [ ] Define the domain module (`src/domain`) with core entities and value objects:
  - `StatusList`, `StatusEntry`, `Status`, `Issuer`, `Credential`, etc.
- [ ] Define inbound application services (`src/application` or in-domain `services`):
  - `PublishStatusList`, `UpdateStatuses`, `GetStatusListToken`.
- [ ] Define outbound port traits (`src/ports`) for:
  - `StatusListRepository`
  - `CredentialRepository`
  - `StatusListCache`
  - `CertificateProvider`
  - `SecretStore`
  - `DnsProvider`
  - `MetricsCollector` (optional, keep if already instrumented)
- [ ] Move existing implementations into adapter modules (`src/adapters/postgres`, `src/adapters/redis`, `src/adapters/aws`, `src/adapters/memory`, etc.).
- [ ] Refactor `AppState` / `startup.rs` so that adapters are injected via constructors, not built inline.
- [ ] Keep the existing behavior intact; no functional changes in this ticket.
- [ ] Move to compile-time feature flags per adapter where appropriate (coordinate with #168).
- [ ] Add unit tests for each application service using in-memory / mock adapters.
- [ ] Update or create `ARCHITECTURE.md`/`docs/hexagonal-architecture.md` explaining the layers and the dependency rule.
- [ ] Update CI and README references.

## Acceptance Criteria

- Domain crates/modules do not depend on Axum, sqlx, AWS SDK, Redis, or any infrastructure crate.
- All outgoing dependencies in domain/application code are expressed through port traits.
- Existing integration tests still pass against the default Postgres/Redis/AWS stack.
- New unit tests pass using mock/in-memory adapters.
- A `memory-only` feature build compiles without Postgres/Redis/AWS dependencies.
- Architecture document is reviewed and linked from README.

## References

- Parent EPIC: #141
- Related: #168 (backend feature flags), #159 (certificate provisioning), #167 (status-list cache), #162 (DNS providers), #155 (S2S auth), #160 (observability)
- Reference project: https://github.com/ADORSYS-GIS/cloud-identity-wallet

est: 1 day

Ngha-Boris:
# Restructure the server around a hexagonal (ports-and-adapters) architecture

Closes #186 (part of EPIC #141)

## What this does

Separates domain logic from infrastructure behind explicit boundaries:

- **`src/domain`** — entities and value objects (`StatusList`, `StatusEntry`, `Status`, `Issuer`, `Credential`, `StatusListRecord`, `StatusListSnapshot`) plus the status-list bitstring invariants (bit-packing, zlib + base64url encoding, bit-width widening, reserved-value rejection). No infrastructure dependency of any kind.
- **`src/ports`** — outbound traits (`StatusListRepository`, `CredentialRepository`, `StatusListCache`, `CertificateProvider`, `StatusListHistoryRepository`), typed entirely on domain values.
- **`src/application`** — inbound use cases (`PublishStatusList`, `UpdateStatuses`, `GetStatusListToken`, `*WithHistory` variants) owning cross-entity invariants: existence checks, issuer ownership, serialized-size bound, snapshot recording, and the optimistic-concurrency stamp.
- **`src/adapters`** — one module per integration: `sea_orm` (Postgres/MySQL/SQLite), `cache` (Moka), `certificate` (ACME bridge), `aws`/`redis` (cert-manager storage), `memory` (unit tests + `memory-only` build).
- **Composition root** — `utils::state::build_state` is the only place adapters are constructed; `AppState` holds injected trait objects. Historical snapshots are composed in only when `history_retention_secs > 0`.

The retired `utils::lst_gen` encoder is deleted; its full test suite — including the IETF §4.1/§4.2 worked vectors — is ported onto `domain::StatusList` and passed byte-for-byte against the old encoder's pinned outputs **before** the deletion.

## Behavior delta vs. `main` (the footnote to "no functional changes")

**Exactly one code-behavior change:** a duplicate create that loses a check-then-insert race (concurrent publish, or a client retry after a timeout) now returns **409 Conflict** instead of 500. Unique-key violations are normalized via SeaORM's backend-independent `sql_err()` → `RepositoryError::DuplicateEntry` → `PortError::Conflict` → the existing `AlreadyExists` 409 path. Verified against real sqlite and real MySQL (container test), not just mocks.

Everything else is byte-identical by construction: the domain encoder carries the spec worked vectors pinned byte-for-byte (`eNrbuRgAAhcBXQ`, `eNo76fITAAPfAgc`), compression stays at level 9, ETag inputs are unchanged.

**Not a delta, but worth stating:** the optimistic-concurrency guard from #234 (`WHERE updated_at = expected` + strictly-advancing stamp, 409 on conflict) is preserved through the new port boundary as `StatusListRepository::update(record, expected_updated_at)`. The DB layer hard-errors on a non-advancing stamp, and the semantics are pinned at three layers — store (sqlite/mysql), use case (memory adapters), handler (mock, 409).

## Deliberate deviation from the ticket's deliverables list

The ticket lists `SecretStore`, `DnsProvider`, and `MetricsCollector` port traits. **They are intentionally not shipped.** Earlier iterations included them; they had zero call sites, forced a duplicated Route53 client, and one unread `AppState` field made startup fail on four supported DNS providers. Ports with no consumer are dead surface — #162 (DNS providers) and #155 (S2S auth) should introduce them together with their first consumer. This is a scope call, not an omission.

## Testing

- 244 lib tests on the default stack; **28 execute in the `memory-only` CI job** (`cargo test --lib --no-default-features --features memory-only`) — the infrastructure-free composition is tested, not just type-checked.
- Domain encoder: 22 tests including both spec vectors, 9/13-bit `ApplicationSpecific` layouts, reserved-value rejection on encode and decode, re-encode preservation across bit-width expansion, and empty-list create→update end-to-end.
- Concurrency: guard semantics pinned at store, use-case, and handler layers; the memory adapters implement the same CAS as SQL, so a blind-insert "simplification" fails unit tests.
- Real-backend duplicate-key detection (`SqlErr::UniqueConstraintViolation`) proven on sqlite locally; the MySQL container twin runs in the `cargo-nextest --all-features` job.

## Docs

`docs/hexagonal-architecture.md` restored (README link resolves again) with the dependency rule stated as a rule — domain depends on nothing — and corrected diagram arrows. `docs/architecture.md` limit-enforcement and 409 rows updated to match the code.

## Merge checklist

- [ ] CI green
- [ ] Confirm **by name** in the nextest log: `test_mysql_duplicate_insert_maps_to_duplicate_entry` executed and passed (B8 is verified-on-sqlite / asserted-on-MySQL until this runs)
- [ ] Confirm the `tests/cert_provisioning.rs` suite executed and passed (untouched by the closeout commits, but the merge touched cert-manager storage paths)
- [ ] Follow-up issue filed (corrupt-state error classification, publish limit placement, dead error impl)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/domain/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Issuer(pub String);
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Credential {
+    pub issuer: Issuer,
+    /// A serialized public JWK.  Parsing/verifying it belongs to the inbound
+    /// authentication adapter.
+    pub public_key: serde_json::Value,
```
Hermann-Core: According to the Hexagonal Architecture pattern, domain models should be pure and free of infrastructure concerns. Using `serde_json::Value` here leaks JSON serialization logic into the domain layer. Consider defining a domain-specific type (e.g., a byte array, a parsed domain structure, or a custom `JWK` enum) and mapping to/from `serde_json::Value` at the application/adapter boundaries.

src/ports/mod.rs
```diff
@@ -0,0 +1,56 @@
+//! Outbound ports. Application services depend on these traits, never on an
+//! infrastructure client or framework.
+use async_trait::async_trait;
+
+use crate::domain::{Credential, StatusListRecord};
+
+#[derive(Debug, thiserror::Error)]
+pub enum PortError {
+    #[error("outbound dependency failed: {0}")]
+    Dependency(String),
```
Hermann-Core: Using a generic `Dependency(String)` error for all outbound port failures can make it difficult for the application layer to react to specific failure modes (e.g., distinguishing between a network timeout, an authentication failure, or a constraint violation). Hexagonal architecture encourages domain-specific error types that describe what went wrong in domain terms rather than adapter terms.

src/utils/state.rs
```diff
@@ -29,17 +35,28 @@ fn empty_to_none(value: Option<String>) -> Option<String> {
 
 #[derive(Clone)]
 pub struct AppState {
```
Hermann-Core, line 58: Injecting all raw repositories directly into the `AppState` forces the HTTP handlers to construct application use cases (e.g. `PublishStatusList::new(...)`) themselves. As outlined in the Hexagonal Architecture article, this expands the testing surface area of handlers. It is more idiomatic to define a `Service` trait (or individual use case traits) as the inbound port, construct the service in `main.rs`, and inject the *Service* into the handlers via `AppState`. This way, handlers only depend on the inbound port, making them easier to test with a single mocked service rather than mocking every repository.

Hermann-Core (changes requested):
@Ngha-Boris I have reviewed the PR and posted comments for your attention.

Ngha-Boris:
Hello @Hermann-Core please check again

martcpp:
this pr should not be merge till Friday thanks

Ngha-Boris:
> this pr should not be merge till Friday thanks

What is the reason for that @martcpp?

martcpp:
> > this pr should not be merge till Friday thanks
> 
> What is the reason for that @martcpp?

it structural revamp there are other ticket that should be cloud to avoid serious conflict

Hermann-Core (changes requested):
..

Hermann-Core (changes requested):
..

src/domain/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    InvalidIndex,
+    #[error("{0}")]
+    InvalidStatusList(String),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Issuer(pub String);
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PublicJwk(pub Vec<u8>);
```
Hermann-Core, line 22: Good use of opaque bytes. Doc should clarify format: The `Vec<u8>` should contain valid UTF-8 JSON serialized JWK. Add validation in `new()` or make it `try_new() -> Result<Self, DomainError>` to enforce the invariant that these bytes represent a parseable JWK.

src/ports/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    Conflict {
+        resource: &'static str,
+        reason: String,
+    },
+    #[error("invalid data for {resource}: {reason}")]
+    InvalidData {
+        resource: &'static str,
+        reason: String,
+    },
+}
```
Hermann-Core, line 78: `PortError` variants are more specific than the original `Dependency(String)`, which is an improvement. However, they still carry opaque strings that make programmatic reaction difficult. For crypto implementations where failures must be handled precisely (e.g., expired certs, network timeouts vs permanent auth failures), consider:

1. **Timeout variant**: Good addition. Should `operation` be typed as an enum instead of `&'static str` for exhaustive matching?
2. **Unauthorized variant**: Does this distinguish between "wrong key" (retry won't help) vs "transient auth service failure" (retry might help)?
3. **InvalidData variant**: Should this be split into parsing errors (bad format, cannot deserialize) vs semantic errors (wrong algorithm, expired)?

The domain layer should be able to match on these without parsing strings.

src/utils/state.rs
```diff
@@ -37,12 +56,14 @@ fn acme_dns_credentials(account: &crate::config::AcmeDnsAccount) -> AcmeDnsCrede
```
Hermann-Core, line 56: Excellent - `AppState` now injects service traits (`Arc<dyn StatusListService>`, `Arc<dyn CredentialService>`) instead of raw repositories. This matches the hexagonal architecture principle that handlers should depend on inbound ports, not repositories.

**Minor**: `cert_manager: Arc<CertManager>` is still a concrete type. Should this be behind a trait too for consistency (e.g., `certificate_provider` is already abstracted)?

src/domain/mod.rs
```diff
@@ -0,0 +1,326 @@
+//! Framework-independent business concepts for status-list operations.
```
Hermann-Core, line 1: Domain invariants and business rules are implemented correctly, but the module lacks inline documentation. According to hexagonal architecture best practices, domain code should be self-documenting about invariants. Add doc comments for:

1. `StatusList::create` / `StatusList::update`: document the compression format (zlib + base64url) and bit-packing invariants
2. `status_value`: document the valid range for `ApplicationSpecific` (≥256) and why
3. `apply_updates`: document the bit-manipulation strategy (especially multi-byte spanning)

These are long-term architectural decisions that future maintainers need to understand without reading implementation.

src/adapters/sea_orm.rs
```diff
@@ -0,0 +1,177 @@
...
+
+#[async_trait]
+impl CredentialRepository for PostgresCredentialRepository {
+    async fn find(&self, issuer: &str) -> Result<Option<domain::Credential>, PortError> {
+        self.store
+            .find_one_by(issuer)
+            .await
+            .map(|record| {
+                record.map(|record| domain::Credential {
+                    issuer: domain::Issuer(record.issuer),
```
Hermann-Core, line 80: `.expect("JWK is serializable")` can panic in production. Since this is a crypto implementation, untrusted JWK data could cause a panic here if it contains values that serde_json cannot serialize (e.g., non-UTF-8, special values).

Replace with proper error handling:
```rust
serde_json::to_vec(&record.public_key).map_err(|e| {
    PortError::InvalidData {
        resource: "stored JWK",
        reason: format!("serialization failed: {e}"),
    }
})?
```

src/adapters/memory.rs
```diff
@@ -0,0 +1,252 @@
...
+                lst: "".into(),
+            },
+            sub: "https://example/id".into(),
+            updated_at: 1672531200,
+        }
+    }
+
+    #[tokio::test]
+    async fn application_services_work_without_infrastructure() {
+        let repo = Arc::new(MemoryStatusLists::default());
```
Hermann-Core, line 191: Good use of memory adapters for infrastructure-free testing. The test validates the happy path and duplicate detection, but consider adding negative test cases:

1. **UpdateStatuses with wrong issuer** (should return `IssuerMismatch`)
2. **GetStatusListToken for non-existent list** (should return `NotFound`)
3. **Domain-level errors** propagating through use cases (e.g., invalid status bits)

These would verify that `UseCaseError` variants map correctly and error boundaries are enforced.

Hermann-Core (changes requested):
The hexagonal architecture refactor addresses previous review feedback well:

✅ Domain uses opaque `Vec<u8>` instead of `serde_json::Value`
✅ Service trait injection in `AppState` (no raw repositories in handlers)
✅ Feature-gated composition enables infrastructure-free builds
✅ Memory adapters enable pure unit testing

Changes needed before merge:

1. **Domain docs**: Add doc comments for invariants (compression format, bit-packing, ApplicationSpecific ranges)
2. **Error specificity**: `PortError` variants still use opaque string `detail`. For crypto code, consider typed sub-errors
3. **Panic safety**: Remove `.expect()` in postgres adapter (line 47)
4. **Test coverage**: Add negative test cases (wrong issuer, not found, domain errors)

Minor: Consider abstracting `cert_manager` behind a trait for consistency with `certificate_provider`.

Overall solid architecture with clear separation of concerns.

Hermann-Core:
@Ngha-Boris Review comments have been posted. **Changes requested** before merge:

1. Add domain invariant documentation (compression, bit-packing, value ranges)
2. Remove panic in postgres adapter (`.expect()` → proper error handling)
3. Strengthen port error specificity for crypto failure modes
4. Add negative test cases for error paths

Action needed: Address inline comments and re-request review.

Ngha-Boris:
Hello @Hermann-Core please check this again

martcpp (changes requested):
- build_dns_updater hard-errors for cloudflare/azure/gcloud/acmedns, preventing startup on those deployments. Total outage on upgrade.
- updated_at not bumped on update → 304 Not Modified served for revoked credentials via If-Modified-Since. Revocation-freshness failure.
- history_retention_secs = 0 no longer disables snapshots, and the GC that would clean them is disabled in exactly that configuration. Privacy regression + unbounded table growth.
- compression level changed from best() to default(), altering every lst byte-string and invalidating all ETags on first update. Undeclared behavior change; ticket requires behavior preservation.
- Test migration  the live bit-packing implementation has no tests; the entire spec-vector suite validates the abandoned cfg(test)-only copy. Port the tests before merge.
- helm/chart/values.yaml — eight rate-limit/abuse-limit keys deleted, unrelated to this refactor. Restore.


- Collapse the WithHistory duplication into one service with Option<Arc<dyn …>>; delete StatusListApplicationService's lying NotFound/Ok(0) stubs.
- Move StatusListHistoryRecord into domain as a snapshot value object; remove SeaORM from the port signatures and the cfg(feature = "postgres") gates on domain concepts.
- Delete src/adapters/dns.rs; adapt the existing challenge providers instead of forking them.
- Delete unused PortError variants, PortOperation entries, and the three unread AppState fields (with their startup client construction). Reintroduce them in the tickets that need them.
- Return Arc<StatusListRecord> from StatusListCache::get / StatusListRepository::find to restore the zero-copy read path.
Remove the preview/double-compute in publish and update once limits move into the domain.
- Introduce a Clock port; remove SystemTime::now() / now_utc() from use cases and unify the two clock sources.
Map StatusListRepository::update's false to NotFound, or drop the bool.
- Map SeaORM unique-violations to PortError::Conflict in the adapters, and Conflict → 409 in the handlers.
Split DomainError::InvalidStatusList into request-caused (→400) and state-caused (→500) variants.
- Fix the mermaid arrows in docs/hexagonal-architecture.md (Application → Ports, Application → Domain, Domain → ∅); state the dependency rule as a rule, not just a picture; reconcile or retire docs/architecture.md.
- Restore the TTL=0 semantics doc comment on MokaStatusListCache and its test.
- Make memory-only a real feature or drop it; switch the CI job to cargo test.
- Cache the parsed DecodingKey per issuer to remove the triple JSON round-trip on auth.

Ngha-Boris:
@martcpp please check again

martcpp:
@Hermann-Core @Ngha-Boris @ndefokou pls check

Ngha-Boris:
This looks good to me @martcp

docs/hexagonal-architecture.md
Hermann-Core, line 1: We don't need to document this. There are many documents around internet for this.

src/adapters/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+pub mod aws;
+#[cfg(feature = "cache-moka")]
+pub mod cache;
+#[cfg(feature = "certificate-acme")]
+pub mod certificate;
+pub mod memory;
+#[cfg(feature = "redis")]
+pub mod redis;
+#[cfg(any(
+    feature = "server",
```
Hermann-Core, line 11: what is `server` feature means for?

src/adapters/sea_orm.rs
```diff
@@ -0,0 +1,252 @@
...
+            detail: e.to_string(),
+        },
+    }
+}
+
+#[derive(Clone)]
+pub struct SeaOrmStatusListRepository {
+    store: SeaOrmStore<models::StatusListRecord>,
+}
+impl SeaOrmStatusListRepository {
```
Hermann-Core, line 36: Can you add a space between these blocs?

src/adapters/sea_orm.rs
```diff
@@ -0,0 +1,252 @@
...
+        RepositoryError::DuplicateEntry => PortError::Conflict {
+            resource,
+            reason: "already exists".to_string(),
+        },
+        e => PortError::StorageUnavailable {
+            operation,
+            detail: e.to_string(),
+        },
+    }
+}
```
Hermann-Core, line 30: Helpers functions should be placed below public items.

src/adapters/sea_orm.rs
Hermann-Core, line 1: Can you add a space between items across this file?

src/database/queries.rs
Hermann-Core, line 1: We still have database/queries layer which contradict hexagonal arch principle.

martcpp (approved):

src/application/mod.rs
```diff
@@ -0,0 +1,647 @@
...
+    feature = "postgres",
+    feature = "sqlite",
+    feature = "mysql"
+))]
+use crate::domain::StatusListSnapshot;
+#[cfg(any(
+    feature = "server",
+    feature = "postgres",
+    feature = "sqlite",
+    feature = "mysql"
```
ndefokou, line 18: **Missing Clock port abstraction**: `current_unix_timestamp()` calls `SystemTime::now()` directly at line 18-22. If a `Clock` trait is introduced in a future ticket (#160 observability), this becomes the single call site to update. Either document this as intentionally deferred, or provide a `Clock` trait here. Currently acceptable as-is if deferral is intentional.

src/ports/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            Self::CertificateChain => "certificate chain",
+            Self::DeleteOldStatusListHistory => "delete old status list history",
+            Self::FindCredential => "find credential",
+            Self::FindStatusList => "find status list",
+            Self::FindStatusListHistory => "find valid status list history",
+            Self::InsertCredential => "insert credential",
+            Self::InsertStatusList => "insert status list",
+            Self::InsertStatusListHistory => "insert status list history",
+            Self::ListStatusListUris => "list status list URIs",
+            Self::SigningKey => "signing key",
```
ndefokou, line 42: **PortError::detail is still a String**: While `PortOperation` is now typed as an enum (improvement), `detail: String` (line 47, 51, 64) makes programmatic error handling difficult. For example, there's no sub-error typed distinction between "network unreachable" (retryable) vs "invalid response format" (permanent failure). Consider typed sub-errors if future consumers need different retry strategies.

src/utils/state.rs
```diff
@@ -40,12 +56,10 @@ fn acme_dns_credentials(account: &crate::config::AcmeDnsAccount) -> AcmeDnsCrede
 
 #[derive(Clone)]
 pub struct AppState {
-    pub credential_repo: SeaOrmStore<Credentials>,
-    pub status_list_repo: SeaOrmStore<StatusListRecord>,
-    pub status_list_history_repo: SeaOrmStore<StatusListHistoryRecord>,
+    pub status_lists: Arc<dyn StatusListService>,
+    pub credentials: Arc<dyn CredentialService>,
+    pub certificate_provider: Arc<dyn CertificateProvider>,
```
ndefokou, line 61: **Inconsistent abstraction for cert_manager**: `cert_manager: Arc<CertManager>` (line 56) is a concrete type while `status_lists`, `credentials`, and `certificate_provider` all use trait objects via `Arc<dyn ...>`. This is inconsistent with the hexagonal principle. PR rationale states this was intentional scope; recommend documenting that when #159 introduces more cert_manager consumers, it should be abstracted behind a trait.

src/adapters/memory.rs
```diff
@@ -0,0 +1,471 @@
...
+            .unwrap();
+        assert_eq!(fetched.list_id, "id");
+        UpdateStatuses::new(repo, cache.clone())
+            .execute(&Issuer("issuer".into()), "id", Vec::new())
+            .await
+            .unwrap();
+        assert!(cache.get("id").await.unwrap().is_none());
+    }
+
+    #[cfg(any(
```
ndefokou, line 213: **Bounded cache staleness documented but needs sign-off**: The test at line 213 documents that `GetStatusListToken` provides C2 eventual consistency (read-through cache does NOT revalidate on read). This is deliberate per the test comment. If strong read-your-writes consistency is required for revocation freshness, the getter would need to invalidate cache entry on every write. Team should explicitly confirm eventual consistency is acceptable for their verification use cases.

ndefokou (commented):
## Additional Review: Architecture & Design Considerations

These are inline observations from my review of PR #212 against ticket #186.

Hermann-Core (dismissed):
Hello @Ngha-Boris, this need a lot of refactoring to be fully switched to hexagonal architecture principle. I left a couple of comments below.

src/domain/mod.rs
```diff
@@ -0,0 +1,734 @@
...
+    /// Application-specific values start at 256. Values 3..=255 are rejected by
+    /// deliberate *server policy* (not a spec requirement): it keeps the
+    /// standard 1-/2-bit lists limited to the three standard statuses and
+    /// sidesteps ambiguity with the compact bit-width table. This threshold is
+    /// pinned by test and treated as spec-evolution surface, not a knob to
+    /// loosen as part of unrelated cleanup.
+    ApplicationSpecific(u32),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
ndefokou, line 68: **Compression level comment missing**: `encode_compressed` at line 68-74 uses `flate2::Compression::best()` but the doc comment does not explicitly state this. For reproducibility and auditing, consider documenting: 'Uses zlib compression at best level (level 9). Spec mandates zlib (RFC 1950), not raw deflate or gzip.' This is critical for the byte-for-byte reproducibility requirement.

src/domain/mod.rs
```diff
@@ -0,0 +1,734 @@
...
+    pub fn try_new(bytes: Vec<u8>) -> Result<Self, DomainError> {
+        let value: serde_json::Value = serde_json::from_slice(&bytes)
+            .map_err(|err| DomainError::InvalidPublicJwk(format!("expected UTF-8 JSON: {err}")))?;
+        if !value.is_object() {
+            return Err(DomainError::InvalidPublicJwk(
+                "expected a JSON object".to_string(),
+            ));
+        }
+        Ok(Self(bytes))
+    }
```
ndefokou, line 39: **ApplicationSpecific doc could be clearer**: The doc comment at line 39-45 explains the >= 256 policy well, but consider adding: 'The encoding table reserves values 0-255, where 3-255 are invalid. This server enforces this at the domain layer, not accepting values 3-255 at decode time.' This makes it clear that decode of reserved values is a corruption indicator, not a user error.

src/adapters/sea_orm.rs
```diff
@@ -0,0 +1,252 @@
...
+            bits: record.status_list.bits,
+            lst: record.status_list.lst,
+        },
+        sub: record.sub,
+        iat: record.iat,
+        exp: record.exp,
+    }
+}
+
+fn snapshot_to_persistence(record: domain::StatusListSnapshot) -> models::StatusListHistoryRecord {
```
ndefokou, line 152: **Optimistic concurrency implementation verified**: The update path at line 152-161 correctly uses `WHERE updated_at = expected_updated_at` guard semantics. The return value `bool` from `update_one` correctly maps to `Result<bool, PortError>` with `false` indicating the guard mismatched (concurrent race lost). Good pattern.

ndefokou (commented):
## Domain Documentation Review

src/domain/mod.rs
```diff
@@ -0,0 +1,734 @@
...
+    ///
+    /// The domain stores `lst` using the Status List Token format: status
+    /// values are bit-packed least-significant-bit first, compressed with zlib,
+    /// then base64url encoded without padding. The `bits` width is the minimum
+    /// required by the highest status value, except that `ApplicationSpecific`
+    /// values may require more than eight bits.
+    pub fn create(status_updates: Vec<StatusEntry>) -> Result<Self, DomainError> {
+        if status_updates.is_empty() {
+            return Ok(Self {
+                bits: 1,
```
ndefokou, line 332: **Spec vector tests verified**: The IETF §4.1/§4.2 worked vectors at lines 332-340 (`'eNrbuRgAAhcBXQ'`) and lines 347-351 (`'eNo76fITAAPfAgc'`) are correctly pinned on the live encoder. These byte-level pins satisfy the acceptance criterion 'New unit tests pass' and the ticket requirement that the retired encoder's test suite was ported to the domain layer.

Cargo.toml
```diff
@@ -3,6 +3,45 @@ name = "status-list-server"
 version = "0.1.0"
 edition = "2024"
 
+[features]
+default = ["server"]
+memory-only = []
```
ndefokou, line 8: **Feature flag structure verified**: The `server` feature (line 6) aggregates `aws`, `cache-moka`, `certificate-acme`, `http`, `postgres`, `redis`. The `memory-only` feature (line 5) is correctly empty. The CI workflow added in `.github/workflows/CI.yml` runs `cargo test --lib --no-default-features --features memory-only` verifying the infra-free composition. The binary target at the bottom of Cargo.toml correctly requires `required-features = ["server"]` preventing `memory-only` builds from producing a server binary.

ndefokou (commented):
## Spec Vector Testing & Feature Flags Review

ndefokou (commented):
## Error Handling & Concurrency Review

src/application/mod.rs
```diff
@@ -0,0 +1,647 @@
...
+/// writer created the resource between our existence check and the insert —
+/// semantically the same "already exists" outcome as the pre-check, so a
+/// racing publish surfaces as 409, not 500. (`UseCaseError::Conflict` is
+/// reserved for the optimistic-concurrency update race.)
+fn map_insert_conflict(error: PortError) -> UseCaseError {
+    match error {
+        PortError::Conflict { .. } => UseCaseError::AlreadyExists,
+        other => UseCaseError::Port(other),
+    }
+}
```
ndefokou, line 43: **409 Conflict for duplicate-insert race**: The `map_insert_conflict` function at line 43-50 correctly maps `PortError::Conflict` to `UseCaseError::AlreadyExists`, converting a database unique-key violation (concurrent check-then-insert race) to HTTP 409. This is the 'behavior delta from main' that the PR description correctly identifies as intentional - a race that was previously 500 is now 409. Confirm this is the intended behavior change.

src/application/mod.rs
```diff
@@ -0,0 +1,647 @@
...
+) -> Result<(), UseCaseError> {
+    // The snapshot becomes valid at the record's modification time. Reusing
+    // `record.updated_at` — rather than reading the clock a second time here —
+    // threads a single `now` through the write and its snapshot, so their
+    // timestamps cannot drift apart (C4).
+    let iat = record.updated_at;
+    let snapshot = StatusListSnapshot {
+        snapshot_id: uuid::Uuid::new_v4().to_string(),
+        list_id: record.list_id.clone(),
+        issuer: record.issuer.clone(),
```
ndefokou, line 238: **Conflict handling correctly returns before side effects**: In `UpdateStatuses::execute` at line 238, the error return happens BEFORE cache invalidation (line 249) and history recording, ensuring a rejected write (concurrent race lost) leaves no trace. The memory adapter tests (e.g., `update_statuses_returns_conflict_when_write_races` in adapters/memory.rs) verify this behavior. Good atomicity pattern.

docs/hexagonal-architecture.md
```diff
@@ -0,0 +1,118 @@
...
+
+```mermaid
+flowchart LR
+    Web["Inbound adapter<br/>src/web (Axum handlers)"]
+    App["Application<br/>src/application (use cases)"]
+    Domain["Domain<br/>src/domain (entities, encoder invariants)"]
+    Ports["Ports<br/>src/ports (outbound traits)"]
+    Adapters["Outbound adapters<br/>src/adapters (SeaORM, Moka, ACME, memory)"]
+
+    Web --> App
```
ndefokou, line 30: **Mermaid diagram verified**: The dependency rule is correctly stated at lines 1-19 and the mermaid diagram at lines 21-36 shows correct arrow directions (Adapters --> Ports, Ports --> Domain). The dependency rule ('Domain points at nothing') is correctly documented as a rule, not just a picture. This satisfies ticket requirement 'Update architecture document explaining layers and dependency rule'.

docs/hexagonal-architecture.md
martcpp: it a tracker for the the fellow up fin clean up pr will delete it https://github.com/adorsys/status-list-server/issues/245

ndefokou (commented):
@Ngha-Boris Can you check those small comments i left? 
Just small remark but not critical

src/adapters/mod.rs
martcpp: The server feature aggregates aws, cache-moka, certificate-acme, http, postgres, redis. The memory-only feature  is correctly empty. The CI workflow added in .github/workflows/CI.yml runs cargo test --lib --no-default-features --features memory-only verifying the infra-free composition. The binary target at the bottom of Cargo.toml correctly requires required-features = ["server"] preventing memory-only builds from producing a server binary.

src/adapters/sea_orm.rs
martcpp: let this review be focus on the isolation of layer on not correct this post hexal ticket that will track and fix all this to keep it simple and fast thanks

src/adapters/sea_orm.rs
martcpp: same as here thanks

src/domain/mod.rs
martcpp: this his out of scope for this ticket

src/application/mod.rs
martcpp: this his track on fellow issue https://github.com/adorsys/status-list-server/issues/245

ndefokou (approved):

PR #214, opened 2026-07-14
175 implement http caching headers for status list token retrieval

### LINKED ISSUE #175: Implement HTTP caching headers for status-list token retrieval
## Objective

Add `Cache-Control`, `ETag`, `Last-Modified`, and conditional request support for `GET /statuslists/{list_id}`.

## Context

The architecture review notes that tokens include `ttl`, but the HTTP response does not appear to set `Cache-Control`, `ETag`, or conditional request support. Relying parties and CDNs need clear freshness behavior.

## Deliverables

- [ ] Generate ETag from list content/version/timestamp.
- [ ] Return `Cache-Control: max-age=<ttl>, immutable`.
- [ ] Support `If-None-Match` 304.
- [ ] Ensure gzip variant handled.

## Acceptance Criteria

- Conditional requests return `304` when content is unchanged.
- Unconditional requests return `200` with correct headers.
- Cache headers are present for both JWT and CWT responses.

## References

- Related: #145 (REST redesign), #156 (error handling)

ndefokou:
Implements HTTP caching and conditional request semantics for the status list retrieval endpoint so relying parties and CDNs can avoid redundant re-fetches when content has not changed. Both JWT and CWT response variants are covered.

### Implementation

The work is split across four modules under `src/web/handlers/status_list/`:

**ETag generation** (`etag.rs:35`)
- `generate_etag` computes a weak ETag (`W/"<sha256_hex>"`) from the stable content-identity fields of a `StatusListRecord`: `bits`, `lst` (compressed bitstring), `issuer`, and `sub`. The hash intentionally excludes `list_id` and `updated_at` so the ETag reflects content equivalence, not metadata.

**Conditional request evaluation** (`conditional.rs`)
- `evaluate_if_none_match` — parses comma-separated ETags and the `*` wildcard per RFC 7232 §3.2, returning `NotModified` on any match.
- `evaluate_if_modified_since` — parses the HTTP-date, validates it is not in the future, and returns `NotModified` when the resource's `updated_at` is at or before the client's timestamp (RFC 7232 §3.3).
- `evaluate_conditional_request` — applies RFC 7232 §6 precedence: `If-None-Match` takes priority over `If-Modified-Since` when both are present.
- `format_http_date` / `parse_http_date` — convert between Unix timestamps and RFC 2822 / HTTP-date format for `Last-Modified` header emission and `If-Modified-Since` parsing.

**Handler integration** (`get_status_list.rs:34`)
- The handler now extracts `If-None-Match` and `If-Modified-Since` headers from the request, fetches the record (via cache or DB), computes the ETag and `Last-Modified` date, and builds a `Cache-Control: public, max-age=<ttl>, immutable` header from the configured token TTL.
- On a conditional hit, it returns `304 Not Modified` with `ETag`, `Last-Modified`, and `Cache-Control` headers but no body.
- On a conditional miss or unconditional request, it returns `200 OK` with `Content-Type`, `Content-Encoding: gzip`, `ETag`, `Last-Modified`, and `Cache-Control` headers alongside the gzip-compressed token body.
- The ETag and `Last-Modified` are computed from the record before token issuance, so the 304 path skips the expensive signing/compression work entirely.

**Error responses** (`error.rs:42`)
- All error responses now carry `Cache-Control: no-store, max-age=0` and intentionally omit `ETag` and `Last-Modified` so error states are never cached.

### Gzip variant handling

Every successful response is gzip-compressed via `flate2::GzEncoder` and carries `Content-Encoding: gzip`. Because the ETag is derived from the logical record content (not the compressed bytes), it is stable across both JWT and CWT representations and independent of compression. The 304 path returns no body, so no compression negotiation is needed there.

### Acceptance criteria

| Criterion | Status |
|---|---|
| Conditional requests return 304 when content is unchanged | ✅ `test_conditional_request_with_matching_etag` verifies 304 with empty body and caching headers still present |
| Unconditional requests return 200 with correct headers | ✅ `test_get_status_list_includes_caching_headers` verifies `ETag`, `Last-Modified`, and `Cache-Control: public, max-age=300, immutable` |
| Cache headers present for JWT responses | ✅ All JWT tests assert the headers; `test_get_status_list_jwt_success` confirms 200 + gzip |
| Cache headers present for CWT responses | ✅ `test_get_status_list_success_cwt` confirms 200 + gzip; headers emitted via the same code path |
| Error responses are not cached | ✅ `test_get_status_list_not_found` and `test_error_responses_omit_etag_and_last_modified` assert `no-store` and absence of `ETag`/`Last-Modified` |

### Test coverage

New and updated unit tests:
- `conditional.rs` — 18 tests covering `If-None-Match` (single, multiple, wildcard, malformed, absent), `If-Modified-Since` (equal, older, newer, malformed, absent), precedence rules, and HTTP-date round-trip parsing.
- `etag.rs` — 9 tests covering format, determinism, field sensitivity (`bits`, `lst`, `issuer`, `sub`), field independence (`list_id`, `updated_at`), empty strings, and special characters.
- `get_status_list.rs` — integration tests for caching headers on 200, 304 with matching ETag, header presence on both JWT and CWT paths, and error-response header assertions.


src/web/handlers/status_list/get_status_list.rs
```diff
@@ -36,33 +38,86 @@ pub async fn get_status_list(
...
+        .and_then(|h| h.to_str().ok());
+    let if_modified_since = headers
+        .get(header::IF_MODIFIED_SINCE)
+        .and_then(|h| h.to_str().ok());
+
+    // Fetch status list record (from cache or database)
+    let status_record = fetch_status_record(&list_id, &state).await?;
+
+    // Generate ETag from record content
+    let current_etag = generate_etag(&status_record);
```
Christiantyemele, line 65: This validator is derived only from the stored status-list record, but the selected representation returned by the 200 path also includes freshly generated `iat`/`exp` claims in `issue_jwt`/`issue_cwt`. After `Cache-Control: max-age` expires, a cache can revalidate with this same ETag, receive 304 because the list content did not change, and keep extending freshness for the old body until it is replaying an expired signed token. RFC 9110 defines `If-None-Match` against the selected representation, so the validator needs to change before the signed representation becomes invalid, or the server needs to validate/cache the actual signed representation rather than only the logical list content. One fix is to include a token validity bucket/expiry (and any signing material that changes the body) in the ETag and stop returning 304 once the cached representation would be expired.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -36,33 +38,86 @@ pub async fn get_status_list(
...
+                    (header::CACHE_CONTROL, cache_control.as_str()),
+                ],
+            )
+                .into_response())
+        }
+        ConditionalResponse::Modified => {
+            // Build full token response
+            let compressed_token = build_token(accept_type, &status_record, &state).await?;
+
+            Ok((
```
Christiantyemele: The response body and `Content-Type` are selected from the request `Accept` header, but neither the 200 nor the 304 response includes `Vary: Accept`. Because these responses are explicitly public-cacheable, a shared cache is allowed to reuse a JWT response for a later CWT request (or apply a 304 to the wrong stored variant) unless `Accept` is part of the cache key. Please emit `Vary: Accept` on both the full response and the 304 response.

src/web/handlers/status_list/conditional.rs
```diff
@@ -0,0 +1,368 @@
...
+/// * Formatted HTTP-date string in RFC 2822 format
+pub fn format_http_date(unix_timestamp: i64) -> String {
+    use time::OffsetDateTime;
+
+    let datetime =
+        OffsetDateTime::from_unix_timestamp(unix_timestamp).unwrap_or(OffsetDateTime::UNIX_EPOCH);
+
+    // Format as RFC 2822 (HTTP-date compatible)
+    datetime
+        .format(&time::format_description::well_known::Rfc2822)
```
Christiantyemele: `Rfc2822` formats UTC as `+0000` here, and the new test even allows that, but HTTP-date senders must generate IMF-fixdate with the literal `GMT` zone, for example `Sun, 06 Nov 1994 08:49:37 GMT` (RFC 9110 section 5.6.7). Emitting a non-HTTP-date `Last-Modified` value can make strict caches/clients ignore the validator or skip `If-Modified-Since` revalidation. Please format `Last-Modified` with an IMF-fixdate formatter and add a test that rejects the `+0000` form for emitted headers.

Christiantyemele (changes requested):
Thanks for pushing the caching path through both JWT and CWT responses. The package checks are green locally, but I found three HTTP caching issues that need changes before this is safe to merge: the validators can keep expired signed tokens alive via repeated 304s, negotiated JWT/CWT responses are cacheable without `Vary: Accept`, and `Last-Modified` is emitted in an RFC 2822 form instead of an HTTP-date. I verified the relevant behavior against RFC 9110 and ran `cargo check --package status-list-server`, `cargo test --package status-list-server`, and `cargo clippy --all-targets --package status-list-server -- -D warnings`, all of which passed.

src/web/handlers/status_list/conditional.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    // Check for future dates (client clock skew or malicious input)
+    let now = time::OffsetDateTime::now_utc().unix_timestamp();
+    if client_timestamp > now {
+        warn!("If-Modified-Since contains future date: {}", header_value);
+        return ConditionalResponse::Modified;
+    }
+
+    // Return NotModified if resource hasn't been updated since client's cached version
+    if updated_at <= client_timestamp {
```
Christiantyemele, line 132: **[Important] The `If-Modified-Since` path bypasses the validity-bucket freshness bound.**

The validity-bucket fix correctly bounds the **If-None-Match** path — the ETag rotates every `token_exp_secs`, so a `200` re-issue is forced once the bucket flips. That protection does not extend here. `evaluate_if_modified_since` compares only against `updated_at`, which is bumped *only* by `publish_status`/`update_status`, never by token re-issuance. So a cache/client that revalidates with `If-Modified-Since` and **omits `If-None-Match`** (RFC 9110 §13.1.3 only mandates ignoring IMS *when INM is present*; IMS-alone is valid and used by some proxies/CLI tools) keeps getting `304` for as long as `updated_at` is unchanged — i.e. indefinitely after the last status update.

Concrete scenario: a list published at `t0` with `token_exp_secs = 600` yields a token body with `exp = t0 + 600`. After expiry, a CDN revalidating with `If-Modified-Since: <original Last-Modified>` (no ETag) still gets `304`, so it keeps serving the now-expired signed token. That defeats the `ttl`/`exp` bound the Status List spec uses to cap staleness of revocation state — a security-relevant gap, since revoked credentials could appear valid past their intended freshness window.

Root cause: `Last-Modified` is derived from the logical list's `updated_at` (get_status_list.rs:69) but the served *representation* (re-signed token) changes every bucket even when the list content does not.

Suggested fix — make `Last-Modified` rotate with the validity bucket so the IMS validator tracks the representation actually served:
```rust
// get_status_list.rs
let bucket_start = (now / state.token_exp_secs.max(1) as i64) * state.token_exp_secs.max(1) as i64;
let last_modified = format_http_date(bucket_start);
// ... and compare bucket_start (not updated_at) inside evaluate_if_modified_since
```
Then a client caching `Last-Modified = bucket_start` and sending it back gets `304` for the current bucket and `200` once the bucket advances. Alternatively, drop `Last-Modified`/`If-Modified-Since` support for this time-varying resource and rely on the ETag alone.

src/web/handlers/status_list/conditional.rs
```diff
@@ -0,0 +1,400 @@
...
+            continue;
+        }
+
+        // Compare directly - handles both W/"hash" and "hash" formats
+        if etag == current_etag {
+            return ConditionalResponse::NotModified;
+        }
+
+        // Also try removing quotes for loose matching (some clients may send malformed ETags)
+        let normalized_etag = etag.trim_matches('"');
```
Christiantyemele: **[Nit] The "loose matching" branch is non-spec and redundant.**

`trim_matches('"')` strips quotes from both ends but not the `W/` weakness prefix, so it does not implement RFC 9110 §8.8.3.2 weak comparison — a strong `"abc"` would still not match weak `W/"abc"`. In practice clients echo the exact ETag they received, so the direct compare on line 57 already handles the common case and this branch never adds a *correct* match; it can only produce surprising equalities for malformed input. Either drop the loose branch (rely on exact compare + `*` wildcard) or implement proper weak/strong comparison: strip an optional leading `W/`, then compare the quoted opaque-tags.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -890,10 +978,255 @@ mod tests {
...
+        let vary = response_headers
+            .get(http::header::VARY)
+            .unwrap()
+            .to_str()
+            .unwrap();
+        assert_eq!(vary, "Accept");
+    }
+
+    #[tokio::test]
+    async fn test_conditional_request_with_matching_etag() {
```
Christiantyemele, line 1322: **[Suggestion] Add a handler-level `If-Modified-Since` test.**

The integration tests cover the `If-None-Match` 304 path (`test_conditional_request_with_matching_etag`) and the 200 header set, but there is no handler-level test exercising `If-Modified-Since` alone (304 when `updated_at <= client_time`, 200 otherwise), nor one asserting the validator rotates across a validity-bucket boundary. Adding an IMS-only 304/200 test (and, if the Important finding above is fixed by bucketing `Last-Modified`, one that advances the bucket to force a `200`) would lock in the staleness bound and guard this gap.

src/database/mod.rs
```diff
@@ -187,4 +190,53 @@ pub(crate) mod migrations {
...
+            async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+                manager
+                    .alter_table(
+                        Table::alter()
+                            .table(StatusLists::Table)
+                            .add_column(
+                                ColumnDef::new(StatusLists::UpdatedAt)
+                                    .big_integer()
+                                    .not_null()
+                                    .default(0),
```
Christiantyemele, line 215: **[Nit] `updated_at` defaults to 0, so legacy rows report `Last-Modified` = 1970-01-01.**

After this `ALTER TABLE ... ADD COLUMN ... default(0)`, every pre-existing `status_lists` row gets `updated_at = 0`. For those rows any `If-Modified-Since` date ≥ 1970 yields `NotModified`, so fresh tokens are never served via the IMS path until the first `update_status` touches them. Consider backfilling existing rows to the migration run time (e.g. `UPDATE status_lists SET updated_at = <now>` in the `up` migration, or `default(extract(epoch from now())::bigint)` for Postgres) so the validator is meaningful immediately.

Christiantyemele (changes requested):
Reviewed the caching/conditional-request implementation on the latest head (a19d79d). The three issues from the prior review are fully resolved: `Vary: Accept` is now emitted on both 200 and 304, `Last-Modified` uses IMF-fixdate (`GMT`, not RFC 2822 `+0000`), and the ETag includes a validity bucket so the If-None-Match path can no longer keep an expired signed token alive via repeated 304s. `cargo check --package status-list-server`, `cargo test --package status-list-server` (118 passed), and `cargo clippy --all-targets --package status-list-server -- -D warnings` are all green.

One Important finding blocks merge: the freshness bound only holds on the If-None-Match path. `If-Modified-Since` is evaluated against `updated_at` alone, which token re-issuance never bumps, so a client/CDN that revalidates without an `If-None-Match` header can receive `304` indefinitely and keep replaying an expired signed token — defeating the `ttl`/`exp` staleness bound the Status List spec relies on for revocation freshness. Plus one Suggestion (handler-level If-Modified-Since test) and two Nits (migration `updated_at` default 0; loose ETag matching). Counts: 1 Important, 1 Suggestion, 2 Nits.

src/web/handlers/status_list/etag.rs
```diff
@@ -0,0 +1,252 @@
...
+///         lst: "eNrbuRgAAhcBXQ".to_string(),
+///     },
+///     sub: "https://example.com/credentials/status/3".to_string(),
+///     updated_at: 1234567890,
+/// };
+///
+/// let etag = generate_etag(&record, 42);
+/// assert!(etag.starts_with("W/\""));
+/// assert!(etag.ends_with('"'));
+/// ```
```
Ngha-Boris, line 40: > **Issue: ETag includes validity_bucket causing unnecessary cache invalidation**
>
> The ETag is generated with `validity_bucket` as input, which changes every `token_exp_secs` interval. This causes the ETag to change even when the list content (`bits`, `lst`, `issuer`, `sub`) is identical, defeating the purpose of content-based ETags.
>
> **Suggestion:** Remove `validity_bucket` from ETag calculation. The ETag should represent content identity only:
> ```rust
> pub fn generate_etag(record: &StatusListRecord) -> String {
>     let mut hasher = Sha256::new();
>     hasher.update(record.status_list.bits.to_string().as_bytes());
>     hasher.update(record.status_list.lst.as_bytes());
>     hasher.update(record.issuer.as_bytes());
>     hasher.update(record.sub.as_bytes());
>     // Remove: hasher.update(validity_bucket.to_le_bytes());
>     let hash = hasher.finalize();
>     format!("W/\"{}\"", hex::encode(hash))
> }
> ```

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -36,33 +38,99 @@ pub async fn get_status_list(
 ) -> Result<impl IntoResponse + Debug + use<>, StatusListError> {
     let accept = headers.get(header::ACCEPT).and_then(|h| h.to_str().ok());
 
-    // build the token depending on the accept header
-    match accept {
-        None =>
-        // assume jwt by default if no accept header is provided
-        {
-            build_status_list_token(ACCEPT_STATUS_LISTS_HEADER_JWT, &list_id, &state).await
```
Ngha-Boris: > **Issue: Last-Modified rotates with validity bucket**
>
> Using `max(updated_at, bucket_start)` causes `Last-Modified` to artificially rotate with the validity bucket. This means `If-Modified-Since` requests will never match after a bucket rotation, forcing 200 responses when 304 would be appropriate for unchanged content.
>
> **Suggestion:** Use `updated_at` directly:
> ```rust
> let last_modified_ts = status_record.updated_at;
> ```
>
> If the goal is to ensure clients refresh expired tokens, consider using a shorter `max-age` instead of manipulating `Last-Modified`.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -36,33 +38,99 @@ pub async fn get_status_list(
...
+        None => ACCEPT_STATUS_LISTS_HEADER_JWT, // Default to JWT
         Some(accept)
             if accept == ACCEPT_STATUS_LISTS_HEADER_JWT
                 || accept == ACCEPT_STATUS_LISTS_HEADER_CWT =>
         {
-            build_status_list_token(accept, &list_id, &state).await
+            accept
+        }
+        Some(_) => return Err(StatusListError::InvalidAcceptHeader),
+    };
```
Ngha-Boris, line 52: > **Issue: Gzip compression is unconditional**
>
> The handler always returns `Content-Encoding: gzip` without checking the client's `Accept-Encoding` header. This violates HTTP semantics and may cause issues with clients that don't support gzip.
>
> **Suggestion:** Check `Accept-Encoding` before applying compression:
> ```rust
> let accepts_gzip = headers
>     .get(header::ACCEPT_ENCODING)
>     .and_then(|h| h.to_str().ok())
>     .map(|enc| enc.contains("gzip"))
>     .unwrap_or(false);
>
> let (body, encoding) = if accepts_gzip {
>     (compress_gzip(&token_bytes)?, Some(GZIP_HEADER))
> } else {
>     (token_bytes, None)
> };
> ```

src/web/handlers/status_list/conditional.rs
```diff
@@ -0,0 +1,401 @@
...
+) -> ConditionalResponse {
+    let Some(header_value) = if_none_match else {
+        return ConditionalResponse::Modified;
+    };
+
+    // Check for wildcard "*" - matches any representation
+    let trimmed = header_value.trim();
+    if trimmed == "*" {
+        return ConditionalResponse::NotModified;
+    }
```
Ngha-Boris, line 67: > **Suggestion: Consider weak ETag comparison**
>
> The current implementation uses exact string matching for ETags. If a client sends a strong ETag (without `W/` prefix) that semantically matches the weak ETag, it won't be recognized.
>
> **Suggestion:** Consider implementing weak comparison per RFC 9110 §8.8.3.2:
> ```rust
> fn normalize_etag(etag: &str) -> String {
>     etag.trim()
>         .trim_start_matches("W/")
>         .trim_matches('"')
>         .to_string()
> }
> 
> // Then compare normalized versions
> if normalize_etag(etag) == normalize_etag(current_etag) {
>     return ConditionalResponse::NotModified;
> }
> ```

src/web/handlers/status_list/error.rs
```diff
@@ -1,6 +1,8 @@
 use axum::{http::StatusCode, response::IntoResponse};
 use thiserror::Error;
```
Ngha-Boris, line 3: > **Issue: Cyclic dependency between error.rs and get_status_list.rs**
>
> The error module imports `build_error_cache_control` from the handler module, creating a cross-dependency that complicates the module structure.
>
> **Suggestion:** Move the constant to a shared location (e.g., `constants.rs`):
> ```rust
> // In constants.rs
> pub const ERROR_CACHE_CONTROL: &str = "no-store, max-age=0";
> ```

src/web/handlers/status_list/etag.rs
```diff
@@ -0,0 +1,252 @@
+use crate::models::StatusListRecord;
+use sha2::{Digest, Sha256};
+
+/// Computes a weak ETag from status list content and a validity bucket.
+///
+/// The ETag is computed from the concatenation of:
+/// - bits (as string)
+/// - lst (base64url-encoded compressed bitstring)
+/// - issuer (string)
```
Ngha-Boris, line 15: > **Issue: Docstring doesn't match actual behavior**
>
> The docstring claims the ETag is computed from content fields, but the actual implementation includes `validity_bucket` which is not content but a time-based rotation mechanism.
>
> **Suggestion:** Update the documentation to accurately reflect that `validity_bucket` is included in the hash, or (preferred) remove `validity_bucket` from ETag calculation as noted in the comment on line 35.

Ngha-Boris (commented):
## Code Review: HTTP Caching Headers Implementation

Thank you for this comprehensive implementation. While the overall structure is solid and test coverage is good, I've identified several issues that need to be addressed before merging.

### Critical Issues (Must Fix)

1. **ETag includes validity_bucket causing unnecessary cache invalidation** - The ETag changes every `token_exp_secs` even when content is identical, defeating content-based caching.

2. **Last-Modified rotates with validity bucket** - Using `max(updated_at, bucket_start)` breaks If-Modified-Since semantics.

3. **Gzip compression is unconditional** - Always returns `Content-Encoding: gzip` without checking `Accept-Encoding`.

### Minor Issues (Should Fix)

4. **Cache-Control includes unnecessary `public` directive** - Ticket specifies `max-age=<ttl>, immutable` only.

5. **Cyclic dependency** - `error.rs` imports from `get_status_list.rs`.

### Suggestions

6. **Consider weak ETag comparison** per RFC 9110 §8.8.3.2.

7. **Verify updated_at updates** in `update_status.rs`.

See inline comments for details.

Hermann-Core (commented):
@ndefokou Comments will be posted for your attention.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -36,33 +38,99 @@ pub async fn get_status_list(
...
+
+    // The served representation is a re-signed token that changes every
+    // validity bucket even when the list content does not, so Last-Modified
+    // must rotate with the bucket. Using the raw `updated_at` (bumped only by
+    // publish/update_status and never by token re-issuance) would let a client
+    // or CDN revalidating with If-Modified-Since alone receive 304
+    // indefinitely and keep replaying an expired signed token, defeating the
+    // ttl/exp staleness bound. `max(updated_at, bucket_start)` keeps content
+    // changes visible on the IMS path while bounding staleness to the bucket.
+    let last_modified_ts = status_record.updated_at.max(bucket_start);
```
Hermann-Core: **Bug — HTTP caching can keep expired signed tokens alive indefinitely.**

The `Last-Modified` header is set to `status_record.updated_at` (line 79), which only changes when the bitstring content is modified via `update_status` or `publish_status`. The ETag is similarly content-based. This means:

1. Client fetches token at T=0, receives `exp=T+token_exp_secs`, `Cache-Control: max-age=token_ttl_secs`, `ETag: W/"hash"`, `Last-Modified: <updated_at>`
2. At T=token_ttl_secs (cache expires), client revalidates with `If-None-Match: W/"hash"` or `If-Modified-Since: <updated_at>`
3. Server returns `304 Not Modified` because content hasn't changed
4. Client's cached token (with `exp=T+token_exp_secs`) is reused
5. If `token_exp_secs > token_ttl_secs`, steps 2-4 repeat indefinitely — the client keeps serving a token whose `exp` claim is now in the past, violating RFC 9110 §4.2.1 and the Status List security model (which relies on `exp` to bound staleness)

The fix from review #2 was to include a `validity_bucket` in the ETag so it rotates every `token_exp_secs`, but that was removed (correctly, to avoid unnecessary cache churn). The correct solution is:

**Set `Last-Modified = max(updated_at, current_validity_bucket_start)` and use that same value for `If-Modified-Since` evaluation** so validators expire aligned with the signed token's `exp`. This bounds staleness to `token_exp_secs` while keeping the ETag content-based.

```rust
let now = OffsetDateTime::now_utc().unix_timestamp();
let bucket_start = (now / state.token_exp_secs as i64) * state.token_exp_secs as i64;
let last_modified_ts = status_record.updated_at.max(bucket_start);
```

Without this, the caching implementation is **unsafe for production** — relying parties can serve revoked credentials indefinitely via repeated 304 responses.

src/web/handlers/status_list/get_status_list.rs
```diff
@@ -36,33 +38,99 @@ pub async fn get_status_list(
...
+    ) {
+        ConditionalResponse::NotModified => {
+            // Return 304 with caching headers but no body
+            Ok((
+                StatusCode::NOT_MODIFIED,
+                [
+                    (header::ETAG, current_etag.as_str()),
+                    (header::LAST_MODIFIED, last_modified.as_str()),
+                    (header::CACHE_CONTROL, cache_control.as_str()),
+                    (header::VARY, "Accept"),
```
Hermann-Core: **Bug — Missing `Vary: Accept-Encoding` header.**

The response body varies based on the `Accept-Encoding` header (gzip vs uncompressed), but only `Vary: Accept` is emitted. Per RFC 9110 §12.5.5, the `Vary` header must list **all** request headers that influence response selection. Without `Vary: Accept-Encoding`, a CDN/proxy might cache a gzip-compressed response and serve it to a client that doesn't support gzip (or vice versa), causing decode failures.

Fix:
```rust
(header::VARY, "Accept, Accept-Encoding"),
```

This applies to both the 304 (line 99) and 200 (line 118) response paths.

Hermann-Core (changes requested):
The HTTP caching implementation has good structure and test coverage, but contains two blocking security issues that allow expired signed tokens to remain cacheable indefinitely: 1. **Missing Last-Modified rotation** — Last-Modified uses only updated_at (database field), not aligned with token validity periods. CDNs/clients can revalidate with If-Modified-Since and receive 304s indefinitely, keeping expired tokens alive. 2. **Missing Vary: Accept-Encoding** — Response body varies by Accept-Encoding (gzip vs raw) but Vary header only lists Accept. CDNs may serve mismatched encodings to clients. Issue #1 violates the Status List spec's revocation freshness model and must be fixed by setting Last-Modified = max(updated_at, validity_bucket_start) to bound staleness to token_exp_secs. Issue #2 causes HTTP protocol violations and decode failures. Both fixes are straightforward. See inline comments for details and proposed solutions.

Christiantyemele (approved):

Ngha-Boris (approved):

martcpp:
@Hermann-Core give a final look at this thanks since you already in the look

Hermann-Core (approved):

Issue #156, opened 2026-07-02
Improve API error handling


Hermann-Core:
## Objective

Improve the API error handling to match the structured, style used by the [cloud-identity-wallet](https://github.com/ADORSYS-GIS/cloud-identity-wallet) project. This ensures consistent, machine-readable error responses across all endpoints.

## Context

Currently, error responses vary in format across different handlers (auth errors, status list errors, database errors). The cloud-identity-wallet project uses a well-structured error response model, We should adopt the same pattern.

These error types are converted to HTTP responses inconsistently.

## Deliverables

- [x] Review cloud-identity-wallet's error handling model ([reference](https://github.com/ADORSYS-GIS/cloud-identity-wallet)).
- [x] Define a unified `ApiError` response struct.
- [x] Implement a central error→HTTP response mapping layer.
- [x] Replace all ad-hoc error responses with the unified model.
- [x] Add logging/tracing for all error responses (structured logs).
- [x] Add tests for error response serialization.
- [x] Ensure the OpenAPI spec (#147) documents the standard error response.

## Acceptance Criteria

- [x] Error responses are consistent across all endpoints (auth, status list, database).
- [x] Structured logging is attached to error responses.
- [x] Error response tests pass.
- [ ] OpenAPI spec includes the standard error schema.

## Status

Implementation complete in PR #193. All CI checks pass (build, clippy with `-D warnings`, fmt, tests). A follow-up merge from main required fixing build/test failures caused by the `add_updated_at` migration (PR #214) — resolved in commit `06bcee5` by adding the missing `updated_at` field to test fixtures, gating the `MigratorTrait` import behind the `sqlite` feature, using manual `MigrationName` impls for unique migration version strings, and replacing raw SQL with the SeaQuery builder for cross-backend compatibility.

## References

- Cloud Identity Wallet error handling: https://github.com/ADORSYS-GIS/cloud-identity-wallet
- Related: #147 (OpenAPI spec)
- Implemented in: PR #193

PR #216, opened 2026-07-15
Adding api rate limiting and enforcing requestproduct bounds

### LINKED ISSUE #171: Add API rate limiting and enforce request/product bounds
## Objective

Protect public retrieval and authenticated management endpoints with rate limiting, and define explicit limits on status-list size, index bounds, update count, and request body size.

## Context

The architecture review notes that rate limiting is documented but not implemented, and that the bit-packing code can grow arrays based on the highest index. Both expose the server to abuse and complicate horizontal scaling.

## Deliverables

- [ ] Add per-route rate limiting (e.g., via `tower-governor`):
  - stricter limits for `POST /api/v1/credentials`, `POST /api/v1/status-lists/...`, `PATCH /api/v1/status-lists/...`;
  - more permissive but bounded limits for `GET /statuslists/{list_id}`.
- [ ] Configure Axum body-size and JSON-size limits for all endpoints.
- [ ] Enforce maximum status-list index and maximum number of statuses per publish/update request.
- [ ] Enforce a maximum serialized status-list size.
- [ ] Add tests for abuse scenarios (oversized body, huge index, high request rate).
- [ ] Update Helm values and `.env.template` with the new tunables.

## Acceptance Criteria

- Public and management endpoints are rate-limited independently.
- Requests exceeding size, index, or update-count limits return a clear `400`/`413`/`429` error.
- Existing valid requests continue to pass.
- Abuse tests are green.

## References

- Related: #145 (REST redesign), #156 (error handling)

ndefokou:
Implements rate limiting, body-size limits, and index/size bounds to protect public retrieval and authenticated management endpoints from abuse, as requested in #171.

### Rate Limiting (`tower-governor`)
- Two independent per-IP governor tiers:
  - **Strict** (default: 10 req/60s) — `POST /credentials`, `PUT/PATCH /status-lists/{list_id}/statuses`
  - **Permissive** (default: 100 req/60s) — `GET /status-lists/{list_id}`, `GET /aggregation`
- Exceeding the burst quota returns `429 Too Many Requests` with `Retry-After` header.

### Body-Size Limits (`tower-http`)
- `RequestBodyLimitLayer` applied globally and per-route-group (default: 2 MiB).
- Oversized bodies return `413 Payload Too Large`.

### Index & Update-Count Bounds
- New `AbuseLimits` struct enforces `max_status_index` in `create_status_list` / `update_status_list`, preventing unbounded bit-packed array growth from high indices.
- Handlers reject requests exceeding `max_statuses_per_request` before any DB or bit-packing work.
- Returns `400 Bad Request` for `TooManyStatuses` or `IndexTooLarge`, `413` for oversized serialized lists.

### Config Tunables
| Tunable | Default | Error |
|---|---|---|
| `rate_limit.strict_burst_size` | 10 | 429 |
| `rate_limit.strict_period_secs` | 60 | 429 |
| `rate_limit.permissive_burst_size` | 100 | 429 |
| `rate_limit.permissive_period_secs` | 60 | 429 |
| `limits.max_body_size_bytes` | 2,097,152 (2 MiB) | 413 |
| `limits.max_status_index` | 100,000 | 400 |
| `limits.max_statuses_per_request` | 5,000 | 400 |
| `limits.max_serialized_list_size` | 1,048,576 (1 MiB) | 413 |

### Deployment
- `.env.template` and `helm/chart/values.yaml` updated with all new tunables.

## Tests
- 12 new tests added (94 total, all green):
  - Handler-level: rejects too-many-statuses (400), oversized index (400), oversized serialized list (413) for both publish and update.
  - Router-level: strict governor returns 429 after burst exhausted while permissive tier stays unaffected; body limit returns 413 for oversized bodies, allows normal ones.
  - Unit-level (`lst_gen.rs`): `IndexTooLarge` and `SerializedListTooLarge` enforced in isolation.
  - Config: defaults and env-override tests for all new tunables.

## Acceptance Criteria
- [x] Public and management endpoints are rate-limited independently.
- [x] Requests exceeding size, index, or update-count limits return clear 400/413/429 errors.
- [x] Existing valid requests continue to pass.
- [x] Abuse tests are green.

Closes #171


src/startup.rs
```diff
@@ -83,22 +106,65 @@ impl HttpServer {
...
+    Arc<GovernorConfig<PeerIpKeyExtractor, NoOpMiddleware>>,
+);
+
+fn build_governor_configs(config: &crate::config::RateLimitConfig) -> GovernorPolicies {
+    let strict = Arc::new(
+        GovernorConfigBuilder::default()
+            .burst_size(config.strict_burst_size)
+            .period(Duration::from_secs(config.strict_period_secs))
+            .finish()
+            .expect("strict governor config requires non-zero burst_size and period"),
```
Christiantyemele: **Important — startup panic on zero-valued rate-limit config**

Both `.expect("strict governor config requires non-zero burst_size and period")` (line 121) and the equivalent on line 128 will panic and crash the server at boot if an operator sets `APP_RATE_LIMIT__STRICT_BURST_SIZE=0` (or `PERIOD_SECS=0`). `GovernorConfigBuilder::finish()` returns `None` for non-zero requirements, and `.expect()` turns that into a process panic.

Since `burst_size` is `u32` and `period_secs` is `u64`, zero values are valid for the config deserializer — no validation guard exists.

**Fix:** return a proper config error instead of panicking:
```rust
fn build_governor_configs(config: &RateLimitConfig) -> color_eyre::Result<GovernorPolicies> {
    let strict = Arc::new(
        GovernorConfigBuilder::default()
            .burst_size(config.strict_burst_size)
            .period(Duration::from_secs(config.strict_period_secs))
            .finish()
            .ok_or_else(|| eyre!("strict governor requires non-zero burst_size and period"))?,
    );
    // ... same for permissive
    Ok((strict, permissive))
}
```

src/startup.rs
```diff
@@ -54,13 +64,26 @@ impl HttpServer {
...
+                    state.clone(),
+                    strict_governor.clone(),
+                    permissive_governor.clone(),
+                    max_body_size,
+                ),
+            )
             .layer(TraceLayer::new_for_http())
             .layer(CatchPanicLayer::new())
             .layer(cors)
+            .layer(RequestBodyLimitLayer::new(max_body_size))
```
Christiantyemele, line 89: **Suggestion — redundant RequestBodyLimitLayer**

A global `RequestBodyLimitLayer::new(max_body_size)` is applied at line 86 (covering all routes including `/`, `/health`, `/metrics`), and the same limit is also applied per-route-group at lines 151, 156, 162. The per-route layer fires first for `/api/v1` routes, making the global layer a no-op for those routes. Pick one strategy to simplify.

src/startup.rs
```diff
@@ -54,13 +64,26 @@ impl HttpServer {
...
+        let max_body_size = config.limits.max_body_size_bytes;
+
+        let (strict_governor, permissive_governor) = build_governor_configs(&config.rate_limit);
+
         let mut router = Router::new()
             .route("/", get(welcome))
             .route("/health", get(health_check))
-            .nest("/api/v1", api_v1_routes(state.clone()))
+            .nest(
+                "/api/v1",
```
Christiantyemele, line 78: **Question — undocumented breaking route restructuring**

This PR restructures route paths and HTTP methods vs `main`:

| Old (main) | New |
|---|---|_logspot|_logspot
| `POST /credentials` | `POST /api/v1/credentials` |tch /statuslists/publish` | `PUT /api/v1/status-lists/{list_id}/statuses` |
| `PATCH /statuslists/update` | `PATCH /api/v1/status-lists/{list_id}/statuses` |
| — | `GET /api/v1/aggregation` (new) |
| `GET /statuslists/{list_id}` | `GET /api/v1/status-lists/{list_id}` |

These are breaking changes for existing API consumers. The PR description only mentions rate limiting and bounds. Are these route changes intentional and in scope for #171?

src/web/handlers/status_list/error.rs
```diff
@@ -60,6 +66,9 @@ impl IntoResponse for StatusListError {
             TokenAlreadyExists => StatusCode::CONFLICT,
             IssuerMismatch => StatusCode::FORBIDDEN,
             ServiceUnavailable => StatusCode::SERVICE_UNAVAILABLE,
+            TooManyStatuses { .. } => StatusCode::BAD_REQUEST,
+            IndexTooLarge(_) => StatusCode::BAD_REQUEST,
+            StatusTooLarge => StatusCode::PAYLOAD_TOO_LARGE,
```
Christiantyemele: **Nit — 413 for oversized generated output is semantically unusual**

`StatusTooLarge => StatusCode::PAYLOAD_TOO_LARGE` (413) implies the *incoming request body* is too large, but `SerializedListTooLarge` is triggered when the *generated/compressed status list* exceeds the limit. This creates ambiguity for clients who receive 413 and cannot tell whether their request body was too big or the resulting list grew too big. Consider 422 Unprocessable Entity.

Christiantyemele (dismissed):
## Automated Code Review — REQUEST_CHANGES

Results: 1 Bug, 1 Important, 1 Suggestion, 1 Question, 1 Nit. Inline comments failed due to line resolution on a non-diff line; all findings posted below.

Verification: `cargo check` ✅ | `cargo clippy --all-targets -- -D warnings` ✅ | `cargo test` — all 94 tests pass (startup 7, lst_gen+config 32, handlers 18, rest filtered).

⚠️ **Note:** Inline review comments could not be posted (GitHub line resolution failed for line 102 which is not in the diff). All findings are in the issue comment below.

---

More detail in the issue comment ⬇️

martcpp:
@ndefokou are you limiting per ip or per user what design method did you chose and why

Hermann-Core (commented):
@ndefokou Comments will be posted for your attention.

src/startup.rs
Hermann-Core: Bug: expect() will panic on zero config values. Use Result return type instead.

src/startup.rs
```diff
@@ -83,22 +106,65 @@ impl HttpServer {
...
     let protected = Router::new()
         .nest(
             "/status-lists/{list_id}/statuses",
             Router::new()
                 .route("/", put(publish_status))
                 .route("/", patch(update_status)),
         )
-        .route_layer(from_fn_with_state(state.clone(), auth));
+        .route_layer(from_fn_with_state(state.clone(), auth))
+        .layer(GovernorLayer::new(strict_governor.clone()))
```
Hermann-Core: Design concern: Per-IP rate limiting on authenticated endpoints can be bypassed via distributed attacks and penalizes users behind NAT. For authenticated write APIs, per-user rate limiting (keying on JWT issuer claim) is more robust and accurate. Consider implementing a custom KeyExtractor that uses issuer identity from the auth middleware.

src/startup.rs
```diff
@@ -83,22 +106,65 @@ impl HttpServer {
...
+        .layer(body_limit);
```
Hermann-Core: Redundant RequestBodyLimitLayer: Applied globally at line 86 covering all routes, then applied again per-route at lines 151,159,165. The per-route layers serve no purpose since the global layer already enforces the same limit. Remove the per-route layers to avoid double-processing overhead.

Hermann-Core (changes requested):
The rate limiting and abuse protection implementation provides good coverage with reasonable defaults, but contains one blocking bug and two design concerns: 1. **Bug (blocking)** — Startup panic on zero-valued rate-limit config. The expect() calls will crash the server instead of returning a clear error. Must be fixed to use Result return type and propagate errors gracefully. 2. **Design concern** — Per-IP rate limiting on authenticated write endpoints can be bypassed and penalizes NAT users. Per-user rate limiting (JWT issuer-based) would be more robust for authenticated APIs. 3. **Performance** — Redundant RequestBodyLimitLayer applied both globally and per-route causes unnecessary double-processing. Issue #1 must be fixed before merge. Issues #2 and #3 are recommendations for production hardening. Tests are comprehensive (94 passing) and defaults are reasonable. See inline comments for details.

ndefokou:
## Review fixes pushed (8df2427)

Thanks for the thorough review @Christiantyemele. Here's what was addressed:

### 1. Important — Governor panic → config error ✅
`build_governor_configs` now returns `color_eyre::Result<GovernorPolicies>` instead of calling `.expect()`. If an operator sets `APP_RATE_LIMIT__STRICT_BURST_SIZE=0` (or `PERIOD_SECS=0`), the server returns a clear config error at boot rather than crashing via panic.

### 2. Suggestion — Redundant RequestBodyLimitLayer ✅
Removed the per-route `RequestBodyLimitLayer` from `api_v1_routes`. The global layer applied at the router level (line 90) already covers all routes. The per-route layers were a no-op for `/api/v1/*` routes since they fired first. The `max_body_size` parameter has been removed from `api_v1_routes`.

### 3. Nit — 413 → 422 for oversized list ✅
Changed `StatusTooLarge` from `413 Payload Too Large` to `422 Unprocessable Entity` in `error.rs`. This avoids ambiguity: 413 implies the incoming request body was too large, but `StatusTooLarge` triggers when the generated/compressed output exceeds the configured maximum. Tests in `publish_status.rs` and `update_status.rs` updated accordingly.

### 4. Question — Route restructuring
Yes, the route restructuring is intentional and in scope for #171. The rate-limiting work required tiering routes by sensitivity (auth-protected writes vs public reads), which surfaced the need for a consistent `/api/v1` namespace and RESTful path conventions. The `/api/v1` prefix and resource-style paths (`/status-lists/{list_id}/statuses`) align the API with standard conventions. These are breaking changes for existing consumers and will be documented in the release notes. If the team prefers route changes to be handled in a separate PR, we can split them out — let us know.

ndefokou:
@Hermann-Core Thank you for the review. Both issues you raised have been addressed in commit `8df2427` (the comments above are on the outdated code). Summary below.

---

### 1. Bug — `expect()` panic on zero-valued rate-limit config ✅ Fixed

`build_governor_configs` now returns `color_eyre::Result<GovernorPolicies>` instead of panicking. Each `GovernorConfigBuilder::finish()` call uses `.ok_or_else(|| eyre!(...))?` so a zero `burst_size` or `period_secs` produces a clear boot-time error rather than crashing the process:

```rust
fn build_governor_configs(
    config: &crate::config::RateLimitConfig,
) -> color_eyre::Result<GovernorPolicies> {
    let strict = Arc::new(
        GovernorConfigBuilder::default()
            .burst_size(config.strict_burst_size)
            .period(Duration::from_secs(config.strict_period_secs))
            .finish()
            .ok_or_else(|| eyre!("strict governor requires non-zero burst_size and period"))?,
    );
    // … same pattern for issuer and permissive tiers
    Ok((strict, issuer, permissive))
}
```

The `?` propagates back through `HttpServer::new` (`src/startup.rs:72`), so an invalid config surfaces as a descriptive `color_eyre` error at startup.

---

### 2. Design concern — Per-IP rate limiting on authenticated write endpoints ✅ Fixed

A new `IssuerKeyExtractor` (`src/web/rate_limit.rs`) implements the `tower_governor::key_extractor::KeyExtractor` trait. It decodes the `iss` claim from the Bearer JWT and uses that as the rate-limit key, falling back to the peer IP when the token is absent or malformed (so unauthenticated requests are still throttled before `auth` rejects them with `401`).

The authenticated `protected` group now uses a dedicated **per-issuer governor** instead of the per-IP strict tier:

```rust
let protected = Router::new()
    .nest("/status-lists/{list_id}/statuses", …)
    .route_layer(from_fn_with_state(state.clone(), auth))
    .layer(GovernorLayer::new(issuer_governor));
```

This gives every verified issuer its own independent quota, so attacks from a single compromised issuer are contained without penalising users behind a shared NAT address.

---

### Bonus — Redundant `RequestBodyLimitLayer` ✅ Fixed

The per-route body-limit layers have been removed; a single global `RequestBodyLimitLayer::new(max_body_size)` at `src/startup.rs:89` now covers all routes.

---

**Verification:** `cargo check` ✅ · `cargo clippy --all-targets -- -D warnings` ✅ · `cargo test` — 100 unit tests pass (including `test_per_issuer_governor_independent_quotas` which exercises the `IssuerKeyExtractor`, and all five `rate_limit` unit tests). The two `cert_provisioning` integration tests fail only because the `ghcr.io/letsencrypt/pebble-challtestsrv` Docker image cannot be pulled in this sandbox — they are unrelated to the rate-limiting changes.

Please re-review when you have a moment.

Hermann-Core (approved):
All three issues from the previous review have been properly fixed.

martcpp:
will love to take a look at this today thanks

src/startup.rs
martcpp, line 1: into_make_service_with_connect_info missing; per-IP-limited routes (all public reads + /credentials) 500 on every request in production.

src/startup.rs
```diff
@@ -83,22 +109,71 @@ impl HttpServer {
...
     let protected = Router::new()
         .nest(
             "/status-lists/{list_id}/statuses",
             Router::new()
                 .route("/", put(publish_status))
                 .route("/", patch(update_status)),
         )
-        .route_layer(from_fn_with_state(state.clone(), auth));
+        .route_layer(from_fn_with_state(state.clone(), auth))
+        .layer(GovernorLayer::new(issuer_governor));
```
martcpp, line 167: per-issuer limiter keys on unverified iss before auth: rate-limit bypass with unthrottled per-request DB lookups, unauthenticated targeted tenant lockout, attacker-controlled key growth. Must move inside the auth boundary + outer per-IP layer.

src/startup.rs
```diff
@@ -83,22 +109,71 @@ impl HttpServer {
...
+
+fn build_governor_configs(
+    config: &crate::config::RateLimitConfig,
+) -> color_eyre::Result<GovernorPolicies> {
+    let strict = Arc::new(
+        GovernorConfigBuilder::default()
+            .burst_size(config.strict_burst_size)
+            .period(Duration::from_secs(config.strict_period_secs))
+            .finish()
+            .ok_or_else(|| eyre!("strict governor requires non-zero burst_size and period"))?,
```
martcpp, line 132: no retain_recent() eviction task; unbounded memory growth, attacker-amplified

src/utils/lst_gen.rs
```diff
@@ -245,20 +287,27 @@ fn reencode_status_array(
...
+        }
         if let Some(entry) = full_statuses.iter_mut().find(|e| e.index == update.index) {
             entry.status = update.status.clone();
         } else {
             full_statuses.push(update.clone());
         }
     }
 
-    create_status_list(full_statuses)
+    create_status_list(full_statuses, limits)
```
martcpp, line 303: legacy lists with indexes above the new bound reject valid bit-growth updates (violates "existing valid requests continue to pass").

src/startup.rs
```diff
@@ -54,13 +66,27 @@ impl HttpServer {
...
+                    state.clone(),
+                    strict_governor.clone(),
+                    issuer_governor.clone(),
+                    permissive_governor.clone(),
+                ),
+            )
             .layer(TraceLayer::new_for_http())
             .layer(CatchPanicLayer::new())
             .layer(cors)
+            .layer(RequestBodyLimitLayer::new(max_body_size))
```
martcpp, line 89: Add DefaultBodyLimit::max alignment

src/startup.rs
```diff
@@ -83,22 +109,71 @@ impl HttpServer {
...
+    );
+    let issuer = Arc::new(
+        GovernorConfigBuilder::default()
+            .burst_size(config.strict_burst_size)
+            .period(Duration::from_secs(config.strict_period_secs))
+            .key_extractor(IssuerKeyExtractor)
+            .finish()
+            .ok_or_else(|| eyre!("issuer governor requires non-zero burst_size and period"))?,
+    );
+    let permissive = Arc::new(
```
martcpp, line 142: Issuer tier silently reuses strict_* tunables

helm/chart/values.yaml
martcpp, line 1: Per-replica limit semantics + ingress client-IP caveat

martcpp (changes requested):
Update openapi.yaml (429/413/422/400 + bounds) and architecture.md
thanks

martcpp:
@ndefokou build  failing

ndefokou:
> @ndefokou build failing

I am on it i saw it

ndefokou:
@martcpp Please review this again

src/startup.rs
martcpp, line 1: ConnectInfo<SocketAddr> is never populated in production, so every PeerIpKeyExtractor-gated route returns HTTP 500 for every request. HttpServer::run passes the bare Router to axum::serve Axum only inserts ConnectInfo when you call .into_make_service_with_connect_info::<SocketAddr>() first (confirmed against the axum 0.8.9 source and its own doc comment: "See also Router::into_make_service_with_connect_info"). PeerIpKeyExtractor::extract returns Err(GovernorError::UnableToExtractKey) when the extension is absent (confirmed in tower_governor 0.8.0 source, key_extractor.rs:76), and that error is converted to 500 Internal Server Error (errors.rs:58-64). This means, as merged, POST /api/v1/credentials, GET /api/v1/aggregation, and GET /api/v1/status-lists/{list_id} would 500 on every real request, and unauthenticated hits on the write endpoints would too (the IssuerKeyExtractor fallback hits the same missing extension). This is not caught by any existing test because every test manually injects the extension (§6). This has to be fixed before merge — change run() to serve self.router.into_make_service_with_connect_info::<SocketAddr>(), and re-verify against a real request (not a hand-built one) that ConnectInfo actually flows through the TraceLayer/CatchPanicLayer/CORS/body-limit stack into the nested router.

martcpp (changes requested):
The per-issuer rate limit on PUT/PATCH /status-lists/{id}/statuses is bypassable with zero cost by sending an unsigned JWT with a fresh, attacker-chosen iss per request —see §5a. As written, the "stricter limits" the issue asks for on the write path don't hold against an actual attacker, only against a legitimate, already-authenticated issuer hammering their own endpoint. Needs to be re-keyed off IP (or a verified issuer) rather than an unverified claim.
  re-evaluate whether PeerIpKeyExtractor is the right choice at all given the Helm/ingress deployment target  SmartIpKeyExtractor is very likely what's actually wanted, or this needs to be paired with trusting only a specific proxy's X-Forwarded-For.

src/web/rate_limit.rs
```diff
@@ -0,0 +1,140 @@
...
+    value.get("iss")?.as_str().map(|s| s.to_string())
+}
+
+fn peer_ip<T>(req: &Request<T>) -> Option<IpAddr> {
+    req.extensions()
+        .get::<axum::extract::ConnectInfo<SocketAddr>>()
+        .map(|addr| addr.ip())
+}
+
+#[cfg(test)]
```
martcpp, line 51: Add one true integration test that boots HttpServer (or at least into_make_service_with_connect_info + oneshot) end-to-end through the real router construction, to catch class-of-bug #1 in CI going forward.

src/startup.rs
martcpp, line 1: Disable or align Axum's built-in DefaultBodyLimit with RequestBodyLimitLayer so max_body_size_bytes works above 2 MiB (§3).

docs/architecture.md
martcpp, line 1: Note in the architecture doc that in-memory governor state is per-replica and does not provide a global rate limit across a horizontally scaled deployment (§1), so operators don't assume otherwise

src/utils/errors.rs
```diff
@@ -8,4 +8,8 @@ pub(crate) enum Error {
     InvalidIndex,
     #[error("Failed to decode lst")]
     DecodeFailed,
+    #[error("status index {0} exceeds the configured maximum")]
+    IndexTooLarge(i32),
+    #[error("serialized status list size {actual} bytes exceeds maximum {max} bytes")]
+    SerializedListTooLarge { actual: usize, max: usize },
 }
```
martcpp, line 15: docs/architecture.md's new line "Handler-level errors use RFC 7807 application/problem+json" is not true for StatusListError its IntoResponse returns a raw Display string body, not JSON . This mismatch predates this PR for the existing error variants, but this PR both restates the claim in prose and adds a new docs/openapi.yaml UnprocessableEntity block claiming application/problem+json content for the new 422 extending an existing inaccuracy into new documentation rather than noting the gap (tracked separately as #156). Either implement the RFC7807 body or soften the doc claim to match current behavior.

martcpp (changes requested):

Hermann-Core (changes requested):
**Critical security vulnerability and production-breaking bug found since last review.**

@martcpp identified two showstopper issues that must be fixed before merge:

## 1. Security: Per-issuer rate limiting is bypassable (CRITICAL)

The `IssuerKeyExtractor` reads the `iss` claim from **unverified JWT tokens** before signature verification happens. An attacker can send unsigned JWTs with arbitrary `iss` values to bypass rate limiting entirely.

**Attack scenario:**
```bash
# Attacker bypasses rate limit by rotating fake issuers
for i in {1..1000}; do
  curl -H "Authorization: Bearer $(fake_jwt_with_iss "attacker-$i")" \
    -X PUT https://example.com/api/v1/status-lists/victim/statuses
done
```

Each request uses a different `iss` claim, so each gets a fresh rate limit bucket. The auth middleware (which verifies signatures) runs **after** rate limiting, so these requests are rate-limited per fake issuer before being rejected for invalid signatures.

**Root cause:** Architecture in `src/startup.rs` line 161-162 applies Governor before auth:
```rust
.route_layer(from_fn_with_state(state.clone(), auth))  // Signature verification
.layer(GovernorLayer::new(issuer_governor));           // Rate limit by unverified iss
```

Layers execute bottom-to-top, so rate limiting sees the unverified JWT.

**Fix required:**
- Switch write endpoints to `PeerIpKeyExtractor` (or `SmartIpKeyExtractor` for proxy deployments)
- OR move rate limiting **after** auth verification (architectural change)
- OR extract issuer from verified claims in auth middleware and pass via extension

The current "per-issuer" rate limiting provides zero protection against actual attackers.

## 2. Production: All PeerIpKeyExtractor routes return 500 (SHOWSTOPPER)

`HttpServer::run()` calls `axum::serve(listener, self.router)` without `.into_make_service_with_connect_info::<SocketAddr>()`. This means `ConnectInfo<SocketAddr>` is never populated, causing `PeerIpKeyExtractor::extract` to return `Err(GovernorError::UnableToExtractKey)`, which tower-governor converts to **500 Internal Server Error**.

**Affected routes** (return 500 on every request):
- POST /api/v1/credentials
- GET /api/v1/aggregation  
- GET /api/v1/status-lists/{id}
- Unauthenticated requests to write endpoints (IssuerKeyExtractor fallback)

**Why tests don't catch this:** All tests manually inject `ConnectInfo` via `builder.extension(ci)` in test helpers.

**Fix:** Change `src/startup.rs` HttpServer::run to:
```rust
axum::serve(
    listener,
    self.router.into_make_service_with_connect_info::<SocketAddr>()
)
```

## Other issues from @martcpp's review:

3. **Body limit conflict:** Axum's default 2 MiB limit overrides configured `max_body_size_bytes` - need to disable `DefaultBodyLimit`

4. **Documentation gap:** Architecture doc should note that in-memory governor state is per-replica (no global rate limit in horizontal scaling)

5. **RFC 7807 claim:** Docs claim `application/problem+json` responses but `StatusListError::IntoResponse` returns plain text (pre-existing but extended by this PR's openapi.yaml additions)

**Verdict:** **Request changes** for critical security vulnerability (#1) and production-breaking bug (#2). Issues #3-5 should also be addressed but are lower priority.

**Tests:** Existing tests pass but miss both critical issues. Need integration test that boots HttpServer end-to-end.

---

@ndefokou The original approval was based on the functional correctness at commit 8df2427. These new issues were introduced or discovered in subsequent merges and martcpp's deeper security review.

ndefokou:
@Hermann-Core @martcpp You can review the ticket all changes have been address.

Hermann-Core (approved):

martcpp (approved):

PR #217, opened 2026-07-15
Refactor/make certificate provisioning configurable

### LINKED ISSUE #159: Make certificate provisioning configurable (ACME or direct store)
## Objective

Make certificate provisioning configurable to support multiple provisioning strategies and refactor the certificate manager to use the builder pattern for construction, removing the inline startup construction.

## Context

The server's certificate manager currently defaults to ACME with DNS-01. In environments where ACME is unavailable or certificates are managed externally (e.g., cert-manager, Kubernetes secrets, mounted files), operators must be able to load pre-provisioned certificates. At the same time, the certificate manager is constructed directly in `src/startup.rs` with many inline parameters, which becomes unwieldy as more strategies, storage backends, DNS providers and caching are added.

## Deliverables

- [ ] Define a `CertProvisioningStrategy` trait with impls:
  - `Acme` — existing ACME-based provisioning (DNS-01 / Pebble / Let's Encrypt).
  - `Store` — provision from a certificate store (mounted files, secrets manager, object store).
- [ ] Implement `Store` backend loading certificate chain and signing material from filesystem, AWS Secrets Manager, or HashiCorp Vault.
- [ ] Add configuration for selecting the provisioning strategy and store parameters.
- [ ] Add hot-reload support for store-based certificates (watch file changes / periodic refresh).
- [ ] Integrate with the certificate chain cache (#158) — invalidate the cache on reload.
- [ ] Create a `CertificateManagerBuilder` with fluent setters for provisioning strategy, cache, DNS provider, storage, challenge type, etc.
- [ ] Move construction logic from `src/startup.rs` into the builder.
- [ ] Validate required fields in `build()` and provide clear errors for missing components.
- [ ] Ensure the builder supports all current and planned features (multiple provisioning strategies, storage, DNS providers, caching).
- [ ] Update unit tests, `.env.template`, Helm values, and documentation.

## Acceptance Criteria

- Operators can choose between ACME and store-based provisioning.
- Store-based provisioning loads certs from filesystem or supported secrets store.
- Hot-reload works for store strategy and invalidates the certificate chain cache.
- Certificate manager is constructed via a builder pattern.
- Missing required components produce clear validation errors at build time.
- All existing tests pass and new tests cover builder and store provisioning paths.
- Code is documented with rustdoc.

## References

- Related: #158 (cert chain caching), #150 (Docker image), #140 (external certificate management)

Estimated time: 2 days

Hermann-Core:
Make certificate provisioning configurable to support multiple provisioning strategies and refactor the certificate manager to use the builder pattern for construction, removing the inline startup construction.

## Acceptance Criteria

- Operators can choose between ACME and store-based provisioning.
- Store-based provisioning loads certs from filesystem or supported secrets store.
- Hot-reload works for store strategy and invalidates the certificate chain cache.
- Certificate manager is constructed via a builder pattern.
- Missing required components produce clear validation errors at build time.
- All existing tests pass and new tests cover builder and store provisioning paths.


martcpp:
@Hermann-Core pls resolve this conflict

src/utils/cert_manager/strategy.rs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            info!("Store certificate material is unchanged");
+            return Ok(current_certificate.unwrap_or(certificate_data));
+        }
+
+        manager.persist_certificate_data(&certificate_data).await?;
+        manager.persist_signing_key(&signing_key_pem).await?;
+        info!("Store certificate material refreshed");
+        Ok(certificate_data)
+    }
```
Ngha-Boris, line 224: > **Issue:** Cache re-population missing after store provisioning
>
> After persisting the certificate data and signing key (lines 217-218), the certificate chain cache is not being updated. The code only persists to storage but doesn't update the in-memory cache that other parts of the system rely on.
>
> **Suggestion:** Add cache update call after line 218:
> ```rust
> manager.cache_provisioned_chain(&certificate_data.certificate).await?;
> ```
> This ensures the cache is invalidated and re-populated with the new certificate data.

src/utils/cert_manager/strategy.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    async fn provision(&self, manager: &CertManager) -> Result<CertificateData, CertError> {
+        manager.request_acme_certificate().await
+    }
+}
+
+/// Source for directly provisioned certificate material.
+#[derive(Debug, Clone)]
+pub enum StoreProvisioningSource {
```
Ngha-Boris, line 52: > **Issue:** Missing HashiCorp Vault Support
>
> The ticket deliverables mention implementing `Store` backend loading from "HashiCorp Vault", but this variant is missing from the `StoreProvisioningSource` enum. Only `Filesystem`, `Storage`, and `SecretsStorage` are implemented.
>
> **Suggestion:** Add a `HashiCorpVault` variant or create a follow-up issue to track this missing deliverable from the ticket requirements.

src/utils/cert_manager/builder.rs
```diff
@@ -0,0 +1,256 @@
+use std::{sync::Arc, time::Duration};
+
+use instant_acme::{Account, HttpClient};
```
Ngha-Boris, line 3: > **Issue:** Unused Import
>
> The `Account` import from `instant_acme` is not used in this file.
>
> **Suggestion:** Remove the unused import:
> ```rust
> use instant_acme::HttpClient;
> ```

src/utils/cert_manager/builder.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub fn build(self) -> Result<CertManager, CertError> {
+        if self.domains.is_empty() {
+            return Err(CertError::Validation(
+                "at least one certificate domain must be configured".to_string(),
+            ));
+        }
+
+        let cert_storage = self.cert_storage.ok_or_else(|| {
+            CertError::Validation("certificate storage backend must be configured".to_string())
+        })?;
```
Ngha-Boris, line 196: > **Issue:** String-based Strategy Name Comparison
>
> Using string comparison (`strategy.name() == "acme"`) for determining strategy type is fragile. If someone changes the name returned by `AcmeProvisioningStrategy`, this validation will break.
>
> **Suggestion:** Consider using a type-based approach or const for the strategy name:
> ```rust
> const ACME_STRATEGY_NAME: &str = "acme";
> let strategy_uses_acme = strategy.name() == ACME_STRATEGY_NAME;
> ```
> Or better yet, add a method to the trait:
> ```rust
> fn is_acme(&self) -> bool { false }  // default impl
> ```

src/config.rs
```diff
@@ -92,6 +92,7 @@ pub struct ServerConfig {
 
 #[derive(Debug, Clone, Deserialize)]
 pub struct CertConfig {
+    pub provisioning_strategy: String,
     pub email: String,
```
Ngha-Boris, line 96: > **Issue:** No Validation for provisioning_strategy
>
> The `provisioning_strategy` field uses `String` type without validation. This allows invalid values to be passed in.
>
> **Suggestion:** Consider using an enum for compile-time safety:
> ```rust
> pub enum ProvisioningStrategy {
>     #[serde(rename = "acme")]
>     Acme,
>     #[serde(rename = "store")]
>     Store,
> }
> ```
> Or add runtime validation in `Config::load()` to ensure only valid values are accepted.

src/utils/cert_manager/strategy.rs
```diff
@@ -0,0 +1,272 @@
...
+impl CertProvisioningStrategy for StoreProvisioningStrategy {
+    fn name(&self) -> &'static str {
+        "store"
+    }
+
+    fn should_provision_existing(
+        &self,
+        _manager: &CertManager,
+        _certificate: &CertificateData,
+    ) -> bool {
```
Ngha-Boris, line 197: > **Issue:** Store Strategy Always Returns `true` for `should_provision_existing`
>
> The `StoreProvisioningStrategy::should_provision_existing` always returns `true`, which means the store will be checked on every renewal cycle regardless of whether the certificate is still valid.
>
> **Suggestion:** Consider if this is the intended behavior. For store provisioning, you might want to:
> 1. Check if the file/storage content has changed (hash comparison)
> 2. Or add a TTL-based check to avoid unnecessary reloads
> 3. Or document why unconditional reloading is the desired behavior
>
> Currently, this could cause unnecessary storage reads.

Ngha-Boris (commented):
Hello @Hermann-Core Please take a look at these comments

src/utils/cert_manager/strategy.rs
Hermann-Core: Fixed — added `manager.cache_provisioned_chain(&certificate_data.certificate).await?;` after the persist calls. Good catch, this was a real bug where stale TLS chains would be served until the cache TTL expired naturally. The ACME path already had this call at `cert_manager.rs:312` but the store path was missing it.

src/utils/cert_manager/strategy.rs
Hermann-Core: HashiCorp Vault support is planned as a follow-up. The current `SecretsStorage` variant is backend-agnostic — a Vault implementation of the `Storage` trait can be plugged in without needing a dedicated enum variant. I'll create a tracking issue for explicit Vault integration.

src/utils/cert_manager/builder.rs
Hermann-Core: `Account` is used on line 244 as `None::<Account>` to type the ACME client mutex. The import is required.

src/utils/cert_manager/builder.rs
Hermann-Core: Fair point about fragility. However, the `name()` method serves double duty for tracing (line 236) and this single comparison. The builder's typed methods (`acme_strategy()` / `store_strategy()`) already prevent misconfiguration — this check just validates ACME-specific requirements post-construction. Adding `is_acme()` to the generic trait would leak strategy-specific concerns into the abstraction. The current approach is contained to the builder's `build()` validation.

src/config.rs
Hermann-Core: Runtime validation already exists in `state.rs` (lines 176-183) and rejects anything other than `"acme"` or `"store"` with a clear error message. Using a `String` allows case-insensitive matching (`eq_ignore_ascii_case`), which is more operator-friendly for config files. The error message explicitly lists the valid options.

src/utils/cert_manager/strategy.rs
Hermann-Core: This is intentional for hot-reload support. The store strategy always returns `true` so the renewal cron checks for new material on every cycle. However, the `provision()` method itself (lines 209-215) compares the loaded material against what's currently stored and short-circuits with a log message when unchanged — no persistence or cache work happens. The cost is one read per cycle, which is necessary to detect external certificate rotations (e.g., an operator replacing a file on disk or updating a secret).

Hermann-Core:
Hello @Ngha-Boris thanks for the review. Comments have been addressed, could you check again?

Ngha-Boris (approved):

PR #218, opened 2026-07-16
Acme dns provider support

### LINKED ISSUE #207: ACME-DNS provider: support >2 identifiers per certificate via per-domain credentials
## Background

The ACME-DNS provider holds a single account (one server URL / username /password / subdomain). Because acme-dns keeps only the **two most recent TXT values** per record by design, every domain's `_acme challenge` CNAME points at the same acme-dns record, and only the last two written digests survive.

## Problem

All authorizations for an order are created *before* the order is polled. For a SAN certificate with **3+ identifiers**, the earliest digests are rotated out of acme-dns before the CA validates them, so the order fails. This gives ACME-DNS an effective hard limit of **≤2 domains per certificate**.

Note: apex + wildcard on one cert (two TXT values at the same name) fits within the 2-value window, so that common case works. The limit only bites at 3+ distinct identifiers.

## Proposed enhancement

Support a per-domain credential map so each identifier can target its own acme-dns subdomain, matching how lego and acme.sh handle multi-domain acme-dns setups. Rough shape:

- Config accepts a map of `domain -> { subdomain, username, password }`  (or a default account plus per-domain overrides).
- Provider selects credentials by the authorization's identifier.
- Falls back to single-account behavior when no map is given.

## Acceptance criteria

- [ ] Per-domain credential selection implemented and config-driven.
- [ ] Backward compatible: existing single-account config works unchanged.
- [ ] wiremock tests cover per-domain selection and the 3-identifier case that currently fails.
- [ ] Docs updated to describe the map and remove the ≤2-identifier caveat once lifted.

blocked by #162

martcpp:
The ACME-DNS provider holds a single account (one server URL / username /password / subdomain). Because acme-dns keeps only the **two most recent TXT values** per record by design, every domain's `_acme challenge` CNAME points at the same acme-dns record, and only the last two written digests survive.

## Acceptance criteria

- [ ] Per-domain credential selection implemented and config-driven.
- [ ] Backward compatible: existing single-account config works unchanged.
- [ ] wiremock tests cover per-domain selection and the 3-identifier case that currently fails.
- [ ] Docs updated to describe the map and remove the ≤2-identifier caveat once lifted.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, time::Duration};
+use std::{collections::HashMap, fmt, marker::PhantomData, time::Duration};
 
 use config::{Config as ConfigLib, ConfigError, Environment};
 use redis::{
@@ -243,6 +243,34 @@ impl AcmeDnsConfig {
                     .to_string(),
             ));
         }
+        // Reject unusable per-domain entries here instead of as an opaque
+        // HTTP 401 at the first renewal
+        for (domain, account) in &self.accounts {
+            let name = domain.trim();
+            let name = name.strip_prefix("*.").unwrap_or(name);
+            if name.trim_end_matches('.').is_empty() {
+                return Err(ConfigError::Message(format!(
+                    "ACME-DNS accounts entry {domain:?} does not name a domain"
+                )));
+            }
+            let empty: Vec<&str> = [
+                ("username", account.username.trim().is_empty()),
+                (
+                    "password",
+                    account.password.expose_secret().trim().is_empty(),
+                ),
+                ("subdomain", account.subdomain.trim().is_empty()),
+            ]
+            .into_iter()
+            .filter_map(|(field, is_empty)| is_empty.then_some(field))
+            .collect();
+            if !empty.is_empty() {
+                return Err(ConfigError::Message(format!(
+                    "ACME-DNS account for {domain} has empty required fields: {}",
+                    empty.join(", ")
+                )));
+            }
+        }
         Ok(())
     }
 }
@@ -257,18 +285,37 @@ where
     D: serde::Deserializer<'de>,
     V: serde::de::DeserializeOwned,
 {
-    #[derive(Deserialize)]
-    #[serde(untagged)]
-    enum MapOrString<V> {
-        Map(HashMap<String, V>),
-        String(String),
+    // A visitor (rather than an untagged enum) so errors inside a map value
+    // surface as-is instead of as "did not match any variant"
+    struct MapOrString<V>(PhantomData<V>);
+
+    impl<'de, V: serde::de::DeserializeOwned> serde::de::Visitor<'de> for MapOrString<V> {
+        type Value = HashMap<String, V>;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("a map or a JSON object string")
+        }
+
+        fn visit_str<E: serde::de::Error>(self, raw: &str) -> Result<Self::Value, E> {
+            if raw.trim().is_empty() {
+                return Ok(HashMap::new());
+            }
+            serde_json::from_str(raw).map_err(E::custom)
+        }
+
+        fn visit_map<A: serde::de::MapAccess<'de>>(
+            self,
+            mut access: A,
+        ) -> Result<Self::Value, A::Error> {
+            let mut map = HashMap::with_capacity(access.size_hint().unwrap_or(0));
+            while let Some((key, value)) = access.next_entry()? {
+                map.insert(key, value);
+            }
+            Ok(map)
+        }
     }
 
-    match MapOrString::deserialize(deserializer)? {
-        MapOrString::Map(map) => Ok(map),
-        MapOrString::String(raw) if raw.trim().is_empty() => Ok(HashMap::new()),
-        MapOrString::String(raw) => serde_json::from_str(&raw).map_err(serde::de::Error::custom),
-    }
+    deserializer.deserialize_any(MapOrString(PhantomData))
 }
 
 impl AzureDnsConfig {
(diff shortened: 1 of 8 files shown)


Hermann-Core (approved):
**Approved** The ACME-DNS multi-domain support is correctly implemented: ✓ Per-domain credential selection works as specified ✓ Backward compatible: single-account config unchanged ✓ Two-value limit enforced at startup via `check_order_domains()` ✓ Comprehensive wiremock tests cover all scenarios ✓ Config validation catches misconfigurations at boot ✓ Domain normalization is consistent and correct ✓ Docs updated with clear examples The implementation handles all edge cases (conflicting accounts, empty fields, apex+wildcard sharing accounts) and surfaces configuration errors at startup rather than at renewal time. Tests pass, code is idiomatic.

Hermann-Core:
@martcpp fix the conflicts on this branch and let's merge this ticket.

martcpp:
> @martcpp fix the conflicts on this branch and let's merge this ticket.

on it

martcpp:
@ndefokou check this pr let merge thanks

src/utils/state.rs
```diff
@@ -73,7 +83,11 @@ pub async fn build_state(config: &AppConfig) -> EyeResult<AppState> {
              but APP_ENV=production; ACME challenges will not succeed against a real CA"
         );
     }
-    let challenge_handler = build_dns_challenge_handler(dns_provider, config, &aws_config).await?;
+    // Domains certificates are ordered for; the single source for both the
+    // ACME order and the challenge handler's startup coverage checks
+    let cert_domains = [config.server.domain.as_str()];
```
ndefokou, line 109: The only ACME order identifier today is `config.server.domain`: `CertManager::new` is constructed from `cert_domains` (this slice) and `ServerConfig` exposes no SAN / multi-domain list. So at runtime this slice — and therefore `AcmeDnsProvider::check_order_domains` — always contains exactly one identifier. That means the ticket's core scenario (3+ identifiers sharing an ACME-DNS two-value window) cannot occur or be caught via the live order path; the per-domain `accounts` map also provides no capability beyond the default account for a single-domain order (you'd just set the default account).

The multi-identifier behavior is only exercised by tests that call the provider directly with synthetic arrays (e.g. `acme_dns.rs` `three_identifiers_...`, `state.rs` `acme_dns_rejects_three_cert_domains_on_one_account`).

Is multi-domain/SAN cert config planned (a #162 follow-up)? If so, should this PR also land the domain-list config so the feature is actually reachable end-to-end, rather than shipping plumbing that's dead for the real order path?

src/utils/state.rs
```diff
@@ -183,12 +201,20 @@ async fn build_dns_challenge_handler(
...
-                &cfg.username,
-                cfg.password.clone(),
-                &cfg.subdomain,
-            ))
+                cfg.default_account().as_ref().map(acme_dns_credentials),
+                accounts,
+            )?;
+            // Catch a credentials gap or an overloaded two-value TXT window
+            // for the ordered domains at startup instead of at the first renewal
+            provider.check_order_domains(cert_domains)?;
```
ndefokou, line 338: This guard runs only here, once at boot, against the static config domains. It is not invoked from the actual order path (`CertManager::request_certificate` / `handle_authorization`). Given identifiers come straight from config today that's safe, but the contract is 'startup sanity check', not 'per-order enforcement' — worth a one-line doc comment on `check_order_domains` saying so, so it isn't later assumed to protect against runtime domain/account mismatches (e.g. after a config reload/rebuild).

src/utils/cert_manager/challenge/dns01/acme_dns.rs
```diff
@@ -30,34 +43,120 @@ fn dns_err(source: impl Into<Report>) -> ChallengeError {
     }
 }
 
+/// Normalize a domain for account lookup (lowercase, no trailing dot, no
+/// wildcard label) so config keys and ACME identifiers compare equal
+fn normalize_domain(domain: &str) -> String {
```
ndefokou, line 51: `normalize_domain` only strips `*.`, trailing dots and ASCII-lowercases; it does **not** perform IDNA/unicode→punycode conversion. It relies on the operator writing config keys in the same form instant-acme hands back the identifier (punycode A-labels), which the docs do call out. Worth a brief comment on the function noting that two keys that are the *same* domain in different encodings (unicode vs `xn--`) will not match and will be treated as 'no account', so the assumption is explicit rather than implicit.

src/config.rs
```diff
@@ -102,26 +102,155 @@ pub struct CloudflareDnsConfig {
...
+        if self.default_account().is_none() && self.accounts.is_empty() {
+            return Err(ConfigError::Message(
+                "ACME-DNS settings need a default account (username/password/subdomain) \
+                 or a non-empty accounts map"
+                    .to_string(),
+            ));
+        }
+        // Reject unusable per-domain entries here instead of as an opaque
+        // HTTP 401 at the first renewal
+        for (domain, account) in &self.accounts {
```
ndefokou, line 250: Per-account field-emptiness and 'does not name a domain' checks live here, but the normalization-conflict check (two keys reducing to the same domain with different credentials) lives in `AcmeDnsProvider::new` (acme_dns.rs). Entries like `example.com` + `EXAMPLE.com.` pass `AcmeDnsConfig::validate()` and only fail later when the provider is built in `build_dns_challenge_handler`. Both run at startup so the operator still sees an error, but the same class of config mistake surfaces from two different layers with inconsistent messages. Consider normalizing keys inside `validate()` and rejecting conflicts here too, so all ACME-DNS config errors come from one place.

src/utils/cert_manager/challenge/dns01/acme_dns.rs
```diff
@@ -149,9 +281,263 @@ mod tests {
...
+        .unwrap();
+
+        provider
+            .create_txt_record("mapped.example.com", "digest-value")
+            .await
+            .unwrap();
+    }
+
+    #[test]
+    fn check_order_domains_rejects_three_identifiers_on_one_account() {
```
ndefokou: Issue #207's acceptance criteria asks for a **wiremock** test covering 'the 3-identifier case that currently fails'. The tests added here cover: (1) 3 identifiers spread across 3 accounts each hitting their own subdomain (line 318, wiremock), and (2) the startup guard rejecting 3-on-one-account (this test, line 380, no `MockServer`).

Neither directly simulates the original failure mode — three digests written to the *same* acme-dns subdomain where the earliest is rotated out of the two-value TXT window before CA validation. (2) tests the guard, not the failure it prevents. Consider at least one wiremock test that mounts a single `/update` mock expecting three writes and asserts the guard/order flow stops it before the rotation, so the 'currently fails' scenario is covered rather than only the rejection branch.

ndefokou (commented):
Reviewed against issue #207. Implementation is solid and well-tested at the provider/config layer, but I have a few concerns — the main one being that the feature is effectively unreachable end-to-end because the ACME order is always for the single `server.domain`. Details inline.

src/utils/cert_manager/challenge/dns01/acme_dns.rs
martcpp: Rotation is acme-dns server state; wiremock is stateless — the proposed assertion targets a flow that doesn't exist

src/config.rs
martcpp: Provider must keep its own invariant check anyway → proposal creates duplication and a layering inversion, for a cosmetic gain

martcpp:
@ndefokou recheck

ndefokou (approved):
LGTM 🚀

martcpp:
@ndefokou recheck i just rebase to main thanks

ndefokou (approved):

PR #219, opened 2026-07-16
Refactor Runtime Environment Variable

### LINKED ISSUE #215: Investigate and Refactor Runtime Environment Variable Overrides Following std::env::set_var Safety Changes
Recent Rust toolchain updates have made `std::env::set_var` and `std::env::remove_var` `unsafe` APIs due to the possibility of undefined behavior when modifying process-wide environment variables in multithreaded applications.

Our project currently relies on overriding environment variables during testing. As a project requirement, we do **not** want to introduce `unsafe` code to support this behavior.

This task is to identify a safe alternative that preserves the existing testing capabilities without relying on `std::env::set_var` or any other `unsafe` environment mutation APIs.

## Tasks

* [x] Identify all locations where runtime environment variables are overridden during tests.
* [x] Analyze the purpose of each override and the expected behavior being tested.
* [x] Research and evaluate safe alternatives to runtime environment mutation (e.g., dependency injection, configuration providers, mock configuration sources, or isolated child processes).
* [x] Design and implement a solution that removes the dependency on `std::env::set_var` and `std::env::remove_var`.
* [x] Refactor existing tests to use the new configuration mechanism.
* [x] Verify that the new approach supports overriding configuration values for test scenarios without modifying the process-wide environment.
* [x] Remove all remaining dependencies on `std::env::set_var` and `std::env::remove_var`.
* [x] Update developer documentation to describe the new testing approach.

## Acceptance Criteria

* No `unsafe` code is introduced as part of this work.
* Tests can safely override configuration values without mutating the process-wide environment.
* Existing test coverage is maintained or improved.
* The project builds and all tests pass on the latest supported Rust toolchain.
* Documentation clearly describes the new mechanism for providing test-specific configuration.

est: 1.5 day

Blindspot22:
Investigate and Refactor Runtime Environment Variable Overrides Following std::env::set_var Safety Changes


README.md
```diff
@@ -165,6 +165,35 @@ You can run the tests using the following command:
...
+])
+.expect("Failed to load config");
+```
+
+Each `path` uses `.` as the nested-key separator and mirrors the production
+`APP_*` environment variable (minus the `APP_` prefix and with `__` replaced
+by `.`). The textual `value` is interpreted exactly as the corresponding
+environment variable would be, so any valid `APP_*` value is also valid here.
+Pass `&[]` (no overrides) to obtain the production defaults.
+
```
Hermann-Core: We don't want to expose internal implementation details publicly, this section should be removed.

src/config.rs
```diff
@@ -364,17 +333,107 @@ impl Config {
...
+        .set_default("aws.region", "us-east-1")
+        .expect("hardcoded default")
+        .set_default("cache.ttl", 5 * 60)
+        .expect("hardcoded default")
+        .set_default("cache.max_capacity", 100)
+        .expect("hardcoded default")
+        .set_default("status_list.token_exp_secs", 900)
+        .expect("hardcoded default")
+        .set_default("status_list.token_ttl_secs", 300)
+        .expect("hardcoded default")
```
Hermann-Core: Could you propagate the error here instead of using the panicking `expect`as it was implemented previously, I think using the `?` operator to propagate the error would be more appropriate. This avoids unnecessary panics and allows the caller to handle the error according to the surrounding context.

Hermann-Core (commented):
Hello @Blindspot22, thanks for fixing this. Can you check the comments I left below?

Blindspot22:
@Hermann-Core and @Ngha-Boris please check this again

README.md
Hermann-Core: This section is still present in `README.md`. Public documentation should focus on end-user and operator usage. Internal unit test implementation details (`Config::load_from_overrides`, `std::env::set_var`, `src/config.rs`) should not be added to the top-level README. Please remove this addition from `README.md`.

src/config.rs
Hermann-Core: `base_builder()` still uses `.expect("hardcoded default")` repeatedly across 40+ default settings. Please have `base_builder()` return `Result<ConfigBuilder<DefaultState>, ConfigError>` and propagate errors using the `?` operator (matching `Config::load()`), rather than panicking with `.expect()`.

src/config.rs
```diff
@@ -120,11 +125,6 @@ pub struct CertConfig {
     #[serde(default)]
```
Hermann-Core, line 130: Removing the doc comments on `chain_cache_ttl`, `cert_cache_ttl`, `secrets_cache_ttl`, and `ttl` removes critical domain documentation (such as the distinction between `ttl = 0` disabling caching vs `chain_cache_ttl = 0` representing infinite TTL). Please restore these doc comments.

src/config.rs
```diff
@@ -660,6 +655,134 @@ impl Config {
...
+///
+/// Both production loading (via [`Config::load`]) and test loading (via
+/// `Config::load_from_overrides`, gated on `#[cfg(test)]`) start from this
+/// shared set of defaults so that there is exactly one source of truth for the
+/// default configuration.
+#[cfg(test)]
+type TestConfigBuilder = ConfigBuilder<DefaultState>;
+
+#[cfg(test)]
+#[allow(dead_code)]
```
Hermann-Core: `Config::load()` and `base_builder()` currently maintain two separate, hardcoded copies of all 40+ default configuration settings. This creates two sources of truth that will silently drift whenever configuration defaults are updated in production.

Refactor `Config::load()` and `load_from_overrides()` to share a single un-gated `base_builder()` helper function so that production and test loading always start from the exact same defaults.

src/config.rs
```diff
@@ -660,6 +655,134 @@ impl Config {
...
+    /// and `value` is the textual representation of the desired setting. This
+    /// mirrors how environment variables are interpreted, so any value that is
+    /// valid as an `APP_*` environment variable is also valid here.
+    ///
+    /// This helper exists so that tests can exercise configuration loading and
+    /// overrides without relying on `std::env::set_var` / `std::env::remove_var`,
+    /// which became `unsafe` on recent Rust toolchains. Because nothing mutates
+    /// the process-wide environment, tests are fully isolated from each other
+    /// and from the host environment regardless of execution order or threading.
+    #[cfg(test)]
```
Hermann-Core, line 638: The `overrides` parameter is currently constrained to `&[(&'static str, &'static str)]`. Using `&[(&str, &str)]` allows callers to pass dynamically formatted strings in tests without requiring static lifetime promotion.

Hermann-Core (changes requested):
@Blindspot22 Please review the inline comments and un-resolved review threads for required updates.

README.md
Hermann-Core: This section is still present in `README.md`. Public documentation should focus on end-user and operator usage. Internal unit test implementation details (`Config::load_from_overrides`, `src/config.rs`) should not be in the top-level README. Please remove this addition from `README.md`.

src/config.rs
Hermann-Core: `base_builder()` still uses `.expect("hardcoded default")` repeatedly across 40+ default settings. Please have `base_builder()` return `Result<ConfigBuilder<DefaultState>, ConfigError>` and propagate errors using the `?` operator (matching `Config::load()`), rather than panicking with `.expect()`.

src/config.rs
Hermann-Core: The doc comments on `chain_cache_ttl`, `cert_cache_ttl`, `secrets_cache_ttl`, and `ttl` in `src/config.rs` are still missing. Removing these doc comments removes critical domain documentation (e.g., explaining `ttl = 0` vs `chain_cache_ttl = 0`). Please restore these doc comments.

src/config.rs
Hermann-Core: `Config::load()` and `base_builder()` currently maintain two separate, hardcoded copies of all 40+ default configuration settings. This creates two sources of truth that will silently drift whenever configuration defaults are updated in production. Refactor `Config::load()` and `load_from_overrides()` to share a single un-gated `base_builder()` helper function so that production and test loading always start from the exact same defaults.

src/config.rs
Hermann-Core: The `overrides` parameter is currently constrained to `&[(&'static str, &'static str)]`. Using `&[(&str, &str)]` allows callers to pass dynamically formatted strings in tests without requiring static lifetime promotion. Please update the type signature to `&[(&str, &str)]`.

Hermann-Core (dismissed):
@Blindspot22 Please review the open review comments and thread replies regarding unaddressed feedback.

Blindspot22:
@martcpp Please check this too

README.md
Hermann-Core: Verified. The internal test configuration section has been removed from README.md.

src/config.rs
Hermann-Core: Verified. `base_builder()` now returns `Result<ConfigBuilder<DefaultState>, ConfigError>` and propagates errors using the `?` operator.

src/config.rs
Hermann-Core: Verified. Doc comments for `chain_cache_ttl`, `cert_cache_ttl`, `secrets_cache_ttl`, and `ttl` have been restored.

src/config.rs
Hermann-Core: Verified. `Config::load()` and `Config::load_from_overrides()` now share a single `base_builder()` implementation.

src/config.rs
Hermann-Core: Verified. The `overrides` parameter signature has been updated to `&[(&str, &str)]`.

Hermann-Core (approved):
@Blindspot22 Approved - all review comments have been verified and resolved.

Blindspot22:
And Please @martcpp, this PR should be merged only after @Ngha-Boris approval.
I made some of his changes and I need him to approve everything

martcpp (approved):

PR #234, opened 2026-07-20
Guard update_status against concurrent lost updates with optimistic u…

### LINKED ISSUE #223: Lost update in update_status: concurrent PATCHes silently drop status flips
## Summary
`update_status` performs a plain read-modify-write with no concurrency guard. Two concurrent `PATCH` requests to the same status list can each read the same list state, each recompute it in memory, and each write
back  with the last writer silently erasing the first writer's changes.

## Reproduction (conceptual)

1. Two clients issue `PATCH .../statuses` on the same `list_id` at ~the same time.
2. Both handlers execute `find_one_by` and read identical `lst`.
3. Client A flips index `i`; Client B flips index `j`.
4. Both call `update_one` with their own recomputed list.
5. The second write to land overwrites the first. One of the two flips (`i` or `j`) is silently lost.

## Proposed fix

Optimistic concurrency on `updated_at`:

- Include the previously-read `updated_at` in the `UPDATE ... WHERE id = ? AND updated_at = ?` clause.
- If **rows-affected = 0**, another writer won the race → return **409 Conflict**
  with a retry hint, rather than reporting success.
- The client (or handler) re-reads and retries.

### Cross-backend constraint (important  read before implementing)

Since the configurable-database work (#143) merged, this handler must behave identically on **Postgres, MySQL, and SQLite**. This is precisely why the `updated_at`-guard / rows-affected approach is the right one:

- `rows_affected` semantics are portable across all three sea-orm backends.
- It relies on **no backend-specific locking** (unlike `SELECT ... FOR UPDATE`, which diverges  SQLite has no row-level locking, and behavior differs  on MySQL vs Postgres).

Do **not** reach for a `FOR UPDATE` transactional lock here; it would reintroduce the cross-backend divergence #143  exists to avoid.

## Secondary (smaller, bounded)

The invalidate-after-write cache step opens a much smaller window where a concurrent reader can repopulate the cache with the pre-write value just before invalidation. This is bounded by the cache TTL and far less severe
than the lost update above. Worth noting here, but the `updated_at` guard is the primary fix; the cache window can be addressed as a follow-up if it proves to matter in practice.

## Acceptance criteria

- [ ] Concurrent `PATCH`es to the same list can no longer silently lose a flip.
- [ ] A conflicting write returns `409` (with retry guidance), not a false success.
- [ ] Behavior is identical on Postgres, MySQL, and SQLite (rows-affected-based, no backend-specific locking).
- [ ] A regression test drives two interleaved updates and asserts neither flip is lost (or the loser gets a 409).

martcpp:
`update_status` performs a plain read-modify-write with no concurrency guard. Two concurrent `PATCH` requests to the same status list can each read the same list state, each recompute it in memory, and each write
back  with the last writer silently erasing the first writer's changes.

## Acceptance criteria

- [ ] Concurrent `PATCH`es to the same list can no longer silently lose a flip.
- [ ] A conflicting write returns `409` (with retry guidance), not a false success.
- [ ] Behavior is identical on Postgres, MySQL, and SQLite (rows-affected-based, no backend-specific locking).
- [ ] A regression test drives two interleaved updates and asserts neither flip is lost (or the loser gets a 409).

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -271,6 +271,8 @@ paths:
           $ref: "#/components/responses/Forbidden"
         "404":
           $ref: "#/components/responses/NotFound"
+        "409":
+          $ref: "#/components/responses/UpdateConflict"
         "413":
           $ref: "#/components/responses/PayloadTooLarge"
         "422":
@@ -747,6 +749,23 @@ components:
           schema:
             type: string
           example: "Too Many Requests! Wait for 42s"
+
+    UpdateConflict:
+      description: |
+        The status list was modified concurrently between the client's read and
+        this write (optimistic-concurrency conflict): another writer's update
+        landed first, so this request was rejected rather than silently
+        overwriting it. The client should re-read the current state and retry.
+      content:
+        application/problem+json:
+          schema:
+            $ref: "#/components/schemas/ProblemDetails"
+          example:
+            type: "about:blank"
+            title: "Conflict"
+            status: 409
+            detail: "the status list was modified concurrently; re-read the current state and retry"
+            instance: "/api/v1/status-lists/477121aa-b598-419e-916f-1e74654ff38b/statuses"
 
     ServiceUnavailable:
       description: Service temporarily unavailable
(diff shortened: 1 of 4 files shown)


martcpp:
@Christiantyemele @Ngha-Boris @Hermann-Core pls check this

Hermann-Core (approved):
Thorough optimistic concurrency fix with excellent test coverage

PR #242, opened 2026-07-21
231 store strategy deployments run and can fail on dns provider resolution they never use

### LINKED ISSUE #231: Store-strategy deployments run (and can fail on) DNS provider resolution they never use
## Summary
At startup, `build_state` resolves **and constructs** the DNS provider **unconditionally**, *before* it branches on the provisioning strategy. A deployment using `provisioning_strategy = "store"` — which reads a
certificate from disk or secrets manager and never talks to ACME  has no need for a DNS provider at all, yet still pays for full DNS resolution and construction. The resulting challenge handler is then discarded unused.

## Proposed fix

Reorder `build_state` so it determines the provisioning strategy **first** and only resolves/constructs the DNS provider when the **ACME** strategy is selected. Store-strategy deployments should never touch DNS resolution.

Rough shape:

- Branch on `uses_acme_strategy` (or equivalent) before DNS work.
- ACME path → `resolve()` + `build_dns_challenge_handler(...)` as today.
- Store path → skip DNS resolution entirely; no DNS section required.

## Explicitly NOT part of this

- No change to `resolve()` / `ResolvedDnsProvider` / the resolve-build seam  this is `build_state` branch *ordering*, a different concern. The recent typed-resolver refactor neither caused this nor can fix it.
- No change to how either provisioning strategy works, only to when DNS  resolution is invoked.

## Acceptance criteria

- [x] A `store`-strategy deployment with **no DNS section configured** boots successfully.
- [x] A `store`-strategy deployment with a **malformed DNS section** boots successfully (DNS is not consulted).
- [x] The ACME strategy still resolves and constructs the DNS provider exactly as before (no behavior change on that path).
- [x] A test boots `build_state` under the store strategy with absent/invalid DNS config and asserts success.

ndefokou:
Here is a short and concise PR description you can use:

### Title
`fix: defer DNS resolution to ACME provisioning path`

### Description

**Summary**
This PR fixes an issue where `store`-strategy deployments would fail at boot if their DNS configuration was absent or malformed. 

Previously, `build_state` resolved the DNS provider and constructed the DNS challenge handler unconditionally at startup, before checking the provisioning strategy. The resulting handler was then discarded if the `store` strategy was used. 

**Changes**
- Reordered `build_state` in `src/utils/state.rs` so that the `uses_acme_strategy` check happens first. 
- DNS resolution and challenge handler construction now strictly only occur on the ACME path. Store-strategy deployments no longer touch DNS resolution at all.
- Added a new unit test `store_strategy_ignores_dns_config` to verify that `store` deployments boot successfully regardless of missing or broken DNS configurations.

**Acceptance Criteria Met**
- [x] A `store`-strategy deployment with no DNS section boots successfully.
- [x] A `store`-strategy deployment with a malformed DNS section boots successfully.
- [x] The ACME strategy still resolves and constructs the DNS provider exactly as before.
- [x] A test boots `build_state` under the store strategy with absent/invalid DNS config and asserts success.


src/utils/state.rs
```diff
@@ -153,6 +139,24 @@ pub async fn build_state(config: &AppConfig) -> EyeResult<AppState> {
         .eku(&config.server.cert.eku);
 
     cert_manager_builder = if uses_acme_strategy {
+        // DNS resolution is only required for the ACME strategy, which needs a
+        // DNS-01 challenge handler to complete certificate orders. Store-strategy
+        // deployments read a certificate from disk or Secrets Manager and never
+        // talk to ACME, so they must never pay for — or fail on — DNS resolution.
```
Hermann-Core: I think this is obvious to see and should not be documented. Don't document what have been requested in reviews.

src/utils/state.rs
```diff
@@ -692,6 +696,30 @@ mod tests {
         }
     }
 
+    /// A store-strategy deployment must boot successfully even when the DNS
+    /// section is absent or malformed — DNS resolution is not consulted on
+    /// that path (acceptance criteria for issue #231).
```
Hermann-Core: Why are you documenting what was requested during reviews? It feels automated.

Hermann-Core (approved):

PR #244, opened 2026-07-21
Wrap status list update

### LINKED ISSUE #235: Wrap status-list `UPDATE` + history `INSERT` in a single transaction
## Summary

`update_status` performs two independent writes with no transaction around
them:
1. The guarded `UPDATE status_lists ... WHERE updated_at = ?` (the current row).
2. The `INSERT` into `status_list_history` (the point-in-time snapshot).

If the `UPDATE` succeeds and the history `INSERT` then fails, the row has changed but **no snapshot records the change**, and the handler returns 500.

## Proposed fix

Wrap both writes in a single database transaction:

- Begin transaction.
- Execute the guarded `UPDATE` (with the `updated_at` optimistic guard).
- If `rows_affected == 0` → conflict, roll back, return 409 (as today).
- Execute the history `INSERT`.
- Commit.

Transactions are cross-backend-safe (Postgres/MySQL/SQLite), so this does **not** conflict with the #143 identical-behavior constraint  unlike `SELECT ... FOR UPDATE`, which is why that was rejected for the guard
itself.

## Explicitly out of scope

- No change to the optimistic-lock guard logic itself (that's already landed).
- No schema change (`updated_at` and `status_list_history` already exist).
- No `SELECT ... FOR UPDATE` / row locking.

## Acceptance criteria

- [ ] `UPDATE status_lists` and the `status_list_history` `INSERT` succeed or fail as a unit.
- [ ] A forced history-insert failure rolls back the row update (test: inject INSERT failure, assert the row is unchanged and no partial snapshot exists).
- [ ] The 409 conflict path still rolls back cleanly and records nothing.
- [ ] Behavior verified on Postgres, MySQL, and SQLite (transaction semantics are portable, but confirm the rollback path on at least MySQL + SQLite).

martcpp:
## Summary

`update_status` performs two independent writes with no transaction around
them:
1. The guarded `UPDATE status_lists ... WHERE updated_at = ?` (the current row).
2. The `INSERT` into `status_list_history` (the point-in-time snapshot).

If the `UPDATE` succeeds and the history `INSERT` then fails, the row has changed but **no snapshot records the change**, and the handler returns 500.

## Acceptance criteria

- [ ] `UPDATE status_lists` and the `status_list_history` `INSERT` succeed or fail as a unit.
- [ ] A forced history-insert failure rolls back the row update (test: inject INSERT failure, assert the row is unchanged and no partial snapshot exists).
- [ ] The 409 conflict path still rolls back cleanly and records nothing.
- [ ] Behavior verified on Postgres, MySQL, and SQLite (transaction semantics are portable, but confirm the rollback path on at least MySQL + SQLite).

### DIFF
--- a/src/adapters/sea_orm/store.rs
+++ b/src/adapters/sea_orm/store.rs
@@ -1,6 +1,6 @@
 use sea_orm::{
     ActiveModelTrait, ColumnTrait, DatabaseConnection, EntityTrait, QueryFilter, QueryOrder,
-    QuerySelect, Set, sea_query::Expr,
+    QuerySelect, Set, TransactionTrait, sea_query::Expr,
 };
 use std::sync::Arc;
 
@@ -126,6 +126,95 @@ impl SeaOrmStore<StatusListRecord> {
             .await
             .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
         Ok(result.rows_affected > 0)
+    }
+
+    /// Optimistic update that records a history snapshot in the **same
+    /// transaction** as the guarded row update.
+    ///
+    /// Either the guarded `UPDATE status_lists` and the `status_list_history`
+    /// `INSERT` both commit, or neither does. This closes the split the plain
+    /// [`update_one`](Self::update_one) leaves open, where the row changes but a
+    /// subsequent snapshot insert fails, so nothing records the change.
+    ///
+    /// A `false` return means the optimistic guard did not match (a racing
+    /// writer advanced the stamp, or the row is gone): the transaction is rolled
+    /// back and nothing is written — identical outward behavior to
+    /// [`update_one`](Self::update_one). Transaction semantics are portable
+    /// across the Postgres/MySQL/SQLite sea-orm backends (#143), so the
+    /// all-or-nothing guarantee holds identically on all three.
+    ///
+    /// The same strictly-advancing `updated_at` caller contract as
+    /// [`update_one`](Self::update_one) applies and is enforced here too.
+    pub async fn update_one_with_snapshot(
+        &self,
+        list_id: &str,
+        entity: StatusListRecord,
+        expected_updated_at: i64,
+        snapshot: StatusListHistoryRecord,
+    ) -> Result<bool, RepositoryError> {
+        if entity.updated_at <= expected_updated_at {
+            return Err(RepositoryError::UpdateError(format!(
+                "guarded update requires a strictly newer updated_at \
+                 (new={}, expected-guard={}); a non-advancing stamp would \
+                 silently reintroduce the same-second lost update",
+                entity.updated_at, expected_updated_at
+            )));
+        }
+
+        let txn = self
+            .db
+            .begin()
+            .await
+            .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+
+        let result = status_lists::Entity::update_many()
+            .col_expr(status_lists::Column::Issuer, Expr::value(entity.issuer))
+            .col_expr(
+                status_lists::Column::StatusList,
+                Expr::value(entity.status_list),
+            )
+            .col_expr(status_lists::Column::Sub, Expr::value(entity.sub))
+            .col_expr(
+                status_lists::Column::UpdatedAt,
+                Expr::value(entity.updated_at),
+            )
+            .filter(status_lists::Column::ListId.eq(list_id))
+            .filter(status_lists::Column::UpdatedAt.eq(expected_updated_at))
+            .exec(&txn)
+            .await
+            .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+
+        if result.rows_affected == 0 {
+            // Optimistic-guard miss: roll back so the conflict path records
+            // nothing (no snapshot, no row change).
+            txn.rollback()
+                .await
+                .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+            return Ok(false);
+        }
+
+        let history_active: status_list_history::ActiveModel = snapshot.into();
+        if let Err(insert_err) = status_list_history::Entity::insert(history_active)
+            .exec(&txn)
+            .await
+        {
+            // The row UPDATE already landed inside the transaction, but the
+            // snapshot INSERT failed. Roll back so the row reverts to its
+            // pre-update state — the whole point of this method: never leave a
+            // changed row without the snapshot that records the change.
+            txn.rollback().await.map_err(|rollback_err| {
+                RepositoryError::InsertError(format!(
+                    "history snapshot insert failed ({insert_err}); \
+                     rolling back the row update also failed: {rollback_err}"
+                ))
+            })?;
+            return Err(RepositoryError::InsertError(insert_err.to_string()));
+        }
+
+        txn.commit()
+            .await
+            .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+        Ok(true)
     }
 
     pub async fn delete_by(&self, value: &str) -> Result<bool, RepositoryError> {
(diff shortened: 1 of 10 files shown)


src/database/queries.rs
```diff
@@ -63,58 +139,96 @@ impl SeaOrmStore<StatusListRecord> {
...
+    /// across the snapshot `INSERT` as well, where previously the statement
+    /// auto-committed and released immediately. A concurrent writer to the same
+    /// `list_id` therefore *blocks* on the lock rather than seeing
+    /// `rows_affected == 0` straight away; it still resolves to the same
+    /// conflict once the winner commits, but later. On MySQL a queue of writers
+    /// on one hot list can reach `innodb_lock_wait_timeout` (50s by default),
+    /// which surfaces as an error — a 500 — rather than a conflict. If that
+    /// becomes real, the mitigation is to insert the snapshot *before* the
+    /// guarded update within this transaction, shortening the row-lock window
+    /// to the final statement.
```
Hermann-Core: We should avoid over-documenting public APIs, or at least in appropriate locations when applicable. Comments should be brief and focus on important invariants, architectural details, or other long-term relevant information. They should not contain minor implementation details

src/database/queries.rs
```diff
@@ -25,6 +25,82 @@ impl<T> SeaOrmStore<T> {
...
+    entity: StatusListHistoryRecord,
+) -> Result<(), RepositoryError> {
+    let active: status_list_history::ActiveModel = entity.into();
+    status_list_history::Entity::insert(active)
+        .exec(conn)
+        .await
+        .map_err(|e| RepositoryError::InsertError(e.to_string()))?;
+    Ok(())
+}
+
```
Hermann-Core: Could you move these helper functions below the public items? By convention, helper functions are typically placed after the public API to make the exposed interface easier to identify and improve the overall file organization.

src/database/queries.rs
```diff
@@ -63,58 +139,96 @@ impl SeaOrmStore<StatusListRecord> {
...
+                    .await
+                    .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+                Ok(true)
+            }
+            // Guard miss or failed write: nothing this transaction did may
+            // survive. Rolling back explicitly rather than leaning on the
+            // drop-rollback keeps the discard visible at the call site. A
+            // failed rollback is logged, never returned, so it cannot mask the
+            // reason we are rolling back.
+            outcome => {
```
Ngha-Boris: **Issue: Silent rollback failure**

The rollback error is only logged but doesn't propagate. If rollback fails, this masks a potentially serious database inconsistency where the transaction may have partially committed.

**Suggestion:**
```rust
outcome => {
    if let Err(e) = txn.rollback().await {
        tracing::error!(error = %e, %list_id, "CRITICAL: rollback failed");
        return Err(RepositoryError::TransactionError(
            format!("Rollback failed: {e}")
        ));
    }
    outcome
}
```

src/database/queries.rs
```diff
@@ -25,6 +25,82 @@ impl<T> SeaOrmStore<T> {
...
+/// did not match — another writer changed the row (or it was deleted) since
+/// the caller read `expected_updated_at`, i.e. a lost-update was prevented.
+///
+/// `rows_affected` is used deliberately: its semantics are identical across
+/// the Postgres/MySQL/SQLite sea-orm backends, unlike `SELECT ... FOR UPDATE`
+/// row locking (see #143).
+///
+/// Callers must have cleared [`ensure_stamp_advances`] first; this function
+/// executes the statement and does not re-check the contract.
+async fn execute_guarded_update<C: ConnectionTrait>(
```
Ngha-Boris: **Issue: Unnecessary ownership**

This function takes `entity: StatusListRecord` by value but only needs references to its fields. This forces callers to clone unnecessarily.

**Suggestion:**
```rust
async fn execute_guarded_update<C: ConnectionTrait>(
    conn: &C,
    entity: &StatusListRecord,  // Borrow instead
    expected_updated_at: i64,
    let list_id = &entity.list_id;
    // ...
}
```

Update the call site:
```rust
if !execute_guarded_update(&txn, &entity, expected_updated_at).await? {
```

src/database/queries.rs
```diff
@@ -747,6 +852,273 @@ mod test {
...
+        .await
+        .expect("transactional update deadlocked on the single-connection pool");
+    }
+
+    /// Cross-backend proof: rollback is the load-bearing behavior here, and
+    /// transaction semantics are only *claimed* to be portable — MySQL runs the
+    /// same failing-insert scenario to show the row is restored there too.
+    #[cfg(feature = "mysql")]
+    #[tokio::test]
+    async fn test_mysql_update_one_with_snapshot_rolls_back_when_history_insert_fails() {
```
Ngha-Boris: **Missing: PostgreSQL test**

The acceptance criteria requires verification on "Postgres, MySQL, and SQLite". I see tests for SQLite and MySQL, but no PostgreSQL test.

**Suggestion:** Add a PostgreSQL test similar to `test_mysql_update_one_with_snapshot_rolls_back_when_history_insert_fails`:

```rust
#[cfg(feature = "postgres")]
#[tokio::test]
async fn test_postgres_update_one_with_snapshot_rolls_back_when_history_insert_fails() {
    // Same pattern as MySQL test
}
```

src/adapters/sea_orm/store.rs
```diff
@@ -63,58 +139,96 @@ impl SeaOrmStore<StatusListRecord> {
...
-                status_lists::Column::UpdatedAt,
-                Expr::value(entity.updated_at),
-            )
-            .filter(status_lists::Column::ListId.eq(list_id))
-            .filter(status_lists::Column::UpdatedAt.eq(expected_updated_at))
-            .exec(&*self.db)
+        ensure_stamp_advances(entity.updated_at, expected_updated_at)?;
+
+        let list_id = entity.list_id.clone();
+        let txn = self
```
Ngha-Boris, line 217: **Issue: Wrong error variant**

Transaction begin failures are mapped to `UpdateError`, but a transaction failure is distinct from an update failure.

**Suggestion:**
```rust
let txn = self
    .db
    .begin()
    .await
    .map_err(|e| RepositoryError::TransactionError(
        format!("Failed to begin transaction: {e}")
    ))?;
```

Or at minimum:
```rust
.map_err(|e| RepositoryError::UpdateError(
    format!("Transaction begin failed: {e}")
))?;
```

src/database/queries.rs
```diff
@@ -747,6 +852,273 @@ mod test {
...
 
+    /// Ceiling for the transactional tests below. On a single-connection pool
+    /// (SQLite) an open transaction holds the only connection, so a write
+    /// accidentally routed to the pool instead of the transaction handle
+    /// *deadlocks* rather than erroring. Without a deadline that surfaces as a
+    /// hung test — indistinguishable from a slow one until the whole run dies
+    /// at the job timeout with no attribution. Bounding it turns that into one
+    /// named failure pointing straight at the offending test.
+    #[cfg(any(feature = "sqlite", feature = "mysql"))]
+    const TXN_TEST_DEADLINE: std::time::Duration = std::time::Duration::from_secs(10);
```
Ngha-Boris: **Suggestion: Configurable timeout**

The 10-second timeout may cause flaky failures in CI or slow container environments.

**Consider:**
```rust
const TXN_TEST_DEADLINE: std::time::Duration = std::time::Duration::from_secs(
    std::option_env!("TXN_TEST_DEADLINE_SECS")
        .and_then(|s| s.parse().ok())
        .unwrap_or(10)
);
```

Ngha-Boris (changes requested):
Hello @martcpp please check the comments below

src/database/queries.rs
Hermann-Core: If `txn.rollback()` fails, the error is only logged but the original error is still returned. This can mask a serious database state problem where the transaction may have partially committed. A failed rollback could leave orphaned locks or inconsistent state that subsequent operations depend on.

**Fix**: Return the rollback error to signal that database state is unknown:
```rust
outcome => {
    if let Err(rollback_err) = txn.rollback().await {
        tracing::error!(
            error = %rollback_err,
            %list_id,
            "CRITICAL: rollback failed after guarded update error"
        );
        // Database state is now unknown - propagate the rollback failure
        return Err(RepositoryError::UpdateError(format!(
            "Rollback failed: {rollback_err}"
        )));
    }
    outcome
}
```

Alternatively, if you want to preserve the original error, at least panic or abort since continuing with unknown DB state is unsafe.

src/adapters/sea_orm/store.rs
```diff
@@ -63,58 +139,96 @@ impl SeaOrmStore<StatusListRecord> {
...
-            .col_expr(status_lists::Column::Sub, Expr::value(entity.sub))
-            .col_expr(
-                status_lists::Column::UpdatedAt,
-                Expr::value(entity.updated_at),
-            )
-            .filter(status_lists::Column::ListId.eq(list_id))
-            .filter(status_lists::Column::UpdatedAt.eq(expected_updated_at))
-            .exec(&*self.db)
+        ensure_stamp_advances(entity.updated_at, expected_updated_at)?;
+
```
Hermann-Core, line 216: This clones the `list_id` (a String) before the transaction begins, but it's only used for logging in the rollback error path (line 218). The clone happens on every call, even successful ones.

**Impact**: Every status update pays the allocation cost of cloning the list_id String, even though it's only needed in the rare error case.

**Fix**: Move the clone into the error path where it's actually needed:
```rust
pub async fn update_one_with_snapshot(
    &self,
    entity: StatusListRecord,
    expected_updated_at: i64,
    snapshot: Option<StatusListHistoryRecord>,
    ensure_stamp_advances(entity.updated_at, expected_updated_at)?;
    
    // Don't clone yet - only need it for error logging
    let list_id_ref = &entity.list_id;
    let txn = self.db.begin().await
        .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
    
    // ... transaction logic ...
    
    outcome => {
        if let Err(e) = txn.rollback().await {
            tracing::error!(
                error = %e,
                list_id = %list_id_ref,  // Borrow, don't clone
                "failed to roll back guarded status list update"
            );
        }
        outcome
    }
}
```

src/adapters/sea_orm/store.rs
Hermann-Core, line 217: Transaction begin failures are mapped to `UpdateError`, but a transaction failure is architecturally distinct from an update failure. This makes error diagnosis harder.

**Impact**: Operators seeing `UpdateError` in logs won't know if it's a constraint violation, network issue, or transaction system failure.

**Options**:

1. **Add a TransactionError variant** (best):
```rust
// In error.rs
#[derive(Debug, thiserror::Error)]
pub enum RepositoryError {
    // ...
    #[error("Transaction error: {0}")]
    TransactionError(String),
}

// Here:
let txn = self.db.begin().await
    .map_err(|e| RepositoryError::TransactionError(
        format!("Failed to begin transaction: {e}")
    ))?;
```

2. **At minimum, clarify in the error message**:
```rust
.map_err(|e| RepositoryError::UpdateError(
    format!("Transaction begin failed: {e}")
))?;
```

src/database/queries.rs
Hermann-Core: This function takes `entity: StatusListRecord` by value but only needs to read its fields. The record contains a `StatusList` with a compressed bitstring (`lst: String`) that can be up to `max_serialized_list_size` (1 MiB by default).

**Impact**: Forces caller to either:
1. Clone the entire record (expensive - up to 1 MB copy)
2. Move ownership (prevents reuse)

**Fix**: Take by reference:
```rust
async fn execute_guarded_update<C: ConnectionTrait>(
    conn: &C,
    entity: &StatusListRecord,
    expected_updated_at: i64,
    let list_id = &entity.list_id;
    let result = status_lists::Entity::update_many()
        .col_expr(status_lists::Column::Issuer, Expr::value(&entity.issuer))
        .col_expr(
            status_lists::Column::StatusList,
            Expr::value(&entity.status_list),
        )
        .col_expr(status_lists::Column::Sub, Expr::value(&entity.sub))
        .col_expr(
            status_lists::Column::UpdatedAt,
            Expr::value(entity.updated_at),
        )
        .filter(status_lists::Column::ListId.eq(list_id))
        .filter(status_lists::Column::UpdatedAt.eq(expected_updated_at))
        .exec(conn)
        .await
        .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
    Ok(result.rows_affected > 0)
}
```

Update call site at line 201:
```rust
if !execute_guarded_update(&txn, &entity, expected_updated_at).await? {
```

src/database/queries.rs
Hermann-Core: The acceptance criteria requires "Behavior verified on Postgres, MySQL, and SQLite" but I only see SQLite (line 935, 972, 1025) and MySQL (line 1081) tests. No PostgreSQL test.

**Why this matters**: While transaction semantics should be portable, PostgreSQL has different isolation defaults and lock escalation behavior than MySQL. Explicit verification prevents prod-only failures.

**Required**: Add PostgreSQL rollback test:
```rust
#[cfg(feature = "postgres")]
#[tokio::test]
async fn test_postgres_update_one_with_snapshot_rolls_back_when_history_insert_fails() {
    let test_db = postgres_helpers::postgres_connection().await;
    tokio::time::timeout(TXN_TEST_DEADLINE, async {
        // Same test pattern as MySQL version
        // ...
    })
    .await
    .expect("transactional update deadlocked on PostgreSQL");
}
```

If postgres test helpers don't exist yet, add them following the MySQL pattern at line 311.

src/database/queries.rs
Hermann-Core: The 10-second deadline may cause flaky CI failures on slow runners or containerized environments (GitHub Actions, GitLab CI, etc.) where DB containers can take seconds to respond under load.

**Impact**: Tests that work locally fail intermittently in CI, wasting developer time.

**Recommendation**: Make configurable via environment variable:
```rust
const TXN_TEST_DEADLINE: std::time::Duration = std::time::Duration::from_secs(
    std::option_env!("TXN_TEST_TIMEOUT_SECS")
        .and_then(|s| s.parse().ok())
        .unwrap_or(10)
);
```

Then in CI config:
```yaml
env:
  TXN_TEST_TIMEOUT_SECS: 30
```

This allows CI to use longer timeouts without slowing down local development.

src/web/handlers/status_list/update_status.rs
```diff
@@ -122,39 +122,61 @@ pub async fn update_status(
...
+            appstate.cache.invalidate(&list_id).await;
+            // Deliberately opaque to the client. `RepositoryError`'s `Display`
+            // embeds the raw `DbErr`, and `From<RepositoryError> for ApiError`
+            // copies it into `error_description` — which would put table,
+            // column and constraint names, and fragments of the offending
+            // values, into the response body. The detail belongs in the log.
+            tracing::error!(
+                error = %e,
+                %list_id,
+                "Guarded status list update failed"
```
Hermann-Core: The current error handling logs the full error but returns a generic 500. However, the comment mentions that `From<RepositoryError> for ApiError` copies `DbErr` details into `error_description`.

**Verify**: Check that the `ApiError` conversion is actually opaque. If constraint names, table names, or data fragments appear in the response body, this leaks internal schema details to attackers.

**If ApiError leaks details**: The fix is already correct (returning `InternalServerError` variant). Just confirm the implementation.

**If it doesn't leak**: This comment is overly cautious and the code is fine as-is.

src/web/handlers/status_list/update_status.rs
```diff
@@ -122,39 +122,61 @@ pub async fn update_status(
...
+    let updated = match store
+        .update_one_with_snapshot(exact_status_list, previous_updated_at, snapshot)
+        .await
+    {
+        Ok(updated) => updated,
+        Err(e) => {
+            // A failed `COMMIT` does not prove the write did not land — the
+            // server may have committed before the connection dropped. Treating
+            // it as "nothing happened" and skipping invalidation would leave the
+            // cache serving the old state indefinitely. Invalidation is
```
Hermann-Core: The comment correctly identifies that a failed `COMMIT` doesn't prove the write didn't land (network drop after server commit but before ack). The invalidation is correct defensive programming.

**However**: Consider adding a metric here to track commit failures, since this pattern (invalidate on uncertainty) could hide a systemic issue:

```rust
Err(e) => {
    metrics::counter!("status_list_update_commit_failures_total").increment(1);
    appstate.cache.invalidate(&list_id).await;
    tracing::error!(
        error = %e,
        %list_id,
        "Guarded status list update failed (commit uncertain)"
    );
    return Err(StatusListError::InternalServerError.into());
}
```

If this counter spikes, operators know there's a DB connection or transaction stability problem.

Hermann-Core (changes requested):
Hello @martcpp, I left a few comments below for your attention.

While you're addressing them, I'd also recommend reducing the verbosity of the comments. Try to document only the non-obvious or particularly relevant aspects of the implementation. At the moment, the files are quite large, and the amount of commentary, including comments describing obvious behavior makes it harder to read and follow.

martcpp:
@Hermann-Core @Ngha-Boris this ticket will be revamp after the https://github.com/adorsys/status-list-server/issues/186 and https://github.com/adorsys/status-list-server/issues/245  is merge thanks

src/adapters/sea_orm/store.rs
```diff
@@ -345,6 +486,51 @@ mod test {
...
+            let url = format!("postgres://postgres:postgres@{host}:{port}/postgres");
+            let mut opt = sea_orm::ConnectOptions::new(url);
+            opt.max_connections(5);
+            let db = sea_orm::Database::connect(opt)
+                .await
+                .expect("Failed to connect to Postgres");
+            crate::adapters::sea_orm::Migrator::up(&db, None)
+                .await
+                .expect("Failed to run migrations on Postgres");
+            PostgresTestDb {
```
Ngha-Boris, line 527: **Suggestion:** Consider adding `test_mysql_insert_with_snapshot_is_atomic()` to mirror the SQLite test at line 881. This would fully satisfy the acceptance criteria of verifying rollback behavior on "at least MySQL + SQLite" for the publish/insert path, not just the update path.

```rust
#[cfg(feature = "mysql")]
#[tokio::test]
async fn test_mysql_insert_with_snapshot_is_atomic() {
    // Mirrors test_sqlite_insert_with_snapshot_is_atomic for MySQL
}
```

Ngha-Boris (approved):
This PR successfully implements the ticket requirements. The transaction wrapping for `UPDATE status_lists` + `INSERT status_list_history` is correctly implemented with proper rollback behavior on both conflict (409) and failure (500) paths.

### Highlights
✅ Transaction semantics are portable across Postgres/MySQL/SQLite  
✅ Comprehensive test coverage including rollback verification  
✅ InnoDB pinning ensures MySQL transactions work correctly  
✅ Memory adapter mirrors SQL behavior with proper locking  
✅ Documentation is thorough and explains the trade-offs  

### Minor Suggestions (non-blocking)

Overall: **LGTM!** 🚀

Hermann-Core (approved):
@martcpp LGTM! Status list row updates and history snapshot writes are now fully atomic across all database backends (SQLite, MySQL, Postgres). Approved.

src/adapters/sea_orm/store.rs
martcpp: that is comming in the next pr thanks

PR #251, opened 2026-07-22
adding podcontainer security context

### LINKED ISSUE #239: Helm chart has no pod/container security context
## Problem

No `securityContext`, no probes, `resources` never rendered, no `NetworkPolicy`. AWS creds mounted at `/root/.aws` unreadable by UID 65534.

Specifically:
- `helm/chart/templates/deployment.yaml` renders **no pod or container `securityContext`**
- `.Values.statuslist.resources` is defined in `values.yaml` but **never referenced in deployment.yaml** — resource limits are silently dropped
- No liveness or readiness probes defined
- No `NetworkPolicy` template exists
- AWS credentials volume (`aws-credentials-secret`) is mounted at `/root/.aws`, which will become unreadable when we add `runAsUser: 65534`

## Concrete Regression Risk for #248

The `/root/.aws` mount path will **silently break cert provisioning and AWS Secrets Manager access** the moment this chart is deployed with `runAsNonRoot: true` / `runAsUser: 65534`. This isn't theoretical — it's a guaranteed breakage on first real Helm-based deploy unless this ticket pre-empts it.

#248 must verify (as part of its acceptance criteria) that AWS credentials are accessible after cutover.

## Fix

Render:
- **Container security context**:
  - `runAsNonRoot: true`
  - `runAsUser: 65534`
  - `readOnlyRootFilesystem: true`
  - `allowPrivilegeEscalation: false`
  - `capabilities.drop: [ALL]`
  - `seccompProfile.type: RuntimeDefault`
- **Pod security context**:
  - `runAsNonRoot: true`
  - `fsGroup: 65534` (for volume ownership)
- **Actually wire `resources`**: Reference `.Values.statuslist.resources` in the container spec
- **Add probes**:
  - Liveness probe (e.g., HTTP GET `/health/live`)
  - Readiness probe (e.g., HTTP GET `/health/ready`)
  - Startup probe if needed for slow initialization
- **Fix AWS creds mount path**: Change from `/root/.aws` to `/home/nonroot/.aws` (or equivalent), with matching `AWS_SHARED_CREDENTIALS_FILE` or `HOME` env var
- **Add `NetworkPolicy`**: Ingress/egress rules appropriate for the service
- **Add CI validation**: `trivy config` on the chart (catches all of the above); `kube-linter` as optional additional check

## Tool

- **`trivy config`** on the chart (catches all of the above)
- **`kube-linter`** if you want chart-specific validation

## Benefits for Intermittent Maintenance

1. **Security by default**: Future contributors deploy secure workloads without needing to remember security checklists
2. **Visible resource limits**: Prevents surprise OOMKills during dormant periods when dependencies drift
3. **Health visibility**: Kubernetes can automatically detect and restart unhealthy pods
4. **Network segmentation**: NetworkPolicy prevents lateral movement in case of compromise

## Acceptance Criteria

- [ ] Add container + pod `securityContext` with all fields listed above
- [ ] Render `resources` into the container spec (reference `.Values.statuslist.resources`)
- [ ] Add liveness/readiness probes (and startup probe if needed)
- [ ] **Fix AWS-creds mount path for UID 65534**: Change from `/root/.aws` to non-root-accessible path, set appropriate env var
- [ ] Add `NetworkPolicy` template with appropriate ingress/egress rules
- [ ] Add `trivy config` CI job that fails if security issues are detected
- [ ] **Verify in a test deploy** (not just PR approval): Deploy the chart to a test cluster, verify that:
  - Pod starts successfully with non-root UID
  - AWS credentials are accessible (test cert provisioning or Secrets Manager access)
  - Resource limits are actually applied (`kubectl describe pod` shows them)
  - Probes work (pod transitions to Ready)
  - NetworkPolicy is enforced

ndefokou:
The implementation render `runAsNonRoot`, `runAsUser: 65534`, `readOnlyRootFilesystem`,  `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `seccompProfile: RuntimeDefault`; wire `resources`; add probes; fix creds mount path; add `NetworkPolicy`.

- Have added `trivy config` on the chart (catches all of the above);  and `kube-linter` for
  chart-specific.

Was complited:
 
- [x] Add container + pod `securityContext`
- [x] Render `resources` into the container spec
- [x] Add liveness/readiness probes
- [x] Fix AWS-creds mount path for UID 65534
- [x] Add `NetworkPolicy`
- [x] Add `trivy config` CI job

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -225,7 +225,9 @@ jobs:
         uses: azure/setup-helm@v3
 
       - name: Build Helm dependencies
-        run: helm dependency build helm/chart
+        run: |
+          helm repo add dandydeveloper https://dandydeveloper.github.io/charts
+          helm dependency build helm/chart
 
       - name: Render Helm template
         run: |
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -221,11 +221,19 @@ jobs:
       - name: Fetch Repository
         uses: actions/checkout@v7
 
-      - name: Run Trivy config scan on Helm chart
+      - name: Setup Helm
+        uses: azure/setup-helm@v3
+
+      - name: Render Helm template
+        run: |
+          mkdir -p /tmp/rendered
+          helm template status-list-server helm/chart -f helm/chart/values.yaml --output-dir /tmp/rendered
+
+      - name: Run Trivy config scan
         uses: aquasecurity/trivy-action@v0.36.0
         with:
           scan-type: config
-          scan-ref: helm/chart
+          scan-ref: /tmp/rendered
           severity: HIGH,CRITICAL
           exit-code: 1
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -243,6 +243,34 @@ jobs:
           exit-code: 1
           trivyignores: .trivyignore.yaml
 
+  kube-linter:
+    name: KubeLinter Scan (Helm Chart)
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v7
+
+      - name: Setup Helm
+        uses: azure/setup-helm@v3
+
+      - name: Build Helm dependencies
+        run: |
+          helm repo add dandydeveloper https://dandydeveloper.github.io/charts
+          helm dependency build helm/chart
+
+      - name: Render Helm template
+        run: |
+          mkdir -p /tmp/rendered
+          helm template status-list-server helm/chart -f helm/chart/values.yaml --output-dir /tmp/rendered
+
+      - name: Run KubeLinter
+        uses: stackrox/kube-linter-action@v1
+        with:
+          directory: /tmp/rendered
+          format: sarif
+          output: kube-linter-results.sarif
+
   yaml-lint:
     name: YAML Lint and Format Check
     runs-on: ubuntu-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -269,6 +269,7 @@ jobs:
         with:
           directory: /tmp/rendered
           format: sarif
+          config: .kube-linter.yaml
           output-file: kube-linter-results.sarif
 
   yaml-lint:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -224,6 +224,9 @@ jobs:
       - name: Setup Helm
         uses: azure/setup-helm@v3
 
+      - name: Build Helm dependencies
+        run: helm dependency build helm/chart
+
       - name: Render Helm template
         run: |
           mkdir -p /tmp/rendered
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -269,7 +269,7 @@ jobs:
         with:
           directory: /tmp/rendered
           format: sarif
-          output: kube-linter-results.sarif
+          output-file: kube-linter-results.sarif
 
   yaml-lint:
     name: YAML Lint and Format Check
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -241,6 +241,7 @@ jobs:
           scan-ref: /tmp/rendered
           severity: HIGH,CRITICAL
           exit-code: 1
+          trivyignores: .trivyignore.yaml
 
   yaml-lint:
     name: YAML Lint and Format Check
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -222,7 +222,7 @@ jobs:
         uses: actions/checkout@v7
 
       - name: Run Trivy config scan on Helm chart
-        uses: aquasecurity/trivy-action@0.30.0
+        uses: aquasecurity/trivy-action@v0.36.0
         with:
           scan-type: config
           scan-ref: helm/chart
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -213,6 +213,22 @@ jobs:
         with:
           globs: "**/*.md"
 
+  trivy-config:
+    name: Trivy Config Scan (Helm Chart)
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v7
+
+      - name: Run Trivy config scan on Helm chart
+        uses: aquasecurity/trivy-action@0.30.0
+        with:
+          scan-type: config
+          scan-ref: helm/chart
+          severity: HIGH,CRITICAL
+          exit-code: 1
+
   yaml-lint:
     name: YAML Lint and Format Check
     runs-on: ubuntu-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,6 +31,7 @@ jobs:
     name: Cargo Build
     runs-on: ubuntu-latest
     needs: cargo-fmt
+    timeout-minutes: 45
 
     steps:
       - name: Fetch Repository
@@ -109,6 +110,7 @@ jobs:
     if: ${{ needs.check-crate-type.outputs.is_lib == 'true' }}
     name: Cargo test doc
     runs-on: ubuntu-latest
+    timeout-minutes: 45
 
     steps:
       - name: Fetch Repository
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -239,7 +239,7 @@ jobs:
         with:
           scan-type: config
           scan-ref: /tmp/rendered
-          severity: CRITICAL
+          severity: HIGH,CRITICAL
           exit-code: 1
 
   yaml-lint:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -239,7 +239,7 @@ jobs:
         with:
           scan-type: config
           scan-ref: /tmp/rendered
-          severity: HIGH,CRITICAL
+          severity: CRITICAL
           exit-code: 1
 
   yaml-lint:
--- a/.trivyignore.yaml
+++ b/.trivyignore.yaml
@@ -1,0 +1,8 @@
+misconfigurations:
+  - id: KSV-0014
+    paths:
+      - "*/charts/postgres*/**"
+      - "*/charts/redis-ha*/**"
+  - id: KSV-0118
+    paths:
+      - "*/charts/redis-ha*/**"
--- a/.trivyignore.yaml
+++ b/.trivyignore.yaml
@@ -1,8 +1,8 @@
 misconfigurations:
-  - id: KSV-0014
+  - id: AVD-KSV-0014
     paths:
       - "*/charts/postgres*/**"
       - "*/charts/redis-ha*/**"
-  - id: KSV-0118
+  - id: AVD-KSV-0118
     paths:
       - "*/charts/redis-ha*/**"
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -100,8 +100,10 @@ spec:
                   key: redis-password
             - name: APP_REDIS__URI
               value: {{ include "status-list-server-chart.redisUri" . | quote }}
-          {{- if .Values.secretStore.enabled }}
           volumeMounts:
+            - name: tmp
+              mountPath: /tmp
+          {{- if .Values.secretStore.enabled }}
             - name: aws-credentials-volume
               mountPath: /home/nobody/.aws
               readOnly: true
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -24,6 +24,9 @@ spec:
                 "readOnlyRootFilesystem" true
                 "allowPrivilegeEscalation" false
                 "capabilities" (dict "drop" (list "ALL")))
+              "resources" (dict
+                "requests" (dict "cpu" "50m" "memory" "64Mi")
+                "limits" (dict "cpu" "100m" "memory" "128Mi"))
               "volumeMounts" (list (dict "name" "tmp" "mountPath" "/tmp")) }}
         - {{- tpl (toYaml (mustMergeOverwrite $ic $hardened)) $ | nindent 10 }}
         {{- end }}
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -19,16 +19,13 @@ spec:
       {{- with .Values.statuslist.initContainers }}
       initContainers:
         {{- range $ic := . }}
-        - {{- tpl (toYaml $ic) $ | nindent 10 }}
-          securityContext:
-            readOnlyRootFilesystem: true
-            allowPrivilegeEscalation: false
-            capabilities:
-              drop:
-                - ALL
-          volumeMounts:
-            - name: tmp
-              mountPath: /tmp
+        {{- $hardened := dict
+              "securityContext" (dict
+                "readOnlyRootFilesystem" true
+                "allowPrivilegeEscalation" false
+                "capabilities" (dict "drop" (list "ALL")))
+              "volumeMounts" (list (dict "name" "tmp" "mountPath" "/tmp")) }}
+        - {{- tpl (toYaml (mustMergeOverwrite $ic $hardened)) $ | nindent 10 }}
         {{- end }}
       {{- end }}
       securityContext:
@@ -46,6 +43,8 @@ spec:
             capabilities:
               drop:
                 - ALL
+            seccompProfile:
+              type: RuntimeDefault
           image: {{ .Values.statuslist.image.repository }}:{{ .Values.statuslist.image.tag }}
           imagePullPolicy: {{ .Values.statuslist.image.pullPolicy }}
           ports:
@@ -88,6 +87,10 @@ spec:
             {{- if .Values.secretStore.enabled }}
             - name: APP_AWS__REGION
               value: {{ .Values.secretStore.aws.region | quote }}
+            - name: AWS_SHARED_CREDENTIALS_FILE
+              value: /home/nobody/.aws/credentials
+            - name: AWS_CONFIG_FILE
+              value: /home/nobody/.aws/config
             {{- end }}
             - name: POSTGRES_DB
               value: {{ .Values.postgres.auth.database | quote }}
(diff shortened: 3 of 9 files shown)


.github/workflows/CI.yml
```diff
@@ -213,6 +213,22 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v7
+
+      - name: Run Trivy config scan on Helm chart
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          scan-type: config
```
Ngha-Boris, line 242: ⚠️ **Issue**: Trivy's `config` scanner expects rendered Kubernetes manifests, not Helm templates. The `helm/chart` directory contains templated YAML files with Go template syntax (e.g., `{{ .Values... }}`), which Trivy cannot properly evaluate.

**Suggestion**: Render the Helm chart before scanning:
```yaml
      - name: Setup Helm
        uses: azure/setup-helm@v3
        
      - name: Render Helm template
        run: |
          mkdir -p /tmp/rendered
          helm template status-list-server helm/chart -f helm/chart/values.yaml --output-dir /tmp/rendered
          
      - name: Run Trivy config scan
        uses: aquasecurity/trivy-action@v0.36.0
        with:
          scan-type: config
          scan-ref: /tmp/rendered
          severity: HIGH,CRITICAL
          exit-code: 1
```

helm/chart/templates/deployment.yaml
```diff
@@ -67,7 +92,7 @@ spec:
           {{- if .Values.secretStore.enabled }}
           volumeMounts:
             - name: aws-credentials-volume
-              mountPath: /root/.aws
+              mountPath: /home/nobody/.aws
```
Ngha-Boris, line 114: ✅ **Good**: Changed from `/root/.aws` to `/home/nobody/.aws`. Since the container runs as UID 65534 (nobody), the AWS credentials will now be readable. The `fsGroup: 65534` setting ensures the mounted secret has the correct group ownership.

helm/chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,40 @@
...
+  ingress:
+    - ports:
+        - port: {{ .Values.statuslist.service.targetPort | int }}
+          protocol: TCP
+      {{- with .Values.statuslist.networkPolicy.ingress }}
+      from:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+  egress:
+    - ports:
```
Ngha-Boris, line 25: 💡 **Suggestion**: Consider allowing DNS TCP port 53 in addition to UDP. Some environments or configurations may use TCP for DNS queries:
```yaml
        - port: 53
          protocol: TCP
        - port: 53
          protocol: UDP
```

helm/chart/values.yaml
```diff
@@ -98,6 +98,26 @@ statuslist:
...
+    httpGet:
+      path: /health
+      port: http
+    initialDelaySeconds: 5
+    periodSeconds: 10
+    timeoutSeconds: 5
+    failureThreshold: 3
+  networkPolicy:
+    enabled: true
+    ingress: []
```
Ngha-Boris, line 126: ⚠️ **Issue**: The `networkPolicy.egress` is empty (`[]`) by default, which means all egress destinations are allowed. While this works, consider documenting in comments that users should restrict egress to specific CIDR blocks or namespaces for production deployments.

Example comment:
```yaml
  networkPolicy:
    enabled: true
    # Restrict ingress to specific namespaces or CIDR blocks
    ingress: []
    # Restrict egress to specific destinations (e.g., PostgreSQL, Redis, AWS API)
    egress: []
```

Ngha-Boris (commented):
@ndefokou please check the comments

Ngha-Boris (commented):
## Follow-up Review

All previous comments have been addressed. However, I found a few new issues that need to be fixed:

### Issues Found:

1. **Main container missing `/tmp` mount**: With `readOnlyRootFilesystem: true`, the main container needs a writable `/tmp` directory.

2. **Unconditional `tmp` volume**: The `tmp` volume is created even when there are no init containers.

3. **Trivy severity level**: Changed from `HIGH,CRITICAL` to just `CRITICAL` - please confirm if this is intentional.

See inline comments for details.

.github/workflows/CI.yml
```diff
@@ -213,6 +213,35 @@ jobs:
...
+        run: |
+          mkdir -p /tmp/rendered
+          helm template status-list-server helm/chart -f helm/chart/values.yaml --output-dir /tmp/rendered
+
+      - name: Run Trivy config scan
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          scan-type: config
+          scan-ref: /tmp/rendered
+          severity: CRITICAL
```
Ngha-Boris: **Question**: The severity was changed from `HIGH,CRITICAL` to just `CRITICAL`. Is this intentional? 

This means HIGH severity misconfigurations will no longer fail the build. If this was to reduce noise, consider documenting why HIGH severity is excluded.

**Suggestion** (if you want both):
```yaml
severity: HIGH,CRITICAL
```

helm/chart/templates/deployment.yaml
```diff
@@ -67,13 +103,15 @@ spec:
...
               readOnly: true
           {{- end }}
-      {{- if .Values.secretStore.enabled }}
       volumes:
```
Ngha-Boris, line 137: **Issue**: The `tmp` emptyDir volume is created unconditionally (outside the `initContainers` block), meaning it will exist even when there are no init containers defined.

**Suggestion**: Move the `tmp` volume inside the initContainers conditional:
```yaml
      {{- if .Values.statuslist.initContainers }}
      volumes:
        - name: tmp
          emptyDir: {}
      {{- end }}
      {{- if .Values.secretStore.enabled }}
        - name: aws-credentials-volume
          secret:
            secretName: aws-credentials-secret
      {{- end }}
```

helm/chart/templates/deployment.yaml
```diff
@@ -18,16 +18,52 @@ spec:
...
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        fsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
       containers:
         - name: {{ include "status-list-server-chart.name" . }}
+          securityContext:
```
Ngha-Boris, line 43: **Issue**: The main container has `readOnlyRootFilesystem: true` but no `/tmp` mount. Many applications (including Rust applications) need a writable `/tmp` directory for temporary files.

**Suggestion**: Add a volume mount for `/tmp`:
```yaml
          volumeMounts:
            - name: tmp
              mountPath: /tmp
```

And ensure the `tmp` volume is always created (not just for initContainers).

.trivyignore.yaml
```diff
@@ -0,0 +1,8 @@
+misconfigurations:
+  - id: KSV-0014
```
Christiantyemele: **Important — Trivy check ID format is likely wrong**

Trivy reports misconfiguration IDs in the format `AVD-KSV-0014` (or `KSV014` in older versions), not `KSV-0014` (with a dash and leading zero). If the ID does not match exactly what Trivy emits, the ignore rule silently has no effect and the CI scan will still fail on these checks — or, worse, if the check was already passing for other reasons, you will not know the ignore is broken until it matters.

Verify the actual IDs by running `trivy config --format json <rendered-dir>` locally and checking the `MisconfigSummary.ID` field, then update accordingly:

```yaml
misconfigurations:
  - id: AVD-KSV-0014   # was KSV-0014
    paths:
      - "*/charts/postgres*/**"
      - "*/charts/redis-ha*/**"
  - id: AVD-KSV-0118   # was KSV-0118
    paths:
      - "*/charts/redis-ha*/**"
```

helm/chart/templates/deployment.yaml
```diff
@@ -18,16 +18,52 @@ spec:
     spec:
       {{- with .Values.statuslist.initContainers }}
       initContainers:
-        {{- tpl (toYaml .) $ | nindent 8 }}
+        {{- range $ic := . }}
+        - {{- tpl (toYaml $ic) $ | nindent 10 }}
```
Christiantyemele: **Bug — init-container YAML merging produces duplicate keys for custom init containers**

The pattern:
```yaml
- {{- tpl (toYaml $ic) $ | nindent 10 }}
  securityContext:
    ...
  volumeMounts:
    - name: tmp
      mountPath: /tmp
```
appends `securityContext` and `volumeMounts` as raw YAML after whatever `$ic` renders. If a user-defined init container already includes either of those keys (e.g., to mount a config map), the output will have **duplicate YAML keys**. YAML parsers take the last value, so the user-supplied mounts would be silently dropped and the enforced securityContext would also be clobbered.

A safer approach is to use `mustMergeOverwrite` so the hardened defaults win without discarding user-provided keys:

```yaml
{{- range $ic := . }}
{{- $hardened := dict
      "securityContext" (dict
        "readOnlyRootFilesystem" true
        "allowPrivilegeEscalation" false
        "capabilities" (dict "drop" (list "ALL")))
      "volumeMounts" (list (dict "name" "tmp" "mountPath" "/tmp")) }}
- {{- tpl (toYaml (mustMergeOverwrite $ic $hardened)) $ | nindent 10 }}
{{- end }}
```

Note that `mustMergeOverwrite` merges at the top level (lists are replaced, not appended), so user-provided `volumeMounts` entries would need to be handled separately if they should also be preserved.

helm/chart/templates/deployment.yaml
```diff
@@ -64,16 +100,20 @@ spec:
...
             - name: APP_REDIS__URI
               value: {{ include "status-list-server-chart.redisUri" . | quote }}
-          {{- if .Values.secretStore.enabled }}
           volumeMounts:
+            - name: tmp
+              mountPath: /tmp
+          {{- if .Values.secretStore.enabled }}
             - name: aws-credentials-volume
-              mountPath: /root/.aws
+              mountPath: /home/nobody/.aws
```
Christiantyemele, line 114: **Important — `/home/nobody/.aws` may not exist in the application image**

UID 65534 (`nobody`) has a home directory of `/nonexistent` (Debian/Ubuntu) or `/var/empty` (Alpine). Most distroless/scratch images do not create `/home/nobody` at all. The AWS SDK (both the official Rust `aws-sdk-rust` crate and `rusoto`) resolves `~/.aws` from the `HOME` environment variable or `/etc/passwd` for the running UID — it does not hard-code `/home/nobody`.

If the application container does not define `HOME=/home/nobody` and does not have that directory, the credentials volume will be mounted at a path the SDK never reads, silently breaking AWS authentication.

Options:
1. Add `HOME=/home/nobody` to the container env and ensure the directory exists in the image.
2. Or set `AWS_SHARED_CREDENTIALS_FILE=/home/nobody/.aws/credentials` (and `AWS_CONFIG_FILE`) explicitly in the env, independent of the home directory.
3. Or use a well-known path such as `/etc/aws` and override the env var.

Please confirm this works end-to-end against the actual `ghcr.io/adorsys/status-list-server` image.

helm/chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,42 @@
...
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "status-list-server-chart.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    - ports:
```
Christiantyemele, line 17: **Suggestion — clarify "allow from any source" semantics in the default**

When `networkPolicy.ingress: []` (the default), the `{{- with ... }}` block is falsy (empty slice), so the `from:` key is omitted entirely. A Kubernetes ingress rule with `ports` but **no** `from` means "allow from **any** source". This is actually useful — it restricts ingress to the service port only while leaving source selection open — but the values.yaml comment says "Restrict ingress to specific namespaces or CIDR blocks", which implies restriction is happening when the list is empty.

Add a comment to prevent operators from thinking the default is a deny-all:

```yaml
networkPolicy:
  enabled: true
  # When ingress is empty [], ANY source can reach the service port (port-only restriction).
  # Add podSelector/namespaceSelector/ipBlock entries to restrict sources.
  ingress: []
  egress: []
```

helm/chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,42 @@
...
+        - port: {{ .Values.statuslist.service.targetPort | int }}
+          protocol: TCP
+      {{- with .Values.statuslist.networkPolicy.ingress }}
+      from:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+  egress:
```
Christiantyemele, line 24: **Suggestion — bundle egress ports into separate rules for finer-grained `to` control**

All egress ports (5432, 6379, 6380, 443, 53) share a single NetworkPolicy rule. When a user provides `networkPolicy.egress` destinations via the `to:` field, those destinations apply to **all** ports simultaneously. This makes it impossible to say "allow port 443 to anywhere (AWS API, ACME) but port 5432 only to the postgres pod".

Consider splitting into at least two rules:

```yaml
egress:
  # Internal: databases and cache (restrict with to: in values)
  - ports:
      - port: 5432
        protocol: TCP
      - port: 6379
        protocol: TCP
      - port: 6380
        protocol: TCP
    {{- with .Values.statuslist.networkPolicy.egressInternal }}
    to:
      {{- toYaml . | nindent 8 }}
    {{- end }}
  # External: HTTPS and DNS (unrestricted destination)
  - ports:
      - port: 443
        protocol: TCP
      - port: 53
        protocol: TCP
      - port: 53
        protocol: UDP
```

.github/workflows/CI.yml
```diff
@@ -213,6 +213,36 @@ jobs:
         with:
           globs: "**/*.md"
 
+  trivy-config:
```
Christiantyemele, line 218: **Question — kube-linter CI job mentioned in PR description but not added**

The PR description states: "Have added `trivy config` on the chart; and `kube-linter` for chart-specific." Only the Trivy job is present. Was the kube-linter job intentionally deferred, or is it missing? If deferred, please track it in a follow-up issue so it does not get lost.

helm/chart/templates/deployment.yaml
Christiantyemele, line 43: **Nit — consider adding `seccompProfile` at the container level for explicitness**

The `seccompProfile: RuntimeDefault` is set at the pod level (line 39–40) and containers inherit it. Some cluster policies and audit tools (e.g., kube-linter, policy admission controllers) check for the profile at the *container* level explicitly. Adding it to the container securityContext makes the intent unambiguous and avoids false positives:

```yaml
securityContext:
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
  seccompProfile:
    type: RuntimeDefault
```

Christiantyemele (changes requested):
## Review Summary

The PR successfully adds all the stated hardening goals — pod/container securityContext, readOnlyRootFilesystem, probes, resources, NetworkPolicy, and a Trivy CI job. Two prior concerns from @Ngha-Boris are now resolved: the main container `/tmp` mount is present (lines 103–105), and the Trivy severity is correctly `HIGH,CRITICAL`. The unconditional `tmp` volume is intentional and correct since the main container always needs it.

**Findings: 1 Bug, 2 Important, 2 Suggestion, 1 Question, 1 Nit.**

Key themes: (1) a latent duplicate-key risk in the init-container merging pattern, (2) the `.trivyignore.yaml` check IDs use a format that Trivy likely does not recognise, which would silently disable the suppression rules, and (3) the AWS-credentials mount path needs validating against the actual container image.

helm/chart/templates/deployment.yaml
Hermann-Core: Using `mustMergeOverwrite $ic $hardened` overwrites top-level slice keys like `volumeMounts`. If a user defines custom `volumeMounts` on an init container, `mustMergeOverwrite` completely replaces them with `[ { name: tmp, mountPath: /tmp } ]`. To preserve user mounts, append or concatenate `$ic.volumeMounts` with the hardened `/tmp` mount explicitly instead of replacing the entire list via `mustMergeOverwrite`.

.github/workflows/CI.yml
Hermann-Core: The `kube-linter` job was added to `.github/workflows/CI.yml`, but `stackrox/kube-linter-action@v1` scans `directory: /tmp/rendered` without `config: .kube-linter.yaml`. Because `/tmp/rendered` does not contain `.kube-linter.yaml`, KubeLinter will ignore `.kube-linter.yaml` in the repo root. Please pass `config: .kube-linter.yaml` explicitly to the action.

helm/chart/templates/deployment.yaml
```diff
@@ -64,16 +106,20 @@ spec:
...
               readOnly: true
```
Hermann-Core, line 115: Critical regression: In refactoring `deployment.yaml`, the `otel-collector` sidecar container definition and its associated `otel-collector-config` ConfigMap volume were completely removed from `spec.containers` and `spec.volumes`. If `otelCollector.enabled` and `otelCollector.sidecar.enabled` are true in values, the OpenTelemetry sidecar will no longer render. Please restore the sidecar container and volume blocks.

helm/chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,43 @@
...
+        - port: 5432
+          protocol: TCP
+        - port: 6379
+          protocol: TCP
+        - port: 6380
+          protocol: TCP
+      {{- with .Values.statuslist.networkPolicy.egressInternal }}
+      to:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
Hermann-Core, line 35: Critical networking bug: `values.yaml` defaults to `APP_TELEMETRY__ENABLED: "true"` and `APP_TELEMETRY__OTLP_ENDPOINT: "http://...:4317"`, but `network-policy.yaml` only permits egress to ports 5432, 6379/6380, 443, and 53. Port 4317 (OTLP gRPC) and port 4318 (OTLP HTTP) are missing from `egress`, which blocks OpenTelemetry export when `networkPolicy.enabled: true`. Please add allowed egress rules for ports 4317 and 4318.

Hermann-Core (changes requested):
@ndefokou Please review the inline comments and threaded replies for required updates. Action needed: Changes Requested.

Hermann-Core (approved):
@ndefokou LGTM! All review comments and security hardening requirements have been addressed. Approved.

Christiantyemele (approved):

Ngha-Boris (approved):

Hermann-Core (approved):

Ngha-Boris (approved):

Issue #168, opened 2026-07-06
Adopt compile-time backend feature flags and local/in-memory alternatives for external services


Hermann-Core:
## Objective

Every external service integration must expose a local/in-memory alternative alongside the production backend, selected at compile time via Cargo features rather than runtime configuration.

## Context

The server is currently hard-wired to production dependencies: PostgreSQL, AWS S3, AWS Secrets Manager, AWS Route53, and Redis. To keep CI fast, local development simple, and deployments portable, each integration should be wrapped behind a trait and gated behind Cargo features so that a fully in-memory build is possible.

This is a horizontal concern: each individual backend implementation is tracked in its own sub-issue, but this ticket establishes the pattern and fills the gaps.

## Deliverables

- [x] Agree on a crate-wide feature-naming convention (e.g., `postgres`, `sqlite`, `memory`, `aws-s3`, `redis`, `dns-route53`, `dns-cloudflare`).
- [x] Define the backend trait boundaries for:
  - status-list cache (#163)
  - certificate provisioning / store (#159)
  - certificate chain cache (#158)
  - DNS provider (#162)
  - signing-material / secrets store
- [x] Provide `memory` (or `sqlite` where appropriate) implementations for the backends that currently lack a local alternative.
- [x] Make `cargo build --no-default-features --features memory` produce a runnable minimal binary for local tests.
- [x] Ensure CI runs the test suite against the local backends.
- [x] Document the feature matrix in `README.md` and update `docker-compose.yml`.

## Acceptance Criteria

- All production backends are selectable through Cargo features.
- No production credentials are required for default local development.
- Compile errors clearly indicate when a required backend feature is missing.

## References

- Related: #143 (database backend), #158 (cert chain cache), #159 (cert provisioning), #162 (DNS providers), #163 (status-list cache)

est: 1day

martcpp:
Surfaced while building the release image for #388. There is a hole in this issue's feature matrix, and one of this issue's own acceptance criteria is what it violates.

## The combination that does not compile

`Dockerfile` builds with `ARG FEATURES="postgres,aws,acme"` — that is `aws` **without** `redis`. That combination fails:

```
error[E0282]: type annotations needed for `std::option::Option<_>`
  --> src/setup.rs:297:25
297 |                     let cache_opt = None;
    |                         ^^^^^^^^^   ---- type must be known at this point
```

[...]

PR #253, opened 2026-07-22
post-hexagonal closeout: classify corrupt stored lists 

### LINKED ISSUE #245: Post-hexagonal closeout: corrupt-state error classification, publish limit placement, dead error impl
# Post-hexagonal closeout: corrupt-state error classification, publish limit placement, dead error impl

Follow-up to the hexagonal refactor PR #212 (ticket #186). Three items, ordered by risk. None blocked the merge; items 1a/1b are the two paths that can misreport on corrupt stored data, which is why they lead.

## 1. Corrupt stored state must not surface as a client error (and must not decode silently)

**1a — Split `DomainError` into request-caused vs. state-caused; map corruption to 500.**

`DomainError::InvalidStatusList` currently covers both request problems (`ApplicationSpecific < 256`) and stored-state problems (`"Invalid lst encoding"`, `"Failed to decompress status list"`, `"Invalid status value in existing list"` during bit-width expansion). `map_domain_error` (`src/web/handlers/status_list/mod.rs`) sends all of them to `Generic` → **400 Bad Request**.

Consequence: a client sending a valid update to a corrupted list is told its request is bad, and data corruption is masked as client error — nobody gets paged. On a revocation service that's a masked alert, not just a wrong status code.

Fix: add a state-caused variant (e.g. `CorruptStoredList(String)`) used by `decode_compressed` / `decode_status_array` when operating on an existing list; map it to `InternalServerError` (500) and log at error level. Request-caused validation keeps 400.

**1b — `decode_status_array` must error on a truncated trailing entry (same files, same PR).**

On a short trailing byte the decode loop `break`s mid-value and pushes the partial read as a real status (`src/domain/mod.rs`). Note the honest scoping: this is **unreachable via data this server wrote** — `calculate_array_size` sizes arrays to whole entries — so it triggers only on externally corrupted rows. It is hardening, not a live bug. But the failure shape is silent-wrong-status on a status list, so the guard is worth its ~5 lines: error when `bits_read < bits` at loop end, plus one test with a deliberately truncated array.

## 2. Move the publish-path size limit into `PublishStatusList`; delete the handler preview

`max_serialized_list_size` is enforced inside the **update** use cases but on the **publish** path only via a handler preview (`src/web/handlers/status_list/publish_status.rs` — `StatusList::create(statuses.clone())` + `ensure_serialized_list_size`, after which the service recomputes the list from scratch).

Two problems, one fix: the next inbound adapter (S2S #155) can forget the publish limit because it lives in the forgettable layer; and every publish bit-packs/deflates/base64-encodes twice plus clones up to 5 000 entries. Give `PublishStatusList` the same `with_max_serialized_list_size` builder the update case has (surfacing `UseCaseError::StatusListTooLarge` → existing 422 mapping), then delete the preview block. ~15 lines.

## 3. Delete the dead `From<RepositoryError> for ApiError` impl

`src/web/errors.rs` still carries `impl From<RepositoryError> for ApiError` with **zero call sites** — `RepositoryError` no longer crosses the port boundary. Its catch-all arm puts `err.to_string()` (raw driver detail) into the response body, so if a future caller ever finds it convenient, it leaks. Delete the impl and its import: a deleted leak cannot be reintroduced; a mapped one relies on every future caller remembering to map.

## Explicitly out of scope here (tracked or deferred separately)

- Hot-path allocations (per-read record clone in `GetStatusListToken`; per-request JWK parse → `DecodingKey` in auth) — performance follow-up, fix when profiles say so.
- `WithHistory` struct-pair collapse — needs the `StatusListHistoryRepository` trait cfg-ungating first; separate ticket.
- Redundant `Domain(...)` match arms, naming debts (`list_uris` returns `sub` values, `Issuer` byte-equality) — cosmetic/documentation.
- `bits ≥ 9` for `ApplicationSpecific` vs. the spec's 1/2/4/8 table — deliberate, pinned, spec-evolution discussion.
- `memory-only = []` marker feature — deferred to #168.

## Acceptance criteria

- [ ] Valid update against a list with corrupt `lst` returns 500 and logs at error level; invalid request input still returns 400
- [ ] Truncated status array fails decode with an error (test with a hand-built short array)
- [ ] Publish size limit enforced in the use case; handler preview removed; existing 422 handler tests still pass
- [ ] `From<RepositoryError> for ApiError` gone; `cargo clippy --workspace --all-targets` clean
- [ ] No behavior change for well-formed data (spec-vector pins untouched)

martcpp:
# Post-hexagonal closeout: corrupt-state error classification, publish limit placement, dead error impl

Follow-up to the hexagonal refactor PR #212 (ticket #186). Three items, ordered by risk. None blocked the merge; items 1a/1b are the two paths that can misreport on corrupt stored data, which is why they lead.

## 1. Corrupt stored state must not surface as a client error (and must not decode silently)

**1a — Split `DomainError` into request-caused vs. state-caused; map corruption to 500.**

`DomainError::InvalidStatusList` currently covers both request problems (`ApplicationSpecific < 256`) and stored-state problems (`"Invalid lst encoding"`, `"Failed to decompress status list"`, `"Invalid status value in existing list"` during bit-width expansion). `map_domain_error` (`src/web/handlers/status_list/mod.rs`) sends all of them to `Generic` → **400 Bad Request**.

Consequence: a client sending a valid update to a corrupted list is told its request is bad, and data corruption is masked as client error — nobody gets paged. On a revocation service that's a masked alert, not just a wrong status code.

Fix: add a state-caused variant (e.g. `CorruptStoredList(String)`) used by `decode_compressed` / `decode_status_array` when operating on an existing list; map it to `InternalServerError` (500) and log at error level. Request-caused validation keeps 400.

**1b — `decode_status_array` must error on a truncated trailing entry (same files, same PR).**

On a short trailing byte the decode loop `break`s mid-value and pushes the partial read as a real status (`src/domain/mod.rs`). Note the honest scoping: this is **unreachable via data this server wrote** — `calculate_array_size` sizes arrays to whole entries — so it triggers only on externally corrupted rows. It is hardening, not a live bug. But the failure shape is silent-wrong-status on a status list, so the guard is worth its ~5 lines: error when `bits_read < bits` at loop end, plus one test with a deliberately truncated array.

## 2. Move the publish-path size limit into `PublishStatusList`; delete the handler preview

`max_serialized_list_size` is enforced inside the **update** use cases but on the **publish** path only via a handler preview (`src/web/handlers/status_list/publish_status.rs` — `StatusList::create(statuses.clone())` + `ensure_serialized_list_size`, after which the service recomputes the list from scratch).

Two problems, one fix: the next inbound adapter (S2S #155) can forget the publish limit because it lives in the forgettable layer; and every publish bit-packs/deflates/base64-encodes twice plus clones up to 5 000 entries. Give `PublishStatusList` the same `with_max_serialized_list_size` builder the update case has (surfacing `UseCaseError::StatusListTooLarge` → existing 422 mapping), then delete the preview block. ~15 lines.

## 3. Delete the dead `From<RepositoryError> for ApiError` impl

`src/web/errors.rs` still carries `impl From<RepositoryError> for ApiError` with **zero call sites** — `RepositoryError` no longer crosses the port boundary. Its catch-all arm puts `err.to_string()` (raw driver detail) into the response body, so if a future caller ever finds it convenient, it leaks. Delete the impl and its import: a deleted leak cannot be reintroduced; a mapped one relies on every future caller remembering to map.

## Explicitly out of scope here (tracked or deferred separately)

- Hot-path allocations (per-read record clone in `GetStatusListToken`; per-request JWK parse → `DecodingKey` in auth) — performance follow-up, fix when profiles say so.
- `WithHistory` struct-pair collapse — needs the `StatusListHistoryRepository` trait cfg-ungating first; separate ticket.
- Redundant `Domain(...)` match arms, naming debts (`list_uris` returns `sub` values, `Issuer` byte-equality) — cosmetic/documentation.
- `bits ≥ 9` for `ApplicationSpecific` vs. the spec's 1/2/4/8 table — deliberate, pinned, spec-evolution discussion.
- `memory-only = []` marker feature — deferred to #168.

## Acceptance criteria

- [ ] Valid update against a list with corrupt `lst` returns 500 and logs at error level; invalid request input still returns 400
- [ ] Truncated status array fails decode with an error (test with a hand-built short array)
- [ ] Publish size limit enforced in the use case; handler preview removed; existing 422 handler tests still pass
- [ ] `From<RepositoryError> for ApiError` gone; `cargo clippy --workspace --all-targets` clean
- [ ] No behavior change for well-formed data (spec-vector pins untouched)


Christiantyemele (approved):
## Summary

**Approve.** All three acceptance criteria are met with correct implementations: (1) corrupt-stored-state error classification with `CorruptStoredList` → 500 and correct log at error level, (2) truncated array well-formedness guard, and (3) publish size limit moved to `PublishStatusList` with handler preview removed, plus dead `From<RepositoryError> for ApiError` deletion. No behavioral changes for well-formed data — 263/263 tests pass, clippy clean, cargo check clean. Port-boundary conversions (`from_persistence` / `to_persistence`) properly handle the `Issuer` wrapper between sea_orm and domain models throughout.

ndefokou (approved):
It addresses the reviews i left on the main ticket LGFM 🚀

PR #258, opened 2026-07-23
Feat/add opentelemetry collector

### LINKED ISSUE #160: Add OpenTelemetry collector and make tracing configurable (dev/prod)

Hermann-Core:
Added an OpenTelemetry collector to the deployment stack and make the tracing, metrics, and logs collector configurable depending on the environment (dev or prod). In dev, they should output to stdout/console; in prod, it should export to an OpenTelemetry collector via OTLP.

## Acceptance Criteria

- [x] Telemetry output is configurable based on `ENVIRONMENT` variable
- [x] Dev mode logs to stdout (no collector required for local development)
- [x] Prod mode exports via OTLP to the collector
- [x] `docker-compose up` brings up the Otel collector alongside the server
- [x] Key operations are instrumented with spans (HTTP requests, DB queries, cert provisioning, list generation)
- [x] Helm chart's `otelCollector.enabled` and `statuslist.env.APP_TELEMETRY__ENABLED` are consistent (no silent telemetry failures)
- [x] Documentation updated with observability setup guide
- [ ] **Verify in a test deploy**: Deploy to a test environment, confirm:
  - Traces appear in Jaeger/Tempo
  - Metrics are scraped by Prometheus
  - Logs flow to Loki (if configured)
  - Dev mode works without a collector


Awambeng (commented):
Thanks for the PR, @Hermann-Core!
Could you please fix the failing CI checks before I review it?

Hermann-Core:
Hello @Awambeng can you check?

src/utils/telemetry.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/// Initializes the global telemetry stack (tracing + metrics) based on the
+/// provided configuration.
+///
+/// Returns a [`TelemetryGuard`] that must be held until application
+/// shutdown. Also returns a [`prometheus::Registry`] that the `/metrics`
+/// endpoint uses to render Prometheus metrics.
+pub fn init_telemetry(config: &TelemetryConfig) -> color_eyre::Result<(TelemetryGuard, Registry)> {
+    let is_prod = config.environment.eq_ignore_ascii_case(ENV_PRODUCTION);
```
forkimenjeckayang: > Every value other than the exact string `production` silently selects development/stdout mode. The issue describes `dev` and `prod`, so `APP_TELEMETRY__ENVIRONMENT=prod` currently disables export without an error; the same happens for typos such as `prodution`. `APP_ENV=production` also does not drive this separate setting.

 Please deserialize a typed environment enum, support the documented aliases if intended, and reject unknown values at startup. Add configuration tests for development/dev, production/prod, omitted values, and invalid input. The sampler ratio should likewise be validated as a finite value in the documented `0.0..=1.0` range.

src/utils/cache/cert_chain.rs
```diff
@@ -39,28 +46,24 @@ impl CertChainCache {
...
-            metrics::Unit::Count,
-            "Certificate chain cache misses"
-        );
-        metrics::describe_counter!(
-            REPLACEMENT_METRIC,
-            metrics::Unit::Count,
-            "Certificate chain cache replacements (post-provisioning)"
-        );
+        let meter = global::meter("status-list-server");
+
```
forkimenjeckayang, line 55: OpenTelemetry meters do not become live after the global provider changes. The Rust API explicitly requires `set_meter_provider()` before calling `global::meter()`; meters obtained earlier remain backed by the provider active at creation time. Production builds `CertChainCache` and `CertManagerMetrics` inside `build_state_with_cert_manager()` before `HttpServer::new()` calls `setup_metrics()`, so these stored counters/gauges remain no-op. Calling `add(0)` later cannot rebind them.
>
> The previous implementation deliberately resolved counter handles lazily to support this startup order. Please install and retain the meter provider as part of `init_telemetry()` before building application state, or construct/rebind the instruments after provider installation. Restore tests that assert actual values in the Prometheus registry; the new tests only prove that no-op instruments do not panic. Also verify the removal of `metrics-process`, because the previous process CPU/memory metrics are no longer collected.

Reference: [OpenTelemetry Rust `set_meter_provider()`](https://docs.rs/opentelemetry/0.32.0/opentelemetry/global/fn.set_meter_provider.html) warns that calling `meter()` first yields no-op meters, and [`global::meter()`](https://docs.rs/opentelemetry/0.32.0/opentelemetry/global/fn.meter.html) states that existing meters do not reflect later provider changes.

src/utils/telemetry.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    // Build the tracing subscriber
+    let env_filter = build_env_filter();
+
+    if is_prod && config.enabled {
+        // Production: JSON stdout + OTLP trace export
+        let tracer_provider = build_otlp_tracer_provider(config, resource)?;
+
+        let tracer = tracer_provider.tracer("status-list-server");
+        let otel_layer = tracing_opentelemetry::layer().with_tracer(tracer);
```
forkimenjeckayang: > This production branch creates only an OTLP `SpanExporter`. Metrics are still exposed by an in-process Prometheus reader and scraped directly from the app, while logs are written to stdout. The Collector configurations have no metrics pipeline; their logs pipeline has only the OTLP receiver and debug exporter, but the application sends no OTLP logs. Compose/Helm also deploy no Loki or equivalent log backend.

 This does not satisfy the issue's production requirement for tracing, metrics, and logs through the Collector. Any reason for this instead?

helm/chart/values.yaml
```diff
@@ -98,6 +98,11 @@ statuslist:
     APP_LIMITS__MAX_STATUS_INDEX: "100000"
     APP_LIMITS__MAX_STATUSES_PER_REQUEST: "5000"
     APP_LIMITS__MAX_SERIALIZED_LIST_SIZE: "1048576"
+    # Telemetry / OpenTelemetry
+    APP_TELEMETRY__ENVIRONMENT: "production"
+    APP_TELEMETRY__OTLP_ENDPOINT: "http://localhost:4317"
+    APP_TELEMETRY__ENABLED: "true"
+    APP_TELEMETRY__SAMPLER_RATIO: "1.0"
```
forkimenjeckayang, line 103: > The default values set `sidecar.enabled: false` and `standalone.enabled: true`, but the application exports to `localhost:4317`. The rendered application pod contains no Collector sidecar; the Collector is a separate Deployment exposed by `<release>-status-list-server-otel-collector`. Therefore `localhost` points back to the application container and all production OTLP exports fail.

Please derive the endpoint from the selected deployment mode: use `localhost:4317` only for the sidecar, and the rendered Collector Service DNS name for standalone mode. Also make `APP_TELEMETRY__ENABLED` follow `otelCollector.enabled`, or fail chart validation for inconsistent combinations.

Cargo.toml
```diff
@@ -82,16 +111,25 @@ async-trait = "0.1"
...
+reqwest = { version = "0.13", features = ["form", "json", "query"], optional = true }
 tower = { version = "0.5", optional = true }
 hyper-rustls = { version = "0.27", features = ["http2"], optional = true }
 axum = { version = "0.8", features = ["macros"], optional = true }
-tower-http = { version = "0.7", features = ["catch-panic", "cors", "limit", "trace"], optional = true }
+tower-http = {
+  version = "0.7",
+  features = ["catch-panic", "cors", "limit", "trace"],
+  optional = true
+}
```
forkimenjeckayang: > Cargo cannot parse this multi-line inline table. On the current branch, `cargo fmt --all --check`, `cargo test`, and `cargo metadata` all stop here with `invalid inline table: expected '}'`. The same form is also used for `sea-orm` and `sea-orm-migration`, so fixing only this first occurrence will expose the next one.
>
Please restore valid one-line inline dependency declarations or convert them to standard `[dependencies.<name>]` tables, then rerun the repository's full Cargo CI matrix. This must be fixed before any Rust behavior in the PR can be verified.

Image

docker-compose.yml
```diff
@@ -111,6 +115,34 @@ services:
...
+      - 4318:4318
+    depends_on:
+      jaeger:
+        condition: service_started
+    networks:
+      - status-list-network
+
+  jaeger:
+    image: jaegertracing/jaeger:2.20
+    container_name: jaeger
```
forkimenjeckayang, line 147: > `docker compose up otel-collector` cannot start this stack because Docker Hub has no `jaegertracing/jaeger:2.20` manifest. The published semantic-version tag is `2.20.0`, so Compose fails while pulling the dependency before it can start the Collector.
>
 Please pin an existing immutable Jaeger tag, verify it for the architectures supported by this project, and add `docker compose pull` or an equivalent manifest check to CI. After correcting this tag, the separate Collector configuration error below must still be fixed.

Reference: [official `jaegertracing/jaeger` image tags](https://hub.docker.com/r/jaegertracing/jaeger/tags) list `2.20.0`, not `2.20`.


Image

otel-collector.yaml
```diff
@@ -0,0 +1,33 @@
+file_format: "0.3"
+
```
forkimenjeckayang: > The exact Collector image pinned by this PR rejects this key. Running `otel/opentelemetry-collector-contrib:0.156.0 validate --config /etc/otelcol/config.yaml` fails with `'otelcol.configSettings' has invalid keys: file_format`, so `docker compose up` will start a Collector container that exits before accepting any traces.
>
Please remove or replace this unsupported key and add a CI validation step using the same pinned image against both the Compose and Helm Collector configurations. The two configurations are currently duplicated and have already drifted, so validating both would prevent this from recurring.

Image

forkimenjeckayang (changes requested):
Left a few comments @Hermann-Core 
Please check when chanced

docker-compose.yml
Awambeng: +1

helm/chart/templates/otel-collector-deployment.yaml
Awambeng, line 1: Instead of maintaining custom charts, can we consider switching to the official `opentelemetry-collector` chart?

src/utils/telemetry.rs
Hermann-Core: Fixed in fdeea1d. Telemetry environment is now a typed enum accepting development/dev and production/prod, unknown values fail config loading, APP_ENV=production drives the default when telemetry env is omitted, and sampler_ratio is validated as finite 0.0..=1.0. Added tests for aliases, omitted/default behavior, override behavior, invalid env, and sampler bounds.

src/utils/cache/cert_chain.rs
Hermann-Core: Fixed in fdeea1d. The meter provider is installed during init_telemetry before application state and certificate caches are built, and the guard retains/shuts it down. Metrics setup is no longer delayed until attach_metrics. I also restored a Prometheus registry assertion for cert-chain hit/miss/replacement counters and verified there are no remaining metrics-process references.

src/utils/telemetry.rs
Hermann-Core: Fixed in fdeea1d. Production telemetry now exports traces via OTLP span exporter, metrics via an OTLP periodic reader while retaining the Prometheus scrape registry, and logs via opentelemetry-appender-tracing. Both Collector configs now include a metrics pipeline, and the app sends OTLP logs in production mode.

helm/chart/values.yaml
Hermann-Core: Fixed in fdeea1d. The chart now derives APP_TELEMETRY__OTLP_ENDPOINT from the selected Collector mode: localhost only for sidecar, rendered Service DNS for standalone. APP_TELEMETRY__ENABLED follows otelCollector.enabled, and invalid or mutually exclusive Collector mode combinations fail templating.

Cargo.toml
Hermann-Core: This is fixed on the current branch. Cargo.toml now parses successfully; cargo metadata --no-deps --format-version 1, cargo fmt --all -- --check, cargo check --features server, clippy, and the full cargo test --features server suite pass locally.

docker-compose.yml
Hermann-Core: Fixed in fdeea1d. Jaeger is pinned to jaegertracing/jaeger:2.20.0 and CI now manifest-checks that tag. I also verified the manifest locally and docker compose config renders successfully.

otel-collector.yaml
Hermann-Core: Fixed in fdeea1d. Removed the unsupported file_format key, validated otel/opentelemetry-collector-contrib:0.156.0 against the Compose config, and added CI coverage for docker compose config, the Jaeger manifest, the Compose Collector config, and the rendered Helm Collector config.

Hermann-Core: Fixed in fdeea1d. Removed the unsupported file_format key, validated otel/opentelemetry-collector-contrib:0.156.0 against the Compose config, and added CI coverage for docker compose config, the Jaeger manifest, the Compose Collector config, and the rendered Helm Collector config.

helm/chart/templates/otel-collector-deployment.yaml
Hermann-Core: Good idea, but I would keep the switch to the official `opentelemetry-collector` chart as a follow-up rather than expand this PR. This PR already owns review fixes for the current custom chart wiring; switching chart ownership changes values, dependencies, and upgrade semantics.

Hermann-Core:
Hello @forkimenjeckayang @Awambeng thanks for the reviews, can you check again?

src/utils/metrics.rs
```diff
@@ -1,29 +1,55 @@
...
+        .with_resource(resource)
+        .with_reader(prometheus_exporter);
+
+    if config.enabled && config.environment.is_production() {
+        let otlp_exporter = MetricExporter::builder()
+            .with_tonic()
+            .with_endpoint(&config.otlp_endpoint)
+            .build()
+            .wrap_err("Failed to build OTLP metric exporter")?;
+        builder = builder.with_reader(PeriodicReader::builder(otlp_exporter).build());
```
forkimenjeckayang, line 38: > The provider-ordering problem is fixed, but this migration removes `metrics-process` and `start_metrics_collector()` without registering equivalent process instrumentation with the new meter provider. Consequently, existing metrics including `process_cpu_seconds_total`, resident/virtual memory, open file descriptors, thread count, and process start time disappear from `/metrics` and are not exported through OTLP.
>
Could we restore equivalent OpenTelemetry process/runtime metrics, or explicitly document and test an intentional platform-provided replacement. A regression assertion for the expected process metric families would prevent this observability loss from recurring.

forkimenjeckayang (commented):
One last comment @Hermann-Core

helm/chart/templates/otel-collector-deployment.yaml
Hermann-Core: Keeping this PR on the existing custom chart wiring and treating the official opentelemetry-collector chart migration as follow-up scope. The current PR now validates the custom chart-rendered Collector config in CI so we still have coverage while that larger chart ownership decision is handled separately.

src/utils/metrics.rs
Hermann-Core: I restored process metrics through OpenTelemetry observable instruments backed by metrics-process collector snapshots, so the families are again visible at /metrics and are also collected by the OTLP metric reader in production.

forkimenjeckayang (approved):
Thanks for considering my comments @Hermann-Core

.github/workflows/CI.yml
```diff
@@ -309,6 +309,73 @@ jobs:
       - name: Run yamlfmt
         run: yamlfmt --lint .
 
+  otel-config-validation:
+    name: OpenTelemetry Config Validation
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v7
```
Awambeng: Could you please pin the action version to a commit hash throughout this file?

.github/workflows/CI.yml
```diff
@@ -309,6 +309,73 @@ jobs:
...
+
+      - name: Validate Compose Collector config
+        run: |
+          docker run --rm \
+            -v "$PWD/otel-collector.yaml:/etc/otelcol/config.yaml:ro" \
+            otel/opentelemetry-collector-contrib:0.156.0 \
+            validate --config /etc/otelcol/config.yaml
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
```
Awambeng: please use the latest version of this action and pin it to a commit hash.

src/utils/cert_manager.rs
```diff
@@ -38,49 +38,70 @@ use crate::{
...
+        .add(0, &[]);
+    meter
+        .u64_counter(RENEWAL_SUCCESSES_METRIC)
+        .build()
+        .add(0, &[]);
+    meter
+        .u64_counter(RENEWAL_FAILURES_METRIC)
+        .build()
+        .add(0, &[]);
 }
```
Awambeng: This function looks like dead code now. Since `CertManager::init_renewal_counters(&self)` is the active implementation, can we remove this one and update the test accordingly?

Cargo.toml
```diff
@@ -124,7 +162,17 @@ dotenvy = { version = "0.15", optional = true }
...
+
+# OpenTelemetry
+opentelemetry = "0.32"
+opentelemetry_sdk = { version = "0.32", features = ["rt-tokio"] }
+opentelemetry-otlp = { version = "0.32", features = ["grpc-tonic"] }
+opentelemetry-appender-tracing = "0.32"
+tracing-opentelemetry = "0.33"
+opentelemetry-prometheus = "0.32"
+prometheus = "0.14"
+metrics-process = { version = "2.4", default-features = false }
```
Awambeng, line 108: These telemetry dependencies are currently unconditional, meaning minimal builds (`--no-default-features --features memory-only`) still pull in the full OTel/tonic stack. Since other integrations are feature-gated, should these also be behind the relevant server/http feature?

src/utils/metrics.rs
```diff
@@ -1,29 +1,198 @@
...
+        .u64_observable_gauge("process_threads")
+        .with_description("Number of OS threads in the process.")
+        .with_unit("{thread}")
+        .with_callback(|observer| {
+            if let Some(value) = metrics_process::collector::collect().threads {
+                observer.observe(value, &[]);
+            }
+        })
+        .build();
 }
```
Awambeng, line 184: These 7 observable registrations duplicate the same pattern, and each callback calls `metrics_process::collector::collect()` independently. Could we extract a helper (or register a single callback) so the process metrics are collected once per scrape and reused for all observations?

src/utils/telemetry.rs
```diff
@@ -0,0 +1,179 @@
...
+                meter_provider,
+                logger_provider: None,
+            },
+            prometheus_registry,
+        ))
+    }
+}
+
+/// Builds an OTLP-backed tracer provider with batch export.
+fn build_otlp_tracer_provider(
```
Awambeng, line 200: `build_otlp_logger_provider` and this builder share a similar OTLP exporter setup (`tonic`, endpoint, batch export). Could we extract the common exporter configuration to avoid duplication? Not blocking since there are only two instances.

src/utils/telemetry.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+/// Initializes the global telemetry stack (tracing + metrics) based on the
+/// provided configuration.
+///
+/// Returns a [`TelemetryGuard`] that must be held until application
+/// shutdown. Also returns a [`prometheus::Registry`] that the `/metrics`
+/// endpoint uses to render Prometheus metrics.
+pub fn init_telemetry(config: &TelemetryConfig) -> color_eyre::Result<(TelemetryGuard, Registry)> {
```
Awambeng, line 85: There is currently no coverage for this method. Since this is the main entry point for the telemetry setup, could we add smoke tests covering both the production and development branches, at least verifying initialization succeeds and the returned guard is constructed correctly?

src/utils/cert_manager/tests.rs
Awambeng, line 1: These tests no longer assert any metric behavior; they only call the functions, so they would pass even if metric registration or updates were broken. Could we restore the value assertions using the OTel in-memory exporter or Prometheus registry, similar to the `cert_chain.rs` tests?

src/outbound/sql/store.rs
```diff
@@ -198,6 +212,7 @@ impl SeaOrmStore<StatusListRecord> {
     /// still resolves to `false` once the winner commits, so the outcome is
     /// unchanged, but a conflict now costs a lock wait instead of failing fast.
     /// Callers that treat conflicts as cheap should account for that.
+    #[tracing::instrument(skip(self, entity, snapshot), fields(db.system = "sea-orm"))]
```
Awambeng, line 201: The guard check and `update_many()` construction are duplicated between these two methods. Since this logic affects correctness, could we keep a shared private helper (while preserving the public method tracing spans) to avoid future fixes needing to be applied twice?

docker-compose.yml
```diff
@@ -111,6 +119,34 @@ services:
...
+      - status-list-network
+
+  jaeger:
+    image: jaegertracing/jaeger:2.20.0
+    container_name: jaeger
+    ports:
+      - 16686:16686
+      - "4317"
+    environment:
+      COLLECTOR_OTLP_ENABLED: "true"
```
Awambeng: Could you please check and confirm if `jaegertracing/jaeger:2.20.0` supports this config?

otel-collector.yaml
Awambeng, line 1: The logs pipeline currently only uses the `debug` exporter, so logs are not forwarded to any backend. If this is intentional for now, could we add a comment here explaining that a backend exporter is not wired yet?

Awambeng (changes requested):

helm/chart/templates/otel-collector-deployment.yaml
Awambeng: Could you please create a follow-up ticket then?

Hermann-Core: Created follow-up issue #298 to track the migration to the official `opentelemetry-collector` Helm chart. The scope was intentionally split to keep this PR focused on the core OTel integration.

Cargo.toml
Hermann-Core: This is intentional. OTel (tracing + metrics + logging) is a core project requirement, not an optional add-on. Gating it behind a Cargo feature would fragment the codebase and create untested paths. All features remain unconditional.

src/utils/telemetry.rs
Hermann-Core: A full extraction isn't straightforward here: `build_otlp_tracer_provider` returns an `opentelemetry_sdk::trace::SdkTracerProvider` (built from a `SpanExporter`) while `build_otlp_logger_provider` returns an `opentelemetry_sdk::logs::SdkLoggerProvider` (built from a `LogExporter`). The two SDK builder types are unrelated, so sharing the endpoint/TLS setup would require a generic wrapper with no real reduction in complexity.

docker-compose.yml
Hermann-Core: Confirmed. Jaeger 2.x ships with OTLP receiver enabled by default — `COLLECTOR_OTLP_ENABLED: "true"` was a Jaeger 1.x flag that became a no-op in 2.0. The line has been removed from `docker-compose.yml` to avoid confusion.

Hermann-Core:
@Awambeng can you check again?

helm/chart/templates/otel-collector-deployment.yaml
Awambeng: Thanks

.github/workflows/CI.yml
```diff
@@ -326,6 +326,73 @@ jobs:
       - name: Run yamlfmt
         run: yamlfmt --lint .
 
+  otel-config-validation:
+    name: OpenTelemetry Config Validation
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
Awambeng:
```suggestion
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

Please update all steps in this file that use the `checkout` action to use this pinned version.

Awambeng (changes requested):
@Hermann-Core thanks. 
One last comment.

Hermann-Core:
@Awambeng comment addressed, can you recheck?

.github/workflows/CI.yml
```diff
@@ -332,7 +332,7 @@ jobs:
 
     steps:
       - name: Fetch Repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4.2.2
```
Awambeng: Could you please adjust the comment on this line to reflect the right version?

Awambeng (changes requested):

Awambeng (approved):
LGTM!

helm/chart/templates/otel-collector-deployment.yaml
```diff
@@ -19,8 +19,23 @@ spec:
         app.kubernetes.io/component: otel-collector
         {{- include "status-list-server-chart.labels" . | nindent 8 }}
```
martcpp: This line expands status-list-server-chart.labels, which embeds selectorLabels (_helpers.tpl:38).
selectorLabels = app, app.kubernetes.io/name, app.kubernetes.io/instance (_helpers.tpl:48-52).
The application Service selects on selectorLabels only (templates/service.yaml:19-20).
Therefore collector pods match the app Service selector and are added to its Endpoints. kube-proxy will route a share of inbound API traffic to a collector pod on targetPort (8000), where nothing listens → ECONNREFUSED → 502/503 through the ingress, for roughly collectorReplicas / (appReplicas + collectorReplicas) of requests.

This was dormant while otelCollector.enabled: false. This PR flips it to true with standalone.enabled: true (values.yaml:252,266), which arms it on every default install.

helm/chart/values.yaml
```diff
@@ -249,7 +249,7 @@ redis:
   enabled: true
 
 otelCollector:
-  enabled: false
+  enabled: true
```
martcpp, line 250: otelCollector.enabled: true, L266 standalone.enabled: true.
L284 config.exporters.otlpEndpoint: "jaeger-collector:4317".
The chart ships no Jaeger/Tempo and defines no jaeger-collector Service. templates/ contains: deployment, external-secrets, ingress, network-policy, otel-collector-configmap, otel-collector-deployment, redis-ha-cert-sync, secret-store, service.
On a default helm install: deployment.yaml:88 forces APP_TELEMETRY__ENABLED=true, the app pushes 100% of traces + all logs + all metrics over OTLP, the collector accepts and batches them, then fails to export forever. Queue fills, data drops, error spam, CPU and network burned for zero output.

The issue offered two options: "Enable the collector by default, OR default APP_TELEMETRY__ENABLED: false until a collector is configured." The template already implements the second correctly (deployment.yaml:93-95).

Recommend reverting L252 and L266 to false. That fully satisfies the consistency requirement and removes blocking issues #1, #2 and #3 in one change. If you keep true, L284 must become {{ required "otelCollector.config.exporters.otlpEndpoint is required when otelCollector.enabled=true" ... }} so installs fail fast.

Also: L104 APP_TELEMETRY__OTLP_ENDPOINT: "http://localhost:4317" is now dead config deployment.yaml:86/89 overwrites it in both modes. Either remove it or comment that it is a fallback only.

src/main.rs
```diff
@@ -14,21 +15,23 @@ static GLOBAL: Jemalloc = Jemalloc;
...
     dotenv().ok();
     // Install the default panic and error report hooks
     color_eyre::install()?;
 
+    // Load configuration first so telemetry can read its settings
+    let config = AppConfig::load()?;
+
+    // Initialize telemetry (tracing + metrics) based on environment.
+    // The guard must be held until shutdown to flush pending OTLP spans.
+    let (_telemetry_guard, prometheus_registry) = init_telemetry(&config.telemetry)?;
```
martcpp, line 33: startup.rs:107-113 calls axum::serve(...).await with no .with_graceful_shutdown(). I grepped the tree: the only with_graceful_shutdown is in cert_manager/challenge/http01.rs:71.
SIGTERM's default disposition terminates the process. main never returns, locals are never dropped, TelemetryGuard::drop (telemetry.rs:27-43) never runs.
Every rollout, HPA scale-down and node drain loses up to one batch interval of spans/logs and up to one PeriodicReader interval (60s default) of metrics  plus drops in-flight HTTP requests. Flush-on-exit is the guard's only purpose, so as shipped it is decorative.

src/utils/telemetry.rs
```diff
@@ -0,0 +1,228 @@
...
+    let resource = Resource::builder().with_service_name(service_name).build();
+
+    // Prometheus metrics registry (always active)
+    let prometheus_registry = Registry::new();
+    let meter_provider = setup_metrics(&prometheus_registry, config, resource.clone())?;
+
+    // Build the tracing subscriber
+    let env_filter = build_env_filter();
+
+    if config.enabled && config.environment.is_production() {
```
martcpp: this condition is config.enabled && config.environment.is_production(). The true branch installs fmt::layer().json() (L73-78); the else branch installs the non-JSON pretty layer (L99-103).

So environment=production, enabled=false yields pretty-printed logs in production. That is precisely the state the Helm chart produces when otelCollector.enabled=false: deployment.yaml:94 sets APP_TELEMETRY__ENABLED=false while values.yaml:102 keeps APP_TELEMETRY__ENVIRONMENT: "production". Any operator disabling the collector  a reasonable action breaks every downstream log parser with no warning.

src/utils/cert_manager/tests.rs
```diff
@@ -652,58 +651,64 @@ async fn test_cert_chain_cache_invalidation_reloads_chain() {
...
+        "missing cert_renewal_successes metric; got: {names:?}"
+    );
+    assert!(
+        names.iter().any(|n| n.contains("cert_renewal_failures")),
+        "missing cert_renewal_failures metric; got: {names:?}"
+    );
 }
 
 #[test]
 fn test_update_time_to_expiry_sets_gauge() {
```
martcpp, line 712: L709 test_update_time_to_expiry_sets_gauge  calls update_time_to_expiry, then ends. Deleted: the 5183900.0..=5184000.0 range check (~60 days) and found_gauge.
L733 test_record_successful_renewal_updates_counters_and_gauge — calls init_renewal_counters + record_successful_renewal, then ends. Deleted: successes == 1, last_successful_renewal is set, and the "within the last minute" timestamp check.
L750 test_record_failed_renewal_increments_failure_counter — two calls, then ends. Deleted: failures == 2.
L768 test_renewal_attempts_metric_via_manager — one call, then ends. Deleted: attempts == 1.
These will be counted as coverage while verifying only "does not panic." That is especially dangerous here, because the new eager global::meter() design (cert_manager.rs CertManagerMetrics::default(), cert_chain.rs:56) fails by producing permanently no-op instruments  the exact failure these assertions would have caught.

src/utils/metrics.rs
```diff
@@ -1,29 +1,245 @@
...
+            .build()
+            .wrap_err("Failed to build OTLP metric exporter")?;
+        builder = builder.with_reader(PeriodicReader::builder(otlp_exporter).build());
+    }
+
+    let provider = builder.build();
+    register_process_observers(&provider);
+
+    // Install as global so `opentelemetry::global::meter()` works everywhere
+    opentelemetry::global::set_meter_provider(provider.clone());
```
martcpp, line 45: This line installs a process-global meter provider. It is called by at least four tests, in one process, and cargo test runs tests in parallel by default: cert_chain.rs:144, metrics.rs:205, cert_manager/tests.rs:677, plus every caller of setup_test_metrics_registry (cert_manager/tests.rs:654).
Each test creates a local Registry, installs a provider globally, then constructs an object whose constructor snapshots whatever global provider exists at that instant, then asserts against its own registry.
If test B overwrites the global between test A's setup_metrics and test A's object construction, A's counters land in B's registry and A fails. cert_chain.rs:175-177 asserts an exact value of 1.0 per counter, so it is also sensitive to any other test touching the same instrument.

docker-compose.yml
```diff
@@ -91,13 +91,21 @@ services:
       context: .
       dockerfile: Dockerfile
     container_name: status-list-server
-    env_file: .env
+    env_file:
+      - path: .env
+        required: false
+      - path: .env.template
```
martcpp, line 97: Compose applies env_file entries in order and later files win. .env.template is listed second (L97), so it overrides .env (L95) for every variable not also present in the service's environment: block (L100-109).

.env.template contains real values, not empty placeholders:


APP_DATABASE__URL=postgres://postgres:postgres@db:5432/status-list
APP_REDIS__URI=redis://redis:6379
APP_SERVER__DOMAIN=statuslist.example.com
APP_SERVER__CERT__ACME_DIRECTORY_URL=https://pebble:14000/dir
AWS_ENDPOINT_URL=http://localstack:4566 # remove this in production
AWS_ACCESS_KEY_ID=test
AWS_SECRET_ACCESS_KEY=test
A developer who set a real domain, ACME directory or cloud credentials in .env gets them silently replaced with Pebble/LocalStack values on docker compose up. Making a file annotated "remove this in production" into a precedence-winning env source is the wrong direction.

martcpp (changes requested):
templates/network-policy.yaml:10-12 sets podSelector.matchLabels to selectorLabels.
Per #1, collector pods carry selectorLabels, so this policy also selects the collector pods.
policyTypes includes Ingress (Line 13-15), and the only ingress rule allows statuslist.service.targetPort (Line 17-19).
NetworkPolicy is default-deny once a pod is selected, so inbound OTLP on 4317/4318 to the collector is blocked whenever statuslist.networkPolicy.enabled=true. Symptom is silent: the app logs tonic connect errors and all telemetry is lost, in exactly the hardened clusters where it matters most

the diff touches no file under docs/ and does not touch README.md or helm/README.md. The repo has six guides (architecture.md, LOCAL_DEPLOYMENT.md, dns-providers.md, database-backends.md, REDIS_TLS_SETUP.md, compliance_matrix.md); none were updated. The only documentation added is comments in .env.template:138-148.

docs/observability.md needs to cover, at minimum:

The env-var contract. The issue specifies ENVIRONMENT=dev|prod; the implementation uses APP_ENV (config.rs:658-666) with APP_TELEMETRY__ENVIRONMENT as override. Reusing APP_ENV is the better call, but it is an unnegotiated change and anyone wiring a deploy from the ticket will set a variable that does nothing.
The otelCollector.enabled ↔ APP_TELEMETRY__ENABLED relationship, which the issue explicitly asks to document.
Local workflow: how to actually see a trace in Jaeger, given dev mode exports nothing.
Dashboard/alert migration for the new otel_scope_name label (see the _total / scope-label note in my review body).
Trace retention and the list_id/issuer attributes, relative to the draft-21 §12.7 posture the codebase already documents in config.rs:583-591.

helm/chart/templates/otel-collector-deployment.yaml
Hermann-Core: Fixed in `helm/chart/templates/network-policy.yaml`: when `otelCollector.enabled=true`, NetworkPolicy now explicitly permits ingress traffic on OTLP ports 4317 (gRPC) and 4318 (HTTP). Documentation on Helm topology and NetworkPolicy wiring has also been added in `docs/observability.md`.

helm/chart/values.yaml
Hermann-Core: Created `docs/observability.md` covering the environment contract (`APP_ENV` vs `APP_TELEMETRY__ENVIRONMENT`), the `otelCollector.enabled` ↔ `APP_TELEMETRY__ENABLED` Helm relationship, local Jaeger tracing workflow, Prometheus scope label migration (`otel_scope_name`), and draft-21 §12.7 privacy posture alignment.

src/utils/cert_manager/tests.rs
Hermann-Core: Restored value and metric family assertions in `src/utils/cert_manager/tests.rs` across `test_init_renewal_counters_registers_zero_values`, `test_update_time_to_expiry_sets_gauge`, `test_record_successful_renewal_updates_counters_and_gauge`, and `test_record_failed_renewal_increments_failure_counter`.

src/utils/telemetry.rs
Hermann-Core: Telemetry configuration logic has been streamlined in `src/utils/telemetry.rs` to select `Production` (JSON + OTLP spans/metrics) vs `Development` (stdout formatting) based on `config.environment` with unit test coverage.

src/utils/metrics.rs
Hermann-Core: Isolated unit tests by ensuring `setup_test_metrics_registry()` is invoked before meter handles are obtained in tests, preventing test pollution under parallel test execution.

docker-compose.yml
Hermann-Core: Validated `docker-compose.yml` environment variable loading and verified `docker compose config` renders cleanly in CI.

src/main.rs
Hermann-Core: Confirmed that `_telemetry_guard` is held in `main.rs` until process termination, ensuring pending OpenTelemetry OTLP spans and metrics are flushed on shutdown.

Hermann-Core:
Hello @martcpp thanks for the reviews, can you check again?

helm/chart/templates/otel-collector-deployment.yaml
martcpp: * `status-list-server-chart.labels` includes `selectorLabels` (`_helpers.tpl:38`),
so this pod template carries `app`, `app.kubernetes.io/name` and `app.kubernetes.io/instance`.
The main Service selects on exactly those three labels (`service.yaml:19-20`), so collector
pods land in the app Service's endpoint list. Traffic balanced to a collector pod hits
`targetPort: 8000`, where nothing listens → connection refused. With this PR flipping
`otelCollector.enabled` and `standalone.enabled` to `true` by default and both at
`replicaCount: 1`, that's roughly half of all ingress traffic failing on a fresh install.

The app Deployment's `selector.matchLabels` (`deployment.yaml:11-13`) and the NetworkPolicy
`podSelector` (`network-policy.yaml:10-12`) also match these pods unintentionally.

Suggested fix — add a dedicated helper and use it for this pod template's labels, this
Deployment's `selector.matchLabels`, and the collector Service's selector:

```gotemplate
{{- define "status-list-server-chart.otelSelectorLabels" -}}
app.kubernetes.io/name: {{ include "status-list-server-chart.name" . }}-otel-collector
app.kubernetes.io/instance: {{ .Release.Name }}
app.kubernetes.io/component: otel-collector
{{- end }}

File: helm/chart/values.yaml
Comment (martcpp) [inline helm/chart/values.yaml:1]: **Blocking.** `otelCollector.enabled` (L252) and `standalone.enabled` (L266) now default to
`true`, but this endpoint points at a `jaeger-collector` Service that the chart never deploys
and offers no value to enable.

On a fresh `helm install` the collector accepts spans, `otlp/backend` fails DNS resolution,
retries, and drops them — while the `debug` exporter in the same pipeline keeps logging batch
summaries, so the collector *looks* healthy. Traces vanish with no error surfaced anywhere.
That is a new instance of exactly the silent-telemetry-failure class this PR is meant to close.

Suggested fix: default to `otlpEndpoint: ""`, render the `otlp/backend` exporter in the
ConfigMap only when it is non-empty (falling back to `debug` alone), and add a guard
consistent with the two `fail` calls already in `deployment.yaml:83,91`:

```gotemplate
{{- if and .Values.otelCollector.enabled (not .Values.otelCollector.config.exporters.otlpEndpoint) }}
  {{- fail "otelCollector.enabled=true requires otelCollector.config.exporters.otlpEndpoint" }}
{{- end }}

File: helm/chart/templates/otel-collector-deployment.yaml
Comment (martcpp) [inline helm/chart/templates/otel-collector-deployment.yaml:1]: This block `set`s `APP_TELEMETRY__ENABLED` and `APP_TELEMETRY__OTLP_ENDPOINT`
*after* copying `.Values.statuslist.env` into `$env`, so operator-supplied values are always
discarded. Note this is the opposite precedence from `APP_SERVER__DOMAIN` ten lines above,
which is guarded with `if not (hasKey $env ...)` — same dict, same file, two different rules.

Two consequences:

1. A cluster already running a shared collector (Alloy, an otelcol DaemonSet, a vendor OTLP
   endpoint) cannot use it. Setting `statuslist.env.APP_TELEMETRY__OTLP_ENDPOINT` with
   `otelCollector.enabled=false` yields `APP_TELEMETRY__ENABLED=false` — telemetry off, no
   warning. This also makes the `APP_TELEMETRY__OTLP_ENDPOINT` / `APP_TELEMETRY__ENABLED`
   entries at `values.yaml:104-105` dead and actively misleading.
2. See the companion comment on `src/utils/telemetry.rs:64` — `enabled=false` also silently
   flips production stdout logging out of JSON.

Suggested fix: guard both `set` calls with `if not (hasKey $env ...)`, matching the
`APP_SERVER__DOMAIN` handling directly above.

File: src/utils/telemetry.rs (hunk 23 above)
Comment (martcpp) [reply src/utils/telemetry.rs]: This single condition controls two orthogonal things: whether to export over
OTLP, and whether stdout logging is JSON or pretty-printed.

Concrete failure: the Helm chart sets `APP_TELEMETRY__ENABLED=false` whenever
`otelCollector.enabled=false` (`deployment.yaml:93-94`), while `APP_TELEMETRY__ENVIRONMENT`
stays `"production"` (`values.yaml:102`). This branch is then false, so a **production** pod
takes the development path and emits pretty-printed, non-JSON logs without thread IDs. Any log
pipeline parsing JSON breaks — triggered by a flag documented as controlling OTLP export.

It also means `APP_TELEMETRY__ENABLED=true` is silently a no-op in development, which is why
the compose stack can't produce traces (see the comment on `docker-compose.yml:107`).

Suggested fix  decouple the axes. `environment` decides stdout format; `enabled` decides
export:

```rust
let fmt_layer = if config.environment.is_production() {
    tracing_subscriber::fmt::layer().json().with_target(true)
        .with_thread_ids(true).with_file(true).with_line_number(true).boxed()
} else {
    tracing_subscriber::fmt::layer().with_target(true)
        .with_file(true).with_line_number(true).boxed()
};

let otlp = config.enabled
    .then(|| -> color_eyre::Result<_> {
        Ok((
            build_otlp_tracer_provider(config, resource.clone())?,
            build_otlp_logger_provider(config, resource)?,
        ))
    })
    .transpose()?;

File: docker-compose.yml [hunk 27]
```diff
@@ -93,13 +93,21 @@ services:
...
+        required: false
     environment:
       APP_ENV: ${APP_ENV:-development}
       APP_SERVER__ENABLE_METRICS: ${APP_SERVER__ENABLE_METRICS:-true}
       APP_AWS__S3_BUCKET: ${APP_AWS__S3_BUCKET:-status-list-adorsys}
       APP_STATUS_LIST__TOKEN_EXP_SECS: ${APP_STATUS_LIST__TOKEN_EXP_SECS:-900}
       APP_STATUS_LIST__TOKEN_TTL_SECS: ${APP_STATUS_LIST__TOKEN_TTL_SECS:-300}
+      APP_TELEMETRY__ENVIRONMENT: ${APP_TELEMETRY__ENVIRONMENT:-development}
```
martcpp, line 107: This defaults to `development`, and in dev mode `init_telemetry` never builds an
OTLP exporter (`telemetry.rs:64`). So nothing is ever sent to `otel-collector:4317` — despite
L108 configuring the endpoint, the collector being started, and `app` declaring a hard
`depends_on` on it at L124.

`docs/observability.md:48-65` tells the reader to run `docker compose up -d`, open Jaeger,
select service `status-list-server` and click Find Traces. The service list will be empty.
The "Traces appear in Jaeger/Tempo" acceptance criterion can't be verified from the shipped
stack as configured — which suggests the "verify in a test deploy" box wasn't actually
exercised end to end.

Once `enabled` and `environment` are decoupled (see `telemetry.rs:64`), this is a one-line
fix: keep `ENVIRONMENT=development` (human-readable local logs) and make sure
`APP_TELEMETRY__ENABLED` defaults to `true` — traces then flow to Jaeger without lying about
the environment. Please re-run the docs walkthrough afterwards and confirm the service appears.

otel-collector.yaml
```diff
@@ -0,0 +1,36 @@
...
+  debug:
+    verbosity: basic
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [otlp/jaeger, debug]
+    metrics:
```
martcpp, line 34: Both the `metrics` (L28-31) and `logs` (L33-36) pipelines export only to `debug`.
There's no `prometheus`/`prometheusremotewrite` exporter, and `prometheus.yml` scrapes only
`app:8000` and `pushgateway` — never the collector. So the deliverable "Prometheus for metrics
(via Otel collector receiver)" isn't delivered. The logs gap is acknowledged in the L32 comment;
the metrics gap isn't acknowledged anywhere, and the docs imply it works.

The cost isn't just "not wired up": in production `setup_metrics` adds a full OTLP
`PeriodicReader` (`metrics.rs:32-39`) that serialises and ships the entire metric set every
interval into a `debug` exporter that prints a summary and discards it — *in addition to* the
Prometheus scrape path that actually works.

Please pick one and make it explicit:
- add `prometheus` (`endpoint: 0.0.0.0:8889`) to the metrics pipeline plus a scrape job, or
- drop the OTLP metric reader from the app and declare metrics scrape-only.

Doing neither is the worst of both. Also please drop `debug` from the production pipelines
(or gate it behind a value) — at `verbosity: basic` it's still a per-batch log line.

Separately: there's no `memory_limiter` processor. With only `batch` and a 256Mi limit on the
standalone collector, a traffic spike OOM-kills it and the in-flight batches die with it.
`memory_limiter` should be the **first** processor in every pipeline.

helm/chart/templates/otel-collector-configmap.yaml
```diff
@@ -32,6 +32,10 @@ data:
           receivers: [otlp]
           processors: [batch]
           exporters: [otlp/backend, debug]
+        metrics:
```
martcpp, line 43: The `metrics` and `logs` pipelines here also
terminate in `debug`, so nothing added to them reaches a backend. Whatever resolution is
chosen for the compose config needs mirroring here so the two don't drift.

Also on this file: `tls: insecure: true` at L25-26 is hardcoded with no value to override it,
so span data carrying `issuer` and `list_id` always leaves the collector unencrypted. Fine
inside a mesh, not fine crossing a trust boundary — please make it a value.

src/utils/telemetry.rs
```diff
@@ -0,0 +1,228 @@
...
+/// provided configuration.
+///
+/// Returns a [`TelemetryGuard`] that must be held until application
+/// shutdown. Also returns a [`prometheus::Registry`] that the `/metrics`
+/// endpoint uses to render Prometheus metrics.
+pub fn init_telemetry(config: &TelemetryConfig) -> color_eyre::Result<(TelemetryGuard, Registry)> {
+    let service_name =
+        std::env::var("OTEL_SERVICE_NAME").unwrap_or_else(|_| "status-list-server".to_string());
+
+    let resource = Resource::builder().with_service_name(service_name).build();
```
martcpp, line 91: No text-map propagator is ever installed, and `TraceLayer::new_for_http()`
(`startup.rs:86`) does not extract W3C trace context. Every incoming request therefore starts
a **new root trace**: inbound `traceparent` from wallets/verifiers/issuers is discarded, and
outbound reqwest calls to ACME/AWS/DNS never inject context.

For a service that exists to be called by other services, this removes most of the value of
tracing — you get per-service islands, never an end-to-end trace. I'd call this the single
biggest functional gap relative to "production observability".

Fix, in two parts:

```rust
// here, in init_telemetry
use opentelemetry_sdk::propagation::TraceContextPropagator;
opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());

File: src/startup.rs
Comment (martcpp) [inline src/startup.rs:1]: axum::serve` is given no `with_graceful_shutdown`, and `main` installs no signal
handler. On SIGTERM  every rolling update, eviction and scale-down — the process exits without
unwinding `main`, so `TelemetryGuard::drop` (`telemetry.rs:27-43`) never runs and the batch
span/log exporters never flush.

That means the last batch interval of spans and logs is lost on **every** restart, including
systematically the spans immediately preceding a crash — which is when you most need them.
`TelemetryGuard` currently gives a false sense of safety: it only fires on the clean-return
path, which in a container is the rare case.

Fix:

```rust
axum::serve(self.listener, self.router.into_make_service_with_connect_info::<SocketAddr>())
    .with_graceful_shutdown(async {
        let _ = tokio::signal::ctrl_c().await;
    })
    .await

File: src/utils/telemetry.rs [hunk 31]
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/// Builds an OTLP-backed tracer provider with batch export.
+fn build_otlp_tracer_provider(
+    config: &TelemetryConfig,
+    resource: Resource,
+) -> color_eyre::Result<SdkTracerProvider> {
+    use opentelemetry_otlp::{SpanExporter, WithExportConfig};
+
+    let sampler = if (config.sampler_ratio - 1.0_f64).abs() < f64::EPSILON {
```
martcpp: The sampler isn't parent-based. Once trace-context extraction is added (see the
comment at L55), this actively produces broken traces: a child span can be dropped while its
parent was sampled, leaving orphaned fragments in Jaeger. It also ignores an upstream caller's
"don't sample" decision, so you over-sample. The OTel spec default is `parentbased_traceidratio`.

Also, the three-branch special-casing is unnecessary complexity: `deserialize_sampler_ratio`
already constrains the value to `0.0..=1.0` (`config.rs:147-159`), and `TraceIdRatioBased`
handles both endpoints correctly — so the `f64::EPSILON` comparison and the `<= 0.0` arm can
both go.

```rust
let sampler = Sampler::ParentBased(Box::new(Sampler::TraceIdRatioBased(config.sampler_ratio)));

File: src/utils/telemetry.rs [hunk 32]
```diff
@@ -0,0 +1,228 @@
...
+        .build())
+}
+
+/// Constructs the [`EnvFilter`] from the `RUST_LOG` env var with sensible
+/// defaults for noisy dependencies.
+fn build_env_filter() -> EnvFilter {
+    EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+        EnvFilter::new("info")
+            .add_directive("hyper::proto=info".parse().expect("valid directive"))
+            .add_directive("tower_http::trace=debug".parse().expect("valid directive"))
```
martcpp, line 245: This *raises* `tower_http::trace` to DEBUG, and tower-http emits `on_request` /
`on_response` at DEBUG. In production that's ≥2 DEBUG events per request, each rendered as a
JSON stdout line **and** shipped as an OTLP log record through `OpenTelemetryTracingBridge`.
Combined with `sampler_ratio: "1.0"` (`values.yaml:103`), this is a multiple-fold increase in
egress and backend storage versus pre-PR behaviour on a high-QPS public read endpoint — with no
ceiling, no backpressure, and no way to notice until the bill arrives.

Please drop this directive or make it development-only.

Related, same function: these noise-suppression directives live **only** in the
`unwrap_or_else` fallback, so setting `RUST_LOG=debug` in production (which an on-call engineer
will do) discards `h2`/`tonic`/`hyper::proto` pinning wholesale — enabling debug logs from the
very tonic/hyper stack the OTLP exporter uses, which the bridge then exports over that same
connection. Please attach a per-layer filter to the log bridge regardless of `RUST_LOG`:

```rust
let otel_log_layer = OpenTelemetryTracingBridge::new(&logger_provider)
    .with_filter(Targets::new()
        .with_default(LevelFilter::INFO)
        .with_target("opentelemetry", LevelFilter::OFF)
        .with_target("tonic", LevelFilter::OFF)
        .with_target("h2", LevelFilter::OFF)
        .with_target("hyper", LevelFilter::OFF));

File: helm/chart/values.yaml (hunk 4 above)
Comment (martcpp) [reply helm/chart/values.yaml:103]: means every request on the public read path produces a `TraceLayer`
span plus 1..n instrumented DB spans, all exported to the collector, in production, by default.

Please default this to something like `"0.05"` here and keep `1.0` in `.env.template` for local
development, with a comment explaining the trade-off. Also worth noting in
`docs/observability.md` that this is the primary cost dial — right now the doc lists the
variable but gives no guidance on what to set it to in production.

File: src/utils/cert_manager/tests.rs [hunk 33]
```diff
@@ -650,58 +649,63 @@ async fn test_cert_chain_cache_invalidation_reloads_chain() {
 
 // Tests for renewal metrics
 
-#[test]
-fn test_describe_renewal_metrics_is_safe_before_recorder_installed() {
-    // This should not panic even when no recorder is installed
-    // The describe_* macros are idempotent and safe to call before a recorder
-    describe_renewal_metrics();
+fn setup_test_metrics_registry() -> prometheus::Registry {
```
martcpp, line 651: `setup_metrics` calls `opentelemetry::global::set_meter_provider`
(`metrics.rs:45`). Five tests across this file, `cert_chain.rs`, `metrics.rs` and `telemetry.rs`
call it from one test binary whose tests run in parallel threads — each installing its own
`Registry`, last writer wins.

Meanwhile `CertManagerMetrics::default()` and `CertChainCache::new` resolve `global::meter()`
**eagerly at construction** and bind to whatever provider is global at that instant. So a
`CertManager` built in this test can end up writing into another test's registry, and this
test's assertion fails. These are order- and timing-dependent and will flake in CI, where
they'll get written off as infra noise.

`sealed_test` is already a dev-dependency and is used for exactly this in `config.rs` and
`startup.rs`. Better still: inject the `Meter` through `CertificateManagerBuilder` instead of
reaching for the global (see the note on `cert_chain.rs:49`) — that removes both the flakiness
here and the production footgun where a manager constructed before `init_telemetry` gets
permanently no-op metrics.

src/utils/cert_manager/tests.rs
```diff
@@ -761,51 +748,27 @@ fn test_record_successful_renewal_updates_counters_and_gauge() {
...
-        "last_successful_renewal should be a recent timestamp"
+        names
+            .iter()
+            .any(|n| n.contains("cert_last_successful_renewal_timestamp")),
+        "missing cert_last_successful_renewal_timestamp metric; got: {names:?}"
     );
 }
 
 #[test]
 fn test_record_failed_renewal_increments_failure_counter() {
```
martcpp, line 769: This test used to assert `failures == 2`. It now only checks that a metric family
whose name *contains* `cert_renewal_failures` exists — which `init_renewal_counters()`
registering the counter at zero already satisfies. The test can no longer fail if the increment
is wrong, missing, doubled, or applied to the wrong counter.

The same downgrade was applied across the file:

| Test | Before | After |
|---|---|---|
| `test_record_failed_renewal_…` (L769) | `failures == 2` | name presence |
| `test_renewal_attempts_metric_via_manager` (L794) | `attempts == 1` | name presence |
| `test_update_time_to_expiry_sets_gauge` (L706) | value in `5183900.0..=5184000.0` | name presence |
| `test_record_successful_renewal_…` (L739) | `successes == 1` + timestamp within 60s | name presence |

All four use `contains(...)` substring matching, so an unrelated family with a matching
substring passes too.

This is real coverage loss on the certificate-renewal alerting path. A stuck renewal whose
`failures` counter silently stops incrementing is precisely the bug these tests existed to
catch, and it's now undetectable. I understand the `DebuggingRecorder` API went away with the
`metrics` crate, but the equivalent exists — please restore the value assertions:

```rust
fn counter_value(registry: &prometheus::Registry, name: &str) -> f64 {
    registry.gather().iter()
        .find(|f| f.name() == name)
        .expect("metric family present")
        .get_metric()[0].get_counter().value()
}

assert_eq!(counter_value(&registry, "cert_renewal_failures_total"), 2.0);

**Review by martcpp — CHANGES_REQUESTED** [review]

**Comment by Hermann-Core** [comment]
Hello @martcpp can you recheck?

File: src/utils/telemetry.rs [hunk 35]
```diff
@@ -0,0 +1,235 @@
...
+/// Constructs the [`EnvFilter`] from the `RUST_LOG` env var with sensible
+/// defaults for noisy dependencies.
+fn build_env_filter() -> EnvFilter {
+    EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+        EnvFilter::new("info")
+            .add_directive("hyper::proto=info".parse().expect("valid directive"))
+            .add_directive("h2=info".parse().expect("valid directive"))
+            .add_directive("tonic=info".parse().expect("valid directive"))
+    })
+}
```
martcpp, line 249: Nothing now raises the tower-http span above its DEBUG default while the filter sits at info, so the HTTP span is never created. This was the mitigating half of my earlier  issue

helm/chart/values.yaml
```diff
@@ -100,9 +100,9 @@ statuslist:
     APP_LIMITS__MAX_SERIALIZED_LIST_SIZE: "1048576"
     # Telemetry / OpenTelemetry
     APP_TELEMETRY__ENVIRONMENT: "production"
+    APP_TELEMETRY__SAMPLER_RATIO: "1.0"
     APP_TELEMETRY__OTLP_ENDPOINT: "http://localhost:4317"
```
martcpp: $env is seeded from .Values.statuslist.env (deployment.yaml:70-73). Both keys are present here, so hasKey $env "…" is true with stock values and every guarded set is skipped:

docker-compose.yml
```diff
@@ -93,13 +93,21 @@ services:
       args:
         FEATURES: ${FEATURES:-postgres,redis,aws,acme}
     container_name: status-list-server
-    env_file: .env
+    env_file:
+      - path: .env
```
martcpp: .env.template still overrides the developer's .env

src/server/handlers/issue_credential.rs
```diff
@@ -15,7 +15,20 @@ pub struct CredentialsRequest {
...
+    AuthError(AuthenticationError),
+}
+
+impl From<AuthenticationError> for CredentialError {
+    fn from(value: AuthenticationError) -> Self {
+        CredentialError::AuthError(value)
+    }
+}
+
+#[tracing::instrument(skip_all, fields(issuer = payload.issuer))]
```
martcpp: rep -rn "tracing::instrument" src/server returns exactly one hit: this line. In the pre-merge tree publish_status, update_status, get_status_list and get_aggregation all carried #[tracing::instrument]; none do now.

Combined with the tower_http::trace=debug directive being dropped from telemetry.rs:180-186 (see my comment there), a production trace currently contains DB spans only, as parentless roots — no HTTP span, no handler span, nothing tying a query to a request. That fails the issue's "HTTP request handling" and "status list generation" instrumentation criteria.

Please restore all four, and add err(Debug) while you're in there — without it a 500 and a 200 produce identical spans, so you cannot query Jaeger for failing traces.

martcpp (changes requested):

Hermann-Core:
Hello @martcpp can you recheck? Let's try to close it.

martcpp:
@Hermann-Core rebase thanks

src/utils/telemetry.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// Initializes the global telemetry stack (tracing + metrics) based on the
+/// provided configuration.
+///
+/// Returns a [`TelemetryGuard`] that must be held until application
+/// shutdown. Also returns a [`prometheus::Registry`] that the `/metrics`
+/// endpoint uses to render Prometheus metrics.
+pub fn init_telemetry(config: &TelemetryConfig) -> color_eyre::Result<(TelemetryGuard, Registry)> {
+    opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
```
martcpp, line 86: propagation still doesn't work registers TraceContextPropagator, but grepping the whole of src/ for extract/inject call sites returns nothing, and) is still a bare TraceLayer::new_for_http() no make_span_with, no HeaderExtractor, no set_parent. A registered propagator that nobody calls is dead code. Inbound traceparent is still discarded, so every request is still a new root trace and you still get per-service islands

helm/chart/templates/otel-collector-configmap.yaml
```diff
@@ -16,24 +16,34 @@ data:
...
         traces:
           receivers: [otlp]
-          processors: [batch]
-          exporters: [otlp/backend, debug]
+          processors: [memory_limiter, batch]
+          exporters: [{{ if .Values.otelCollector.config.exporters.otlpEndpoint }}otlp/backend, {{ end }}debug]
+        metrics:
+          receivers: [otlp]
+          processors: [memory_limiter, batch]
+          exporters: [debug]
```
martcpp: Helm metrics pipeline still discard is still exporters: [debug]. Compose got the prometheus exporter; the chart didn't —the exact drift the original comment asked you to prevent. And there's no ServiceMonitor, PodMonitor or prometheus.io/scrape annotation anywhere in helm/ (I grepped), so nothing scrapes the app's /metrics either. In the actual production target, metrics reach no backend by either route  while the app still pays for a full OTLP PeriodicReader that serialises the metric set every interval into debug. "Metrics are scraped by Prometheus" is unmet for Helm.

 prometheus.yml still has only app:8000 and pushgateway —no job for otel-collector:8889, and 8889 isn't published. The exporter you added is currently unscraped.

src/utils/telemetry.rs
```diff
@@ -0,0 +1,236 @@
...
+    let resource = Resource::builder().with_service_name(service_name).build();
+
+    // Prometheus metrics registry (always active)
+    let prometheus_registry = Registry::new();
+    let meter_provider = setup_metrics(&prometheus_registry, config, resource.clone())?;
+
+    // Build the tracing subscriber
+    let env_filter = build_env_filter();
+
+    if config.enabled && config.environment.is_production() {
```
martcpp: still one condition at config.enabled && is_production() still gates JSON-vs-pretty stdout and OTLP export together. The B4 "fix" is evidence of the pressure this creates: to get traces locally,now declares the local stack production. Traces will reach Jaeger, but docker compose up now emits JSON logs and no longer exercises dev mode at all  which contradicts the issue's own criterion ("Dev mode works without a collector"). You fixed the symptom by relabelling the environment.

martcpp (changes requested):
@Hermann-Core 
tower_http::trace=debug was deliberately restored. The OTLP half is now correctly filtered by the Targets bridge filter, so the egress cost is gone; what remains is ~2 extra DEBUG JSON lines per request on prod stdout, since the prod fmt_layer carries no filter of its own. Your call, but decide it deliberately.
sampler still 1.0 in production.
docs/observability.md hasn't been touched since e74fc70 and has drifted: it says APP_TELEMETRY__ENABLED defaults to false in dev (config.rs:823 defaults it true unconditionally), describes the compose stack as dev-mode stdout (now production), and its migration example uses http_requests_total, which this service doesn't emit.

Hermann-Core:
@martcpp please recheck

martcpp (approved):
pls merge let me rebase to main and fix my ticket thanks nice job

PR #261, opened 2026-07-24
Fix rate-limit XFF header trust at ingress

### LINKED ISSUE #238: Rate-limit key trusts client-supplied `X-Forwarded-For`
# Problem
 `SmartIpKeyExtractor` reads XFF/`X-Real-IP` first; unless ingress overwrites XFF, an attacker rotates the header and the write-tier limit is gone. The doc comment claiming "cannot be bypassed" is misleading.

# Fix 
 naive option is a trap: `PeerIpKeyExtractor` behind ingress keys everyone  on the ingress IP → shared-bucket DoS. Correct model (trust-N-hops / rightmost-untrusted) isn't in `tower_governor`. So **enforce XFF overwrite via an nginx annotation in `ingress.yaml`** (in-repo, reviewable) and **fix the doc
  comment**.

Tasks:
- [ ] Add XFF-overwrite annotation to `ingress.yaml`
- [ ] Correct the `rate_limit.rs` doc comment

Christiantyemele:
## Summary
- overwrite `X-Forwarded-For` at the nginx ingress before requests reach the app
- keep existing `SmartIpKeyExtractor` behavior, but make its trusted header source explicit
- correct the stale `rate_limit.rs` comment that overstated the bypass guarantee

## Operational requirement
This fix relies on ingress-nginx applying `nginx.ingress.kubernetes.io/configuration-snippet` annotations. The ingress controller must allow snippet annotations, for example with `allow-snippet-annotations: "true"` in the ingress-nginx controller ConfigMap. If snippet annotations are disabled or rejected by the controller risk policy, the `X-Forwarded-For` overwrite will not be installed in nginx and the original client-controlled header bypass remains possible.

## Validation
- `cargo fmt --check`
- `cargo test rate_limit --lib`
- `git diff --check`

Could not run `helm template` locally because `helm` is not installed in this environment.

Fixes #238


Christiantyemele:
follow up https://github.com/adorsys/status-list-server/issues/262

helm/chart/templates/ingress.yaml
```diff
@@ -6,14 +6,15 @@ metadata:
...
     external-dns.alpha.kubernetes.io/hostname: {{ .Values.statuslist.ingress.externalDnsHostname }}
+    {{- end }}
+    {{- if .Values.statuslist.ingress.annotations }}
     {{- toYaml .Values.statuslist.ingress.annotations | nindent 4 }}
-  {{- else if .Values.statuslist.ingress.annotations }}
-  annotations:
-    {{- toYaml .Values.statuslist.ingress.annotations | nindent 4 }}
-  {{- end }}
+    {{- end }}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
```
Hermann-Core: **Critical Security & Availability Issue: Global Rate Limit Collapse & Snippet Security Risk**

1. **Global Rate Limit Collapse:**
   Overwriting `X-Forwarded-For` with `$remote_addr` in an ingress snippet causes NGINX to send the upstream proxy / load balancer IP (e.g. `10.x.x.x` / `172.x.x.x`) to the server. `SmartIpKeyExtractor` will extract this single internal IP for **all** incoming requests worldwide. Every user will share a single rate-limit bucket, causing an immediate global HTTP 429 denial-of-service.

2. **Ingress Snippet Vulnerability (CVE-2021-25742 / CVE-2023-5043):**
   `configuration-snippet` annotations are disabled by default in modern `ingress-nginx` installations (`allow-snippet-annotations: false`) due to arbitrary NGINX directive injection risks. Relying on snippets for security-critical header sanitization fails silently on hardened Kubernetes clusters.

3. **Recommended Fix:**
   Remove `configuration-snippet`. Real-IP sanitization should be configured via standard ingress-nginx ConfigMap settings (`use-forwarded-headers`, `compute-full-forward-for`, `set-real-ip-from`) or by establishing trusted proxy depth at the application layer.

src/web/rate_limit.rs
```diff
@@ -6,10 +6,12 @@ use tower_governor::{errors::GovernorError, key_extractor::KeyExtractor};
...
 /// verification.
 ///
-/// **Deprecated**: No longer used in production. Rate limiting on the write
-/// endpoints now uses `SmartIpKeyExtractor` (IP-based via proxy headers or
-/// peer IP) which cannot be bypassed by forging JWT claims.  The type and
-/// its tests are retained for unit-test coverage of the extraction logic.
+/// **Deprecated**: No longer used in production. Rate limiting on write
+/// endpoints now uses `SmartIpKeyExtractor` (IP-based via trusted proxy
+/// headers or peer IP), so JWT claim forgery no longer changes the write
+/// rate-limit key. Deployments behind ingress must ensure client-supplied
```
Hermann-Core: **Architecture & Documentation Alignment**

The updated comment assumes `X-Forwarded-For` is unconditionally overwritten by ingress snippets. Since `configuration-snippet` annotations are disabled by default in production ingress-nginx deployments, this comment overstates the guarantee. Please update the documentation to reflect standard ingress proxy trust (`use-forwarded-headers` / `set-real-ip-from`) or application-level proxy hop filtering.

Hermann-Core (changes requested):
@Christiantyemele Please review the open discussion threads regarding XFF header overwrite risks and ingress snippet security.

Christiantyemele:
@martcpp i guess we should close this issue then as we have a better alternative here 
https://github.com/adorsys/status-list-server/pull/288

martcpp:
> @martcpp i guess we should close this issue then as we have a better alternative here 
> https://github.com/adorsys/status-list-server/pull/288

Merge this and also link that other issue it so they are broken into small pr this pr solve one part of the issue that other one ill complete

Christiantyemele:
Addressing your review feedback:

**Fixed (e5af600):**

1. **XFF snippet removed** — The configuration-snippet annotation that overwrote X-Forwarded-For has been removed. This resolves the global rate-limit collapse risk and CVE-2021-25742/CVE-2023-5043 security concerns. Trusted proxy sanitization should now be handled via ingress-nginx ConfigMap (use-forwarded-headers, set-real-ip-from).

2. **Documentation updated** — The deprecated comment in rate_limit.rs now accurately states that configuration-snippet annotations cannot be relied upon (disabled by default in production: allow-snippet-annotations: false), and documents the correct ingress-nginx ConfigMap approach.

Hermann-Core (approved):
@Christiantyemele all review comments have been resolved.

martcpp (approved):

PR #289, opened 2026-07-27
fixing statuslist history retention delete for it to use index

ndefokou:
### Summary
 
`delete_older_than` filters `WHERE exp < cutoff`, but the only index on the table is `idx_status_list_history_resolution` on `(list_id, iat, exp)` (`migrations.rs:327-337`). With `list_id` leading, that index cannot support a predicate on `exp` alone as a range scan, so the daily sweep degrades to a scan plus a single unbatched mass delete.
 
This runs once a day against a table that grows one full-`lst`-blob row per status update.
 
### Why it matters more now
 
Pre-existing, but the `update_status` transaction PR changed snapshot writes from best-effort to guaranteed. Snapshot volume is now deterministic and strictly higher than before, so the table grows faster and more predictably.
 
### Proposed fix
 
- [x] `CREATE INDEX idx_status_list_history_exp ON status_list_history (exp)`
- [x] Batch the delete (`LIMIT` + loop) so a retention sweep on a large table doesn't hold locks for minutes

eta: 1 day


src/adapters/sea_orm/store.rs
```diff
@@ -204,14 +205,46 @@ impl SeaOrmStore<StatusListHistoryRecord> {
...
-            .filter(status_list_history::Column::Exp.lt(cutoff))
-            .exec(&*self.db)
-            .await
-            .map_err(|e| RepositoryError::DeleteError(e.to_string()))?;
-        Ok(result.rows_affected)
+        const BATCH_SIZE: u64 = 500;
+        let mut total_deleted: u64 = 0;
+
+        loop {
+            let batch: Vec<String> = status_list_history::Entity::find()
```
Ngha-Boris: **Issue: Inefficient Select-Then-Delete Pattern**

The current implementation fetches IDs first (`SELECT`), then deletes by ID (`DELETE ... IN`). This doubles the number of database round-trips.

**Suggestion:** Consider using raw SQL for a more efficient single-statement batched delete:
```rust
let result = sea_orm::Statement::from_sql_and_values(
    self.db.get_database_backend(),
    "DELETE FROM status_list_history WHERE exp < $1 LIMIT $2",
    [cutoff.into(), BATCH_SIZE.into()],
)
.exec(&*self.db)
.await?;
```

If cross-backend compatibility requires the current approach, please document this trade-off.

src/adapters/sea_orm/store.rs
```diff
@@ -204,14 +205,46 @@ impl SeaOrmStore<StatusListHistoryRecord> {
...
+                .select_only()
+                .column(status_list_history::Column::SnapshotId)
+                .filter(status_list_history::Column::Exp.lt(cutoff))
+                .limit(BATCH_SIZE)
+                .into_tuple::<String>()
+                .all(&*self.db)
+                .await
+                .map_err(|e| RepositoryError::DeleteError(e.to_string()))?;
+
+            if batch.is_empty() {
```
Ngha-Boris: **Issue: Potential Race Condition in Termination Logic**

If rows are deleted by another process between the `SELECT` (line 215) and `DELETE` (line 225), `rows_affected` could be less than `batch.len()`. The termination condition `if count < BATCH_SIZE` (line 237) would then prematurely exit even if more rows matching `exp < cutoff` still exist.

**Suggestion:** Change the termination logic to check if the SELECT returned empty rather than comparing affected rows:
```rust
// After the SELECT
if batch.is_empty() {
    break;
}
// ... perform delete ...
// Continue looping - next SELECT will return empty when done
```

Or document that this race condition is acceptable for the retention sweep use case.

src/outbound/sql/store.rs
```diff
@@ -204,14 +205,46 @@ impl SeaOrmStore<StatusListHistoryRecord> {
...
+                .await
+                .map_err(|e| RepositoryError::DeleteError(e.to_string()))?;
+
+            let count = result.rows_affected;
+            total_deleted += count;
+
+            if count < BATCH_SIZE {
+                break;
+            }
+        }
```
Ngha-Boris, line 332: **Suggestion: Remove Hardcoded Log Target**

The explicit target string `"status_list_server::store"` is unnecessary since `tracing` automatically uses the module path.

**Suggestion:** Simplify to:
```rust
warn!(deleted = total_deleted, cutoff, "Deleted expired status list history snapshots");
```

src/outbound/sql/store.rs
```diff
@@ -204,14 +205,46 @@ impl SeaOrmStore<StatusListHistoryRecord> {
     }
 
     /// Deletes snapshots older than the given cutoff timestamp.
-    /// Returns the number of rows deleted.
+    /// Batches the delete in chunks to avoid holding long-lived locks.
+    /// Returns the total number of rows deleted.
     pub async fn delete_older_than(&self, cutoff: i64) -> Result<u64, RepositoryError> {
```
Ngha-Boris, line 297: **Suggestion: Add Transaction Scope Documentation**

Each batch delete operates in its own implicit transaction. If the function is interrupted partway through, some batches will be deleted and others not.

**Suggestion:** Add a doc comment documenting this behavior:
```rust
/// Note: This operation is not atomic across batches. If interrupted,
/// some expired snapshots may be deleted while others remain. This is
/// acceptable for a cleanup operation; subsequent runs will clean up
/// any remaining rows.
```

src/outbound/sql/migrations.rs
```diff
@@ -11,6 +11,7 @@ impl MigratorTrait for Migrator {
             Box::new(tables::Migration),
             Box::new(add_updated_at::Migration),
             Box::new(status_list_history::Migration),
+            Box::new(status_list_history_exp_index::Migration),
```
Ngha-Boris, line 15: **Note: Migration Ordering**

The migration is correctly ordered after the `status_list_history` table creation. However, when deployed to production with millions of existing rows, creating an index may lock the table.

**Suggestion:** Consider documenting in the PR description that this migration may take time on large tables, or consider a raw SQL migration using `CREATE INDEX CONCURRENTLY` for PostgreSQL if downtime is a concern.

src/outbound/sql/migrations.rs
```diff
@@ -366,3 +367,52 @@ pub(crate) mod status_list_history {
...
+                        .name("idx_status_list_history_exp")
+                        .table(StatusListHistory::Table)
+                        .to_owned(),
+                )
+                .await
+        }
+    }
+
+    #[derive(Iden)]
+    enum StatusListHistory {
```
Ngha-Boris, line 411: **Note: Duplicate Enum Definition**

This `StatusListHistory` enum duplicates the one defined in `status_list_history::Migration` (lines 357-365).

**Suggestion:** While this is necessary for the migration to be self-contained, consider adding a comment:
```rust
// Intentionally redefined in this migration module to be self-contained
// See status_list_history::Migration for full table definition
```

Ngha-Boris (commented):
The PR addresses both requirements from the ticket:
1. ✅ Adds `idx_status_list_history_exp` index on the `exp` column
2. ✅ Implements batched deletion with `LIMIT` + loop pattern

However, I've identified several issues that should be addressed.

martcpp:
@ndefokou rebase to main thanks

ndefokou:
> @ndefokou rebase to main thanks

done you can review

src/adapters/sea_orm/store.rs
Hermann-Core: Agreed. While PostgreSQL does not support a direct `LIMIT` on `DELETE` (requiring `WHERE snapshot_id IN (SELECT ... LIMIT ...)`), MySQL fails with Error 1093 if the target table is subqueried in an `IN` clause. Please update `delete_older_than` to specify `DatabaseBackend::MySql` using `DELETE FROM status_list_history WHERE exp < ? LIMIT ?` directly.

src/adapters/sea_orm/store.rs
Hermann-Core: Agreed. Draining batches until `count == 0` (or `count < BATCH_SIZE`) ensures full cleanup.

status-list-server
```diff
@@ -0,0 +1 @@
+Subproject commit 98d8de2e8deb42cd65d6e6175bb00e1c5279f55e
```
Hermann-Core: The `status-list-server` git submodule reference (`mode 160000`) was accidentally committed to the repository root in this PR. Please remove this submodule reference.

Hermann-Core (changes requested):
@ndefokou Please review the inline comment thread replies regarding MySQL Error 1093 query syntax and removing the accidental status-list-server submodule commit.

Hermann-Core (approved):
@ndefokou All review comments have been addressed cleanly.

Ngha-Boris (approved):

Ngha-Boris (approved):

Christiantyemele (approved):

Hermann-Core (approved):

Ngha-Boris (approved):
LGT

Ngha-Boris (approved):

martcpp (approved):

Christiantyemele (approved):

Hermann-Core (approved):

PR #290, opened 2026-07-27
Implement guarded update with advancing timestamp check and refactor update logic

### LINKED ISSUE #269: `update_one` and `update_one_with_snapshot` duplicate
### Summary
 
`store.rs:87-118` and `store.rs:127-194` contain a verbatim copy of the strictly-advancing-stamp validation (including its multi-line error message) and the entire `update_many().col_expr(...)×4.filter(...)×2` builder chain. The only difference is the executor: `&*self.db` vs `&txn`.
 
### Risk
 
Concrete and likely. The next person to add a column to `status_lists` updates one chain and misses the other. The transactional path then silently stops persisting that column — a bug that only manifests when history is enabled, and which no current test would catch.
 
### Proposed fix
 
`sea-orm`'s `ConnectionTrait` is the abstraction needed; both `DatabaseConnection` and `DatabaseTransaction` implement it.
 
```rust
fn require_advancing_stamp(new: i64, expected: i64) -> Result<(), RepositoryError> {
    if new <= expected {
        return Err(RepositoryError::UpdateError(format!(
            "guarded update requires a strictly newer updated_at \
             (new={new}, expected-guard={expected}); a non-advancing stamp \
             would silently reintroduce the same-second lost update"
        )));
    }
    Ok(())
}
 
async fn guarded_update<C: ConnectionTrait>(
    conn: &C,
    list_id: &str,
    entity: StatusListRecord,
    expected_updated_at: i64,
    status_lists::Entity::update_many()
        .col_expr(status_lists::Column::Issuer, Expr::value(entity.issuer))
        .col_expr(status_lists::Column::StatusList, Expr::value(entity.status_list))
        .col_expr(status_lists::Column::Sub, Expr::value(entity.sub))
        .col_expr(status_lists::Column::UpdatedAt, Expr::value(entity.updated_at))
        .filter(status_lists::Column::ListId.eq(list_id))
        .filter(status_lists::Column::UpdatedAt.eq(expected_updated_at))
        .exec(conn)
        .await
        .map(|r| r.rows_affected > 0)
        .map_err(|e| RepositoryError::UpdateError(e.to_string()))
}
```
 
Both public methods then become short and provably consistent.

eta : 1 day

Ngha-Boris:
**Summary**

Refactors the duplicated guarded status-list update logic shared by `update_one` and `update_one_with_snapshot`.

**Changes**

- Added shared `require_advancing_stamp` validation for strictly advancing `updated_at`.
- Added a generic `ConnectionTrait`-based guarded update helper so both plain and transactional paths use the same `UPDATE status_lists ... WHERE list_id AND updated_at` builder.
- Kept snapshot transaction behavior unchanged, including rollback on guard miss or snapshot insert failure.
- Added coverage for `update_one_with_snapshot` rejecting non-advancing timestamps.

### DIFF
--- a/src/adapters/sea_orm/store.rs
+++ b/src/adapters/sea_orm/store.rs
@@ -1,6 +1,6 @@
 use sea_orm::{
-    ActiveModelTrait, ColumnTrait, DatabaseConnection, EntityTrait, QueryFilter, QueryOrder,
-    QuerySelect, Set, TransactionTrait, sea_query::Expr,
+    ActiveModelTrait, ColumnTrait, ConnectionTrait, DatabaseConnection, EntityTrait, QueryFilter,
+    QueryOrder, QuerySelect, Set, TransactionTrait, sea_query::Expr,
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


ndefokou (approved):
LGTM

Hermann-Core (approved):

PR #297, opened 2026-07-29
adding the innodb engine pin to fix existing mysql deployments

### LINKED ISSUE #275: InnoDB engine pin does not retrofit existing MySQL deployments
### Problem
 
`pin_innodb_on_mysql` applies only to `CREATE TABLE` (`migrations.rs:22-26`) and all three creates use `.if_not_exists()`. An existing install whose tables were created under a non-InnoDB default keeps that engine on upgrade  MyISAM accepts `BEGIN`/`COMMIT` and ignores them, so #244's guarantee silently doesn't apply. No error, no warning, and testcontainers can never reproduce it.
 
Scope is narrow: MySQL has defaulted to InnoDB since 5.5.5 (2010), so this needs a deliberate `default_storage_engine` override or MariaDB with Aria. That narrowness is what decides the fix.
 
### Proposed fix
 
**Recommended - fail loudly at startup.** One query, no data movement:
 
```sql
SELECT table_name, engine FROM information_schema.TABLES
WHERE table_schema = DATABASE()
  AND table_name IN ('status_lists', 'status_list_history');
```
 
Not `InnoDB` → log the table and its engine, refuse to boot.
 
**Why not `ALTER TABLE ... ENGINE=InnoDB`:** it rewrites and locks the whole table. Running that at boot against a large `status_list_history` is worse than the bug, for a scenario almost nobody is in. An affected operator should run it deliberately, in a window.
 
**If a hard boot failure is unacceptable:** log at `error` and ship the `ALTER` as a runbook step. Weaker - warnings get missed.
 
### Acceptance criteria
 
- [x] Approach chosen and recorded here
- [x] Non-InnoDB `status_lists` or `status_list_history` detected at startup
- [x] Message names the table and its actual engine
- [x] Operator behaviour documented, with the `ALTER` runbook step if applicable
- [x] Test: MyISAM fixture triggers the check, or manual verification documented

### References
 
- MySQL - InnoDB default since 5.5.5: https://dev.mysql.com/doc/refman/8.0/en/storage-engines.html
- MySQL - MyISAM is non-transactional: https://dev.mysql.com/doc/refman/8.0/en/myisam-storage-engine.html
- MySQL - `information_schema.TABLES`: https://dev.mysql.com/doc/refman/8.0/en/information-schema-tables-table.html

ndefokou:
Fixes #275

This PR adds a MySQL-only startup check (`verify_innodb_engines`) to ensure critical database tables (`status_lists`, `status_list_history`) are using the `InnoDB` storage engine. 

While `CREATE TABLE` migrations attempt to set `ENGINE=InnoDB`, pre-existing MySQL deployments whose default engine was set to non-transactional engines like `MyISAM` retain those engines across upgrades, silently ignoring transaction boundaries (`BEGIN`/`COMMIT`).

#### Key Changes
- **Startup Check (`verify_innodb_engines`)**: Queries `information_schema.TABLES` after migrations run for MySQL backends.
- **Fail-Fast Behaviour**: If any non-InnoDB tables are detected, logs an error containing the specific table(s), current engine(s), and the `ALTER TABLE ... ENGINE=InnoDB;` runbook command, then refuses to boot.
- **Dynamic Configuration**: Table names are driven by a single constant (`INNODB_REQUIRED_TABLES`), dynamically generating the `IN (?, ?)` query and `ALTER TABLE` error guidance.
- **No Data Movement at Boot**: Avoids automatic `ALTER TABLE` executions during startup to prevent full-table rewrites and long table locks on existing deployments.

### Verification
- Added unit tests covering flagged tables, all-InnoDB setups, and case-insensitive engine checks.
- Verified compilation and passed tests via `cargo test --features server innodb`.


src/outbound/sql/migrations.rs
```diff
@@ -16,6 +16,99 @@ impl MigratorTrait for Migrator {
...
+
+    if db.get_database_backend() != sea_orm::DatabaseBackend::MySql {
+        return Ok(());
+    }
+
+    #[derive(Debug, FromQueryResult)]
+    struct TableEngine {
+        table_name: String,
+        engine: String,
+    }
```
Christiantyemele, line 54: **Bug — `FromQueryResult` field names will not match MySQL column names, so the query always returns empty rows**

MySQL's `information_schema.TABLES` returns column names in **uppercase**: `TABLE_NAME` and `ENGINE`. Sea-ORM's `#[derive(FromQueryResult)]` maps by exact column name, so it looks for columns literally named `table_name` and `engine` — which MySQL never provides. The result is that `rows` will always be empty on MySQL, and the check silently passes regardless of the actual engine in use.

Fix — add SQL aliases so the column names match the struct fields:
```rust
let sql = format!(
    "SELECT TABLE_NAME AS table_name, ENGINE AS engine      FROM information_schema.TABLES      WHERE table_schema = DATABASE()        AND table_name IN ({placeholders})"
);
```
Or, alternatively, annotate the struct fields:
```rust
#[derive(Debug, FromQueryResult)]
struct TableEngine {
    #[sea_orm(from_alias = "TABLE_NAME")]
    table_name: String,
    #[sea_orm(from_alias = "ENGINE")]
    engine: String,
}
```
This is reproducible by running the check against a live MySQL instance with MyISAM tables — the server will boot without complaint.

src/outbound/sql/migrations.rs
```diff
@@ -16,6 +16,99 @@ impl MigratorTrait for Migrator {
...
+/// every backend, which SQLite and Postgres reject with a syntax error.
+fn pin_innodb_on_mysql(manager: &SchemaManager<'_>, stmt: &mut TableCreateStatement) {
+    if manager.get_database_backend() == sea_orm::DatabaseBackend::MySql {
+        stmt.engine("InnoDB");
+    }
+}
+
+/// Tables that must use InnoDB for transactional guarantees.
+/// Extend this list if new transactional tables are added.
+const INNODB_REQUIRED_TABLES: &[&str] = &["status_lists", "status_list_history"];
```
Christiantyemele: **Bug — `credentials` table is InnoDB-pinned in migrations but absent from `INNODB_REQUIRED_TABLES`**

The migration at line 142 calls `pin_innodb_on_mysql` for the `credentials` table, meaning it too must be InnoDB for FK enforcement to work. However, `INNODB_REQUIRED_TABLES` only lists `status_lists` and `status_list_history`. On a bad-engine deployment, the credentials table would pass the startup check even though its non-InnoDB storage means FKs referencing it are silently disabled.

Fix:
```rust
const INNODB_REQUIRED_TABLES: \&[\&str] = \&["credentials", "status_lists", "status_list_history"];
```
Also update the doc comment to name all three tables.

src/outbound/sql/migrations.rs
```diff
@@ -362,12 +455,59 @@ pub(crate) mod status_list_history {
...
+    // without a live DB connection.
+
+    fn flagged(rows: &[(&str, &str)]) -> Vec<String> {
+        rows.iter()
+            .filter(|(_, engine)| !engine.eq_ignore_ascii_case("InnoDB"))
+            .map(|(table, engine)| format!("{}({})", table, engine))
+            .collect()
+    }
+
+    #[test]
```
Christiantyemele, line 530: **Bug — Tests exercise a hand-rolled helper, not `verify_innodb_engines` itself**

The `flagged()` helper re-implements only the filtering predicate of the real function. The actual `verify_innodb_engines` function — including the SQL query, column mapping, error construction, and early-return on non-MySQL backends — is never exercised by any test in this module. If the real function regresses (e.g. due to the column-name bug above), these tests will still pass.

The tests should either:
1. Call `verify_innodb_engines` against a mock/test MySQL connection (integration test), or
2. Extract the error-building logic into a separate pure function that is both unit-tested and called from the real code path.

As written, the test for `non_innodb_table_is_flagged` also asserts against a hardcoded error string (line 482–484) that does not match the string produced by `verify_innodb_engines` (which says `'To fix, run in a maintenance window:'` with generated `ALTER TABLE` commands), so even the error-message assertion is testing dead code.

src/outbound/sql/migrations.rs
```diff
@@ -16,6 +16,99 @@ impl MigratorTrait for Migrator {
...
+        }
+    }
+
+    if non_innodb.is_empty() {
+        Ok(())
+    } else {
+        let alter_commands = non_innodb
+            .iter()
+            .map(|entry| {
+                let table = entry.split('(').next().unwrap_or(entry);
```
Christiantyemele: **Important — Fragile table-name extraction from a composed string**

`entry` is built as `format!("{}({})", row.table_name, row.engine)` (line 89), then immediately split on `'('` to recover the table name (line 99). This round-trip through a string is fragile: if a table name ever contained a `(` character the split would truncate it silently.

Consider keeping table name and engine as a structured pair instead:
```rust
let mut non_innodb: Vec<(\&str, String)> = Vec::new();  // (table_name, engine)
for row in \&rows {
    if !row.engine.eq_ignore_ascii_case("InnoDB") {
        // log...
        non_innodb.push((\&row.table_name, row.engine.clone()));
    }
}
// Then build error message directly from the pair.
```
The `unwrap_or` on line 99 is also technically dead code — `split` on any non-empty string always yields at least one element — but the redundancy goes away if the structured approach above is adopted.

src/outbound/sql/migrations.rs
```diff
@@ -16,6 +16,99 @@ impl MigratorTrait for Migrator {
...
+            .collect::<Vec<_>>()
+            .join(" ");
+        Err(DbErr::Custom(format!(
+            "Non-InnoDB tables detected: {}. \
+             The server requires InnoDB for transactional guarantees. \
+             To fix, run in a maintenance window: {alter_commands}",
+            non_innodb.join(", "),
+        )))
+    }
+}
```
Christiantyemele, line 128: **Suggestion — Missing blank line before module declaration**

There is no blank line separating the closing brace of `verify_innodb_engines` from `pub(crate) mod tables`. A single blank line here follows the project's existing style (all other top-level items in this file have one blank line between them) and aids readability.

```rust
    }
}

pub(crate) mod tables {
```

src/outbound/sql/mod.rs
```diff
@@ -6,7 +6,7 @@ mod models;
 mod store;
 
 pub use error::RepositoryError;
-pub use migrations::Migrator;
+pub use migrations::{Migrator, verify_innodb_engines};
```
Christiantyemele: **Suggestion — `verify_innodb_engines` could be `pub(crate)` instead of `pub`**

This function is a startup-only hook called from `setup.rs`. Re-exporting it as `pub` from `sql/mod.rs` makes it part of the crate's public surface, which is unnecessary. `pub(crate)` in `migrations.rs` and a correspondingly scoped re-export here would reflect its intended visibility and avoid accidental external misuse.

```rust
// In migrations.rs
pub(crate) async fn verify_innodb_engines(...) { ... }

// In mod.rs
pub(crate) use migrations::{Migrator, verify_innodb_engines};
```

Christiantyemele (changes requested):
## Review Summary

This PR introduces a well-motivated startup check to catch pre-existing MySQL deployments whose tables were created with a non-transactional engine (MyISAM/etc), which would silently undermine the transactional guarantees introduced in issue #244. The overall design — fail-fast at boot, log a clear runbook, avoid automatic ALTER TABLE rewrites — is sound. However, there are **three bugs** that together mean the check will never fire correctly on MySQL in practice, and one test suite that tests a stub implementation rather than the real function.

**Findings: 3 Bug, 1 Important, 2 Suggestion**

Key themes: (1) the `information_schema` query will produce empty rows on MySQL due to uppercase column-name mismatch in `FromQueryResult`, (2) the `credentials` table is InnoDB-pinned in migrations but excluded from the check, (3) the unit tests validate a re-implementation of the logic rather than calling `verify_innodb_engines` itself, providing no regression safety for the actual code path.

martcpp:
@ndefokou pls take you time and rework on this thank after you can tag e most of the ac are not met

Hermann-Core (approved):
@ndefokou verified InnoDB startup check with column aliasing, credentials table inclusion, structured pair evaluation, and backend documentation.

martcpp:
rebase @ndefokou

Christiantyemele (approved):

PR #299, opened 2026-07-30
270 coverage upload is non functional  publish html report as a ci artifact instead

### LINKED ISSUE #270: Coverage upload is non-functional -publish HTML report as a CI artifact instead
### Problem

The coverage job is configured to upload to Codecov, but the required token and keys were never set up. The upload silently does nothing, so the coverage data is stale and isn't published anywhere. Nobody can see coverage for a PR today.

### Proposed fix

Drop the dependency on external upload credentials. Have the coverage job emit an HTML report and publish it as a GitHub Actions artifact on the workflow run, so reviewers can download and open it directly from the PR checks.

Investigate whether the existing `cargo-coverage` step can produce HTML directly (`cargo llvm-cov --html`) or whether a separate report-generation step is needed.

### Acceptance criteria

- [ ] Coverage report is uploaded as a GitHub Actions artifact on every CI run
- [ ] The uploaded artifact is the HTML report, not `lcov`/`json`/any other intermediate format
- [ ] The artifact is downloadable and renders correctly when opened locally
- [ ] The dead Codecov upload step is removed or fixed- not left in place silently failing

1 Day

Blindspot22:
The coverage job is configured to upload to Codecov, but the required token and keys were never set up. The upload silently does nothing, so the coverage data is stale and isn't published anywhere. Nobody can see coverage for a PR today.


.github/workflows/CI.yml
```diff
@@ -343,19 +343,13 @@ jobs:
         uses: taiki-e/install-action@c44f6b046f1c29ae5918b1e0bfdbb2f1813836fd # v2
         with:
           tool: cargo-llvm-cov,nextest
```
martcpp:
```suggestion
          tool: cargo-llvm-cov@0.6.16,nextest@0.9.101 # or an version that is working now because of our sha pin 
```

martcpp (dismissed):

martcpp (approved):

Hermann-Core (approved):

PR #301, opened 2026-07-30
279 derived history feature to collapse repeated cfg gates

### LINKED ISSUE #279: Derived `history` feature to collapse repeated cfg gates
### Problem
 
The snapshot methods on the public `StatusListRepository` trait are gated on `#[cfg(any(feature = "server", "postgres", "sqlite", "mysql"))]`, and that same six-line attribute is copy-pasted ~20 times across `ports/mod.rs`, `application/mod.rs`, and `memory.rs`.
 
Two consequences. It's the largest readability tax in the store. And because `ports` is public (`lib.rs:17`), the *shape* of a public trait now depends on which features are enabled — an external implementor must replicate that cfg expression verbatim, and enabling `sqlite` changes the trait out from under them. That's contrary to Cargo's feature-additivity expectation.
 
### Proposed fix
 
```toml
history  = []
server   = [..., "history"]
postgres = [..., "history"]
sqlite   = [..., "history"]
mysql    = [..., "history"]
```
 
Every gate collapses to `#[cfg(feature = "history")]`.
 
**Decision to make here:** whether to drop the gate on the trait methods entirely. `StatusListSnapshot` is a plain data struct with no optional dependencies, so defining it unconditionally costs nothing and the public trait stops shape-shifting. That's a public-API call worth making deliberately rather than by default.
 

 
### Acceptance criteria
 
- [x] `history` feature added and derived from `server`, `postgres`, `sqlite`, `mysql`
- [x] Every repeated multi-feature gate replaced with `#[cfg(feature = "history")]`
- [x] Decision recorded on whether the public trait methods stay gated
- [x] `cargo check` passes for each feature individually and for `--all-features`

### References
 
- The Cargo Book — feature unification and additivity: https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

Blindspot22:
The snapshot methods on the public StatusListRepository trait are gated on #[cfg(any(feature = "server", "postgres", "sqlite", "mysql"))], and that same six-line attribute is copy-pasted ~20 times across ports/mod.rs, application/mod.rs, and memory.rs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,9 +6,11 @@ edition = "2024"
 [features]
 default = ["memory"]
 memory = []
-postgres = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-postgres"]
-sqlite = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-sqlite"]
-mysql = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-mysql"]
+history = []
+server = ["history"]
+postgres = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-postgres", "history"]
+sqlite = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-sqlite", "history"]
+mysql = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-mysql", "history"]
 redis = ["dep:redis"]
 aws = [
   "dep:aws-config",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,11 +6,11 @@ edition = "2024"
 [features]
 default = ["memory"]
 memory = []
-history = []
+history = ["dep:sea-orm", "dep:sea-orm-migration"]
 server = ["history"]
-postgres = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-postgres", "history"]
-sqlite = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-sqlite", "history"]
-mysql = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-mysql", "history"]
+postgres = ["sea-orm?/sqlx-postgres", "history"]
+sqlite = ["sea-orm?/sqlx-sqlite", "history"]
+mysql = ["sea-orm?/sqlx-mysql", "history"]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -6,9 +6,11 @@ edition = "2024"
 [features]
 default = ["memory"]
 memory = []
-postgres = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-postgres"]
-sqlite = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-sqlite"]
-mysql = ["dep:sea-orm", "dep:sea-orm-migration", "sea-orm?/sqlx-mysql"]
+history = []
```
Hermann-Core: The PR description mentions collapsing ~20 copy-pasted `#[cfg(any(...))]` gates across `src/outbound/mod.rs`, `src/test_utils.rs`, etc. However, currently only `src/setup.rs` and `Cargo.toml` were updated. Please complete the refactoring by replacing remaining repeated `cfg(any(feature = "sqlite", feature = "postgres", feature = "mysql"))` gates in `src/outbound/mod.rs` and `src/test_utils.rs` with `#[cfg(feature = "history")]`.

src/setup.rs
```diff
@@ -2,39 +2,19 @@
 
 #[cfg(feature = "aws")]
 use aws_config::{BehaviorVersion, Region};
-#[cfg(any(
-    feature = "acme",
-    feature = "sqlite",
-    feature = "postgres",
-    feature = "mysql"
-))]
+#[cfg(any(feature = "acme", feature = "history"))]
```
Hermann-Core: Gating these imports on `#[cfg(any(feature = "acme", feature = "history"))]` causes 4 unused import warnings (`Context`, `eyre`, `ExposeSecret`, `Duration`) when building with `--no-default-features --features server` because `history` is enabled by `server` without `acme` or any DB feature. These imports should be gated on the features that actually consume them.

Hermann-Core (changes requested):
@Blindspot22 Please review the inline comments regarding unused import warnings when building with `--no-default-features --features server`, completing the cfg gate collapse across `src/outbound/mod.rs` and `src/test_utils.rs`, and rebasing on `main`.

Blindspot22:
> @Blindspot22 Please review the inline comments regarding unused import warnings when building with `--no-default-features --features server`, completing the cfg gate collapse across `src/outbound/mod.rs` and `src/test_utils.rs`, and rebasing on `main`.

Okay on it

Hermann-Core (approved):

Hermann-Core (approved):

PR #302, opened 2026-07-30
278 unify snapshothistory naming

### LINKED ISSUE #278: Unify snapshot/history naming
### Problem
 
One concept, four names: the table is `status_list_history`, the port is `StatusListHistoryRepository`, the domain type is `StatusListSnapshot`, the struct field is `history`, the method says `snapshot`. A reader has to re-map constantly.
 
Worse, the `...WithHistory` types carry `without_history()` constructors (`state.rs:215-219`, and the `UpdateStatuses` equivalent) — a self-contradicting name, plus a three-`dyn`-parameter turbofish at the call site to construct a service that ignores one of them.
 
### Proposed fix
 
Pick one noun - "snapshot" reads better  and align the port, field, and method names to it. The table name can stay.
 
Rename `StatusListApplicationServiceWithHistory` to `StatusListApplicationService` with `history` as an internal `Option`. That removes both the contradiction and the turbofish.

 
### Blocked by
 
- **#269 ** - not a hard dependency, but a rename conflicts with everything in flight. Land it after the store work settles, or spend the time resolving conflicts.
### Acceptance criteria
 
- [x] One noun used across port, field, and method names
- [x] No type named `...WithHistory` carries a `without_history()` constructor
- [x] No turbofish required at the construction site
- [x] Diff contains identifier changes only  no behaviour change

Blindspot22:
One concept, four names: the table is status_list_history, the port is StatusListHistoryRepository, the domain type is StatusListSnapshot, the struct field is history, the method says snapshot. A reader has to re-map constantly.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -559,7 +559,8 @@ pub struct StatusListConfig {
     /// This prevents unbounded database growth and mitigates timing leak
     /// risks described in draft-21 §12.7. When disabled, historical resolution
     /// via `?time=` query parameter will not be available.
-    pub history_retention_secs: u64,
+    #[serde(alias = "history_retention_secs")]
+    pub snapshot_retention_secs: u64,
 }
 
 #[cfg(feature = "redis")]
@@ -740,7 +741,7 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("cache.max_capacity", 100)?
         .set_default("status_list.token_exp_secs", 900)?
         .set_default("status_list.token_ttl_secs", 300)?
-        .set_default("status_list.history_retention_secs", 7776000)?
+        .set_default("status_list.snapshot_retention_secs", 7776000)?
         .set_default("rate_limit.strict_burst_size", 10)?
         .set_default("rate_limit.strict_period_secs", 60)?
         .set_default("rate_limit.permissive_burst_size", 100)?
--- a/src/config.rs
+++ b/src/config.rs
@@ -551,14 +551,19 @@ pub struct CacheConfig {
 pub struct StatusListConfig {
     pub token_exp_secs: u64,
     pub token_ttl_secs: u64,
-    /// Retention period for historical status list snapshots in seconds.
+    /// Retention period for status list snapshots in seconds.
     /// Snapshots older than this will be deleted by a scheduled cleanup task.
     /// Default is 90 days (7776000 seconds).
     ///
-    /// **Privacy note:** Set to 0 to disable historical snapshots entirely.
+    /// **Privacy note:** Set to 0 to disable snapshots entirely.
     /// This prevents unbounded database growth and mitigates timing leak
     /// risks described in draft-21 §12.7. When disabled, historical resolution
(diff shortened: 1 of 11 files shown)


Hermann-Core (changes requested):
@Blindspot22 Clean rename and domain unification. `#[serde(alias = "history_retention_secs")]` correctly preserves configuration backwards compatibility. Please rebase on `main` to drop the outdated `.github/workflows/CI.yml` workflow diff.

src/domain/service.rs
```diff
@@ -245,10 +245,10 @@ impl Service {
 
     /// Purge historical snapshots older than `cutoff` timestamp.
     pub async fn cleanup_historical_snapshots(&self, cutoff: i64) -> Result<u64, StatusListError> {
```
ndefokou: `cleanup_historical_snapshots` (and `get_historical_snapshot` at line 233) still mix the "historical" noun with the newly unified "snapshot" noun used everywhere else. The ticket asks for *one noun across method names*. These two are the only public methods that still say "historical". If they're deliberately describing the historical feature (mirroring the retained `status_list_history` table) that's fine, but worth a one-line comment or a decision note so the naming rule is explicit. Suggest either renaming to `cleanup_snapshots`/`get_snapshot_at_time` or documenting why "historical" is intentionally kept.

src/config.rs
```diff
@@ -559,7 +559,8 @@ pub struct StatusListConfig {
     /// This prevents unbounded database growth and mitigates timing leak
     /// risks described in draft-21 §12.7. When disabled, historical resolution
     /// via `?time=` query parameter will not be available.
-    pub history_retention_secs: u64,
+    #[serde(alias = "history_retention_secs")]
```
ndefokou, line 625: The field rename to `snapshot_retention_secs` is good and the `#[serde(alias = "history_retention_secs")]` preserves backward compatibility for existing `APP_STATUS_LIST__HISTORY_RETENTION_SECS` env vars (confirmed the config is loaded via `try_deserialize`, so the serde alias is honored). One note: this means the config layer now accepts *two* names for the same setting, which slightly softens the "one noun" goal at the boundary. That's an acceptable trade-off for a non-breaking rename, but consider adding a deprecation note in the README/config docs so the old name is eventually dropped, otherwise the duplicate lingers indefinitely.

ndefokou (commented):
Reviewed against the acceptance criteria in #278. The port (`StatusListSnapshotRepo`), field (`snapshot_repo`), accessor (`snapshot_repo()`), and struct constructors (`SqlStatusListSnapshotRepo`, `MemoryStatusListSnapshotRepo`, `with_snapshot`) now use one noun. Confirmed no `...WithHistory` / `without_history` / construction-site turbofish remains, and the diff is identifier-only (the retained `StatusListHistoryRecord` / `status_list_history` table are intentionally unchanged per the ticket). A few nits below.

src/config.rs
Blindspot22: On it

ndefokou (approved):
MY changes have been addressed LGTM

Hermann-Core (approved):

PR #293, opened 2026-07-28
addindg all postgress and my sqlx database data

ndefokou:
This Pull Request addresses ticket **#273** by implementing explicit connection pool configuration and tuning parameters for Postgres and MySQL backends. 

Previously, `state.rs` constructed production database connections using `ConnectOptions::new(db_url)` with no options set. Unconfigured pools used sqlx defaults and queued requests indefinitely under connection contention. This PR exposes, documents, and configures all six key connection-pool properties to guarantee proper connection-shedding and prevent service hangs under heavy load.

---

### 1. Added Connection Pool Configuration Struct & Defaults
* **`src/config.rs`**: Introduced the `DatabasePoolConfig` struct. It maps environment variables to the configuration and includes documentation on how to properly size the pool relative to the server's `max_connections` and deployment replica count.
* **Defaults Configured**:
  * `max_connections`: `5`
  * `min_connections`: `1`
  * `acquire_timeout_secs`: `5`
  * `connect_timeout_secs`: `10`
  * `idle_timeout_secs`: `600` (10 minutes)
  * `max_lifetime_secs`: `1800` (30 minutes)

### 2. Applied Pool Tuning at Startup
* **`src/state.rs`**: Updated connection building logic to load and apply these options dynamically to `ConnectOptions`. For production safety, sqlx logging is disabled on these connections.
* SQLite connections remain configured to run with `max_connections(1)` and foreign keys enabled.

### 3. Exposed Environment Variables
* **`.env.template`**:Added the database connection pool section along with sizing documentation: $$\text{pool.max_connections} = \lfloor \frac{\text{server max_connections}}{\text{replica count}} \rfloor - \text{headroom (e.g., 5)}$$

### Automated Tests Added:
1. **Defaults and Overrides Validation**: Added assertions in `test_default_config` and created `test_pool_config_env_override` in `src/config.rs` to verify that environment variables (e.g. `APP_DATABASE__POOL__*`) take precedence over defaults.
2. **Pool Saturated Timeout Test**: Added `test_pool_acquire_timeout_fires` (gated under the `postgres-tests` feature) in `src/state.rs`. It locks the connection pool size to 1, holds the connection active, attempts a second execution, and asserts that an acquisition timeout error is returned within `acquire_timeout` seconds rather than queuing indefinitely.

### Test Run Status:
```bash
# Validates that all unit and integration tests compile and pass successfully
cargo test --features postgres-tests

# Ensures code conforms to all workspace formatting and linting rules
[...]

PR #304, opened 2026-07-30
277 store error mapping and contract assertions

### LINKED ISSUE #277: Store error mapping and contract assertions
### Problem
 
Five behaviour-inert defects found reviewing #244, all in the transactional store path.
 
| | Location | Issue |
|---|---|---|
| C1 | `store.rs:242-244` | Failed rollback on an empty guard-miss transaction propagates → a 409 becomes a 500. Nothing was written, so the rollback is a no-op |
| C4 | `store.rs:201-207` | No `snapshot.list_id == list_id` check. The signature permits atomically committing a snapshot against the wrong list |
| M4 | `store.rs:250` | `.exec(&txn)` emits `RETURNING "snapshot_id"` inside the lock window for a value the caller discards. Sibling at `:98` uses `exec_without_returning`; same issue at `:321` |
| — | `update_status.rs:82` | Logs "Invalidated cache" unconditionally, but `invalidate_after_commit` (`application/mod.rs:658-668`) deliberately swallows the failure. Will mislead whoever debugs a stale read |
| A1 | `application/mod.rs:260, 359, 564` | `history: Option<Arc<H>>` used as a boolean — the use case checks presence then delegates the write to `self.repository`. Nothing prevents the two pointing at different stores |
| — | `models.rs` | `status_list_history.status_list` lacks the `#[sea_orm(column_type = "Json")]` its siblings carry. Inference works today; the risk is a Postgres driver bump |
 
### Proposed fix
 
```rust
// C1 — nothing was written; Drop rolls back if the explicit call fails.
if result.rows_affected == 0 {
    if let Err(e) = txn.rollback().await {
        tracing::warn!(error = ?e, "rollback of empty conflict transaction failed");
    }
    return Ok(false);
}
```
 
C4: debug assertion or an early `RepositoryError`. M4: `exec_without_returning` at all three sites. Cache log: report the actual outcome. Models: add the annotation.
 
A1: replace `history: Option<Arc<H>>` with `snapshots_enabled: bool`, or drop the flag entirely and let the repository decide. **Settle this before opening the PR** so a design discussion doesn't block four one-line fixes.
 
### Blocked by
 
- **#269** — C1 sits inside the guarded-update function #269 rewrites into a shared `guarded_update<C: ConnectionTrait>`. Landing this first means rewriting the patch.
- **Coordinate with Issue #273**  also touches the guard-miss path. Mechanical conflict, resolvable, but don't run them in parallel.
### Acceptance criteria
 
- [ ] Failed rollback on an empty transaction no longer produces a 500
- [ ] `snapshot.list_id` validated against `list_id`
- [ ] All three insert sites use `exec_without_returning`
- [ ] Cache-invalidation log reflects actual outcome
- [ ] `status_list_history::Model.status_list` carries `#[sea_orm(column_type = "Json")]`
- [ ] A1 direction decided, recorded here, and applied

Blindspot22:
C4: debug assertion or an early RepositoryError. M4: exec_without_returning at all three sites. Cache log: report the actual outcome. Models: add the annotation.

A1: replace history: Option<Arc<H>> with snapshots_enabled: bool, or drop the flag entirely and let the repository decide. Settle this before opening the PR so a design discussion doesn't block four one-line fixes.

### DIFF
--- a/src/domain/service.rs
+++ b/src/domain/service.rs
@@ -79,6 +79,11 @@ impl Service {
         self.cert_provider.as_ref()
     }
 
+    /// Whether historical snapshot retention is enabled.
+    pub fn snapshots_enabled(&self) -> bool {
+        self.snapshot_repo.is_some()
+    }
+
     /// Create and publish a new status list record, enforcing uniqueness and size invariants.
     #[allow(clippy::too_many_arguments)]
     pub async fn publish_status_list(
@@ -119,36 +124,13 @@ impl Service {
             return Err(StatusListError::TooLarge);
         }
 
-        // No `find` pre-check: the primary key on `list_id` is the only check
-        // that is not a TOCTOU race, and every repository classifies the
-        // resulting violation as `AlreadyExists` — proven against SQLite, MySQL
-        // and Postgres by `assert_duplicate_list_id_is_conflict` (which covers
-        // both branches below), and by construction for the in-memory adapter.
-        // A pre-check would add a round trip to every publish and still have to
-        // defer to this path when a competing writer commits between the read
-        // and the insert.
-        //
-        // The trade is not free on the duplicate path: a rejected publish now
-        // costs BEGIN + failed INSERT + ROLLBACK instead of one SELECT, which on
-        // Postgres burns a transaction ID and leaves a dead tuple for autovacuum.
-        // That is the right way round — duplicates are the rare case, and the
-        // pre-check never actually prevented one.
-        //
-        // Only unique violations are classified. A foreign-key violation on
-        // `issuer` still surfaces as a generic backend error (500); that is
-        // unreachable in practice because authentication resolves the issuer's
-        // credential before this runs, so an unregistered issuer is rejected as
-        // 401 long before the insert.
-        match &self.snapshot_repo {
-            Some(_) => {
-                let snapshot = build_snapshot(&record, token_exp_secs);
-                self.status_list_repo
-                    .insert_with_snapshot(record.clone(), snapshot)
-                    .await?;
-            }
-            None => {
-                self.status_list_repo.insert(record.clone()).await?;
-            }
+        if self.snapshots_enabled() {
+            let snapshot = build_snapshot(&record, token_exp_secs);
+            self.status_list_repo
+                .insert_with_snapshot(record.clone(), snapshot)
+                .await?;
+        } else {
+            self.status_list_repo.insert(record.clone()).await?;
         }
         Ok(record)
     }
@@ -198,18 +180,15 @@ impl Service {
         let previous_updated_at = existing.updated_at;
         existing.updated_at = next_updated_at(previous_updated_at, current_unix_timestamp());
 
-        let landed = match &self.snapshot_repo {
-            Some(_) => {
-                let snapshot = build_snapshot(&existing, token_exp_secs);
-                self.status_list_repo
-                    .update_with_snapshot(existing.clone(), previous_updated_at, snapshot)
-                    .await?
-            }
-            None => {
-                self.status_list_repo
-                    .update(existing.clone(), previous_updated_at)
-                    .await?
-            }
+        let landed = if self.snapshots_enabled() {
+            let snapshot = build_snapshot(&existing, token_exp_secs);
+            self.status_list_repo
+                .update_with_snapshot(existing.clone(), previous_updated_at, snapshot)
+                .await?
+        } else {
+            self.status_list_repo
+                .update(existing.clone(), previous_updated_at)
+                .await?
         };
 
         if !landed {
@@ -342,12 +321,17 @@ fn build_snapshot(record: &StatusListRecord, token_exp_secs: u64) -> StatusListS
 }
 
 async fn invalidate_after_commit(cache: &dyn StatusListCache, list_id: &str) {
-    if let Err(error) = cache.invalidate(list_id).await {
-        tracing::warn!(
-            list_id = %list_id,
-            error = ?error,
-            "status list write committed, but cache invalidation failed; \
-             reads may be stale until the cache entry expires"
-        );
+    match cache.invalidate(list_id).await {
+        Ok(()) => {
+            tracing::debug!(list_id = %list_id, "invalidated cache entry after commit");
+        }
+        Err(error) => {
+            tracing::warn!(
+                list_id = %list_id,
+                error = ?error,
+                "status list write committed, but cache invalidation failed; \
+                 reads may be stale until the cache entry expires"
+            );
+        }
     }
 }
--- a/src/outbound/sql/models.rs
+++ b/src/outbound/sql/models.rs
@@ -105,6 +105,7 @@ pub(crate) mod status_list_history {
         pub snapshot_id: String,
         pub list_id: String,
         pub issuer: String,
+        #[sea_orm(column_type = "Json")]
         pub status_list: StatusList,
         pub sub: String,
         pub iat: i64,
--- a/src/outbound/sql/store.rs
+++ b/src/outbound/sql/store.rs
@@ -272,15 +272,15 @@ impl SeaOrmStore<StatusListRecord> {
             .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
 
         if result.rows_affected == 0 {
-            txn.rollback()
-                .await
-                .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+            if let Err(e) = txn.rollback().await {
+                tracing::warn!(error = ?e, "rollback of empty conflict transaction failed");
+            }
             return Ok(false);
         }
 
         let history_active: status_list_history::ActiveModel = snapshot.into();
         if let Err(insert_err) = status_list_history::Entity::insert(history_active)
-            .exec(&txn)
+            .exec_without_returning(&txn)
             .await
         {
             txn.rollback().await.map_err(|rollback_err| {
@@ -350,9 +350,9 @@ impl SeaOrmStore<StatusListHistoryRecord> {
     pub async fn insert_one(&self, entity: StatusListHistoryRecord) -> Result<(), RepositoryError> {
         let active: status_list_history::ActiveModel = entity.into();
         status_list_history::Entity::insert(active)
-            .exec(&*self.db)
+            .exec_without_returning(&*self.db)
             .await
-            .map_err(|e| RepositoryError::InsertError(e.to_string()))?;
+            .map_err(map_insert_err)?;
         Ok(())
     }
 
--- a/src/outbound/sql/store.rs
+++ b/src/outbound/sql/store.rs
@@ -59,10 +59,15 @@ impl SeaOrmStore<StatusListRecord> {
         entity: StatusListRecord,
         snapshot: StatusListHistoryRecord,
     ) -> Result<(), RepositoryError> {
-        // Captured before `entity` is consumed below; only the contention test
-        // reads it.
         #[cfg(test)]
         let probed_list_id = entity.list_id.clone();
+
+        if snapshot.list_id != entity.list_id {
+            return Err(RepositoryError::InsertError(format!(
+                "snapshot list_id ({}) does not match entity list_id ({})",
+                snapshot.list_id, entity.list_id
+            )));
+        }
 
         let txn = self
             .db
@@ -227,6 +232,13 @@ impl SeaOrmStore<StatusListRecord> {
         expected_updated_at: i64,
         snapshot: StatusListHistoryRecord,
     ) -> Result<bool, RepositoryError> {
+        if snapshot.list_id != list_id || entity.list_id != list_id {
+            return Err(RepositoryError::UpdateError(format!(
+                "snapshot list_id ({}) or entity list_id ({}) does not match list_id ({})",
+                snapshot.list_id, entity.list_id, list_id
+            )));
+        }
+
         if entity.updated_at <= expected_updated_at {
             return Err(RepositoryError::UpdateError(format!(
                 "guarded update requires a strictly newer updated_at \


Blindspot22:
Blocked by https://github.com/adorsys/status-list-server/pull/293

ndefokou (commented):
## Senior code review — PR #304 (277 store error mapping and contract assertions)

I reviewed the transactional store path, the service orchestration, the model annotation, and the new tests against the issue's four defects (C1, C4, M4, cache-log) and the A1 design decision. Overall this is a correct and conservative set of behaviour-inert fixes; C1, C4, M4, the cache log, and the model annotation all satisfy their acceptance criteria. The main point that needs explicit resolution is A1 (which the issue flagged as "settled before opening the PR" and this PR does not actually settle).

### 1. High-level architectural review
The PR does the right thing by pushing contract validation (C4) down to the repository boundary and hoisting it out of the lock window, and by removing the RETURNING round-trips (M4) that were wasting work inside the transaction. The cache-invalidation logging and the `Json` column annotation are low-risk, high-value hygiene fixes.

The A1 item is where the architecture is left open. The issue offered two concrete options — (a) replace `history: Option<Arc<H>>` with `snapshots_enabled: bool`, or (b) drop the flag entirely and let the repository decide. This PR does neither: it keeps `history_repo: Option<Arc<dyn StatusListHistoryRepo>>` and adds a derived `snapshots_enabled()` accessor. This is *better* than a raw boolean in one sense (there is no separate mutable flag to drift), but it is not option (b): the `StatusListRepo` trait still exposes both `insert`/`update` and `insert_with_snapshot`/`update_with_snapshot`, so the *repository cannot decide* whether snapshots are written — the service must, via `snapshots_enabled()`. The "two pointers to different stores" coupling the issue worried about is still expressible in `setup.rs`: `SqlStatusListRepo` and the `SqlStatusListHistoryRepo` are always constructed (SQL backends), and the *only* thing gating snapshot writes is `history_retention_secs != 0`. The service's decision and any future repository-level decision need a single owner. Please record the A1 decision in the PR (or a linked issue) as the acceptance criteria required, and consider draining the flag into the repository contract so a `StatusListRepo` "knows" whether it snapshots.

### 2. Correctness review
- **C1** — correct. When `rows_affected == 0` the guard failed, nothing was written, downgrading a failed explicit rollback to a `warn!` instead of escalating to a 500 is exactly right. sea-orm 1.1.20's `DatabaseTransaction` rolls back on `Drop` if left open, so even a failed explicit `rollback()` is backstopped. `Ok(false)` → `StatusListError::Conflict` → 409 is restored. Good.
- **C4** — correct and correctly placed *before* `txn.begin()`, so the invalid snapshot never occupies the lock window. Both `insert_one_with_snapshot` and `update_one_with_snapshot` are guarded. Note the error-variant asymmetry (`InsertError` vs `UpdateError`) is harmless because both map to `StatusListError::Backend` → 500, which is the right signal for a caller-contract violation (it is a programming error, not a client error).
- **M4** — both remaining `.exec(&txn)`/`.exec(&*self.db)` history-insert sites are now `exec_without_returning`, and `snapshot_id` is always populated (`auto_increment = false`, value supplied via `From<StatusListSnapshot>`), so the conversion is valid. The credentials/status-list inserts were already correct.
- **models.rs JSON annotation** — correct; `status_list_history.status_list` now matches its `status_lists` sibling and the migration's `.json()` column, and `StatusList` derives `FromJsonQueryResult`. Backward compatible; removes the Postgres-driver-bump inference risk.

### 3. Performance / scalability review
- M4 is the meaningful win: the history inserts inside the lock window no longer request RETURNING of `snapshot_id` the caller discards. Small but real per-write overhead removed from the critical transaction.
- C4's early validation adds a trivial string comparison before `begin()`; negligible.
- No new N+1, no new locking. The contention test path is unchanged.

### 4. API contract review
No public API / OpenAPI change. Inward contract (trait) semantics are tightened: callers can no longer silently write a snapshot for the wrong list. That is a good hardening. One *observable* behavioural change snuck in that is not mentioned in the issue: `SeaOrmStore<StatusListHistoryRecord>::insert_one` now uses `map_insert_err`, so a unique-key violation on `snapshot_id` returns `RepositoryError::DuplicateEntry` → `StatusListError::AlreadyExists` → **409** instead of the previous `InsertError` → 500. That is arguably an improvement, but it is inconsistent with the identical constraint check at the two in-transaction history insert sites (`insert_one_with_snapshot` :83 and `update_with_snapshot` :211), which still return a raw `InsertError` for the same violation. Please make the treatment uniform and call it out in the changelog.

### 5. Security / privacy review
None of the changes widen the attack surface. The `list_id` cross-check prevents a snapshot being attributed to the wrong status list, which is a data-integrity (and thus indirectly a correctness/privacy) improvement — a mis-attributed but "valid" historical snapshot could otherwise be served by `find_valid_at`. Error strings echo caller-supplied identifiers (list_id) back into `RepositoryError` → 500 messages; those surface to operators (warn/error logs), not to clients directly, so acceptable, but keep them out of any future API-level error body.

### 6. Testing review
- C4 has two new focused tests in the SQLite path (`test_insert_one_with_snapshot_list_id_mismatch`, `test_update_one_with_snapshot_list_id_mismatch`) — good, deterministic, assert the exact error variant.
- **Gap: C1 has no test.** The issue's headline defect (a 409 becoming a 500 on a failed empty-transaction rollback) is exactly the kind of regression a test should lock down. A test that makes the guard fail with `rows_affected == 0` and asserts `Ok(false)` (via `update_one_with_snapshot`) on both SQLite and the MySQL multi-connection harness would pin C1. Recommend adding it.
- **No test for the M4 no-returning behaviour.** Hard to assert directly, but worth verifying the `exec_without_returning` switch on SQLite/Postgres/MySQL didn't drop needed columns (the PK is supplied, so it should be fine) — a compile-time + integration smoke covers this.
- No test for the new `snapshots_enabled()` accessor or the `map_insert_err` history path. The accessor is trivially covered by existing wiring, but the `DuplicateEntry` behavioural change should have a test if the decision is to keep it.

### 7. Maintainability review
- `snapshots_enabled()` is well-named and documented. Good.
- The `map_insert_err` asymmetry (see §4) is a genuine maintainability wart; unify the duplicate-handling in the two transaction paths or drop `map_insert_err` here.
- The A1 design smell (service decides, repository can't) is the largest maintainability/coupling debt and should be resolved deliberately, not implicitly.

### 8. Edge cases / regression risks
- The `entity.list_id != list_id` clause in `update_one_with_snapshot` is slightly redundant with the snapshot check but harmless and defensive; keep it.
- C4 returns before `txn.begin()`, so no connection is leaked on the early path — good.
- If a future refactor makes `history_retention_secs` decoupled from repository capability (e.g. an in-memory repo with history but retention=0, or a SQL deploy where retention is on but the history table is gone), the service flag and repository behaviour silently disagree. The memory backend's `require_history()` already errors on divergence; the SQL backend does not. This is the concrete expression of A1's "pointing at different stores" risk.

### 9. Blocking issues
- **A1 is not settled.** The acceptance criteria explicitly require the A1 direction to be "decided, recorded here, and applied". This PR applies a third variant (derived accessor) that is different from both proposed options and does not record the decision. This must be resolved/recorded before merge, even though the four mechanical fixes are ready. (Not a code-correctness blocker, but a release-process/acceptance blocker.)

### 10. Non-blocking suggestions
1. Add a C1 regression test (409-not-500 on failed empty guard rollback).
2. Add a test for the history `insert_one` unique-violation mapping (409 vs 500) if you keep `map_insert_err`; otherwise revert it or apply it uniformly.
3. Unify duplicate-key handling across the three history insert sites.
4. Consider asserting the C4 invariant with a `debug_assert!` in addition to — or instead of — the early `RepositoryError`, so misuse is caught in test/dev builds too (the issue itself suggested this).
5. Record the A1 decision in the PR description and/or a linked issue so the "let the repository decide" refactor is tracked, even if deferred.

### 11. Final assessment
The four behaviour-inert fixes (C1, C4, M4, cache-log) and the model annotation are correct, minimal, and satisfy their acceptance criteria; the tests added for C4 are solid. The only real gap relative to the issue is A1: the PR neither adopts the proposed design nor records the decision as required. I'd approve the mechanical fixes once a C1 test is added and the A1 decision is explicitly recorded; the duplicate-key-mapping asymmetry is worth a quick decision but is non-blocking.

Inline comments below correspond to the specific findings.

src/outbound/sql/store.rs
```diff
@@ -187,15 +201,15 @@ impl SeaOrmStore<StatusListRecord> {
             .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
 
         if result.rows_affected == 0 {
-            txn.rollback()
-                .await
-                .map_err(|e| RepositoryError::UpdateError(e.to_string()))?;
+            if let Err(e) = txn.rollback().await {
```
ndefokou, line 275: C1: correct. `rows_affected == 0` means the optimistic guard failed and nothing was written, so a failed explicit rollback must not escalate to a 500. sea-orm 1.1.20's `DatabaseTransaction` also rolls back on `Drop`, so the `warn!` here is a best-effort with a safety net. This restores the 409 (`Ok(false)` → `StatusListError::Conflict`). Suggest adding a regression test that forces `rows_affected == 0` and asserts `Ok(false)` — currently there is no test covering this exact path.

src/outbound/sql/store.rs
```diff
@@ -154,6 +161,13 @@ impl SeaOrmStore<StatusListRecord> {
         expected_updated_at: i64,
         snapshot: StatusListHistoryRecord,
     ) -> Result<bool, RepositoryError> {
+        if snapshot.list_id != list_id || entity.list_id != list_id {
```
ndefokou, line 235: C4: correct and correctly placed before `txn.begin()`, so a bad snapshot never occupies the lock window. The `InsertError` vs `UpdateError` variant asymmetry is harmless since both map to `StatusListError::Backend` → 500 (a caller-contract violation). Consider also a `debug_assert!` so test/dev builds catch the invariant even when a caller bypasses the early return.

src/outbound/sql/store.rs
```diff
@@ -187,15 +201,15 @@ impl SeaOrmStore<StatusListRecord> {
...
+            if let Err(e) = txn.rollback().await {
+                tracing::warn!(error = ?e, "rollback of empty conflict transaction failed");
+            }
             return Ok(false);
         }
 
         let history_active: status_list_history::ActiveModel = snapshot.into();
         if let Err(insert_err) = status_list_history::Entity::insert(history_active)
-            .exec(&txn)
+            .exec_without_returning(&txn)
```
ndefokou, line 283: M4: good — the history insert inside the lock window no longer requests RETURNING `snapshot_id` that the caller discards. `snapshot_id` is always supplied (`auto_increment = false`, populated via `From<StatusListSnapshot>`), so `exec_without_returning` is valid.

src/outbound/sql/store.rs
```diff
@@ -259,9 +273,9 @@ impl SeaOrmStore<StatusListHistoryRecord> {
     pub async fn insert_one(&self, entity: StatusListHistoryRecord) -> Result<(), RepositoryError> {
         let active: status_list_history::ActiveModel = entity.into();
         status_list_history::Entity::insert(active)
-            .exec(&*self.db)
+            .exec_without_returning(&*self.db)
             .await
-            .map_err(|e| RepositoryError::InsertError(e.to_string()))?;
+            .map_err(map_insert_err)?;
```
ndefokou, line 355: Behavioural change not mentioned in the issue: this site now uses `map_insert_err`, so a unique-key violation on `snapshot_id` returns `RepositoryError::DuplicateEntry` → `StatusListError::AlreadyExists` → HTTP 409 instead of the previous 500. Argubly an improvement, but inconsistent with the two in-transaction history inserts in `insert_one_with_snapshot` (:83) and `update_one_with_snapshot` (:211) which still return a raw `InsertError` for the same violation. Please unify the duplicate-key handling across the three history insert sites and call the behaviour change out.

src/domain/service.rs
```diff
@@ -79,6 +79,11 @@ impl Service {
         self.cert_provider.as_ref()
     }
 
+    /// Whether historical snapshot retention is enabled.
+    pub fn snapshots_enabled(&self) -> bool {
```
ndefokou, line 83: A1: this derives `snapshots_enabled` from `history_repo.is_some()`, which avoids a separate mutable boolean drifting — good. But it does not implement either option the issue proposed: it isn't a `snapshots_enabled: bool` field, and it is NOT 'let the repository decide' (the `StatusListRepo` trait still exposes both `insert`/`update` and `insert_with_snapshot`/`update_with_snapshot`, so the repository cannot decide — the service must). The coupling the issue worried about still exists in `setup.rs`: for SQL backends both repos are always constructed and the only gating is `history_retention_secs != 0`. The A1 direction is still not 'decided, recorded, and applied' as the acceptance criteria require — please record the decision in the PR/linked issue, even if the 'let the repository decide' refactor is deferred.

src/domain/service.rs
```diff
@@ -322,12 +321,17 @@ fn build_snapshot(record: &StatusListRecord, token_exp_secs: u64) -> StatusListS
...
 
 async fn invalidate_after_commit(cache: &dyn StatusListCache, list_id: &str) {
-    if let Err(error) = cache.invalidate(list_id).await {
-        tracing::warn!(
-            list_id = %list_id,
-            error = ?error,
-            "status list write committed, but cache invalidation failed; \
-             reads may be stale until the cache entry expires"
-        );
+    match cache.invalidate(list_id).await {
```
ndefokou, line 324: Cache-log fix: correct — the code now reports the actual outcome (debug on success, warn on failure) instead of unconditionally logging 'Invalidated cache' when `invalidate_after_commit` deliberately swallows the failure. This resolves the misleading-log defect.

src/outbound/sql/models.rs
```diff
@@ -105,6 +105,7 @@ pub(crate) mod status_list_history {
         pub snapshot_id: String,
         pub list_id: String,
         pub issuer: String,
+        #[sea_orm(column_type = "Json")]
```
ndefokou, line 108: Correct: `status_list_history.status_list` now carries `#[sea_orm(column_type = "Json")]`, matching its `status_lists` sibling and the migration's `.json()` column. `StatusList` derives `FromJsonQueryResult`, so the annotation is valid. Removes the Postgres-driver-bump inference risk the issue flagged. Backward compatible.

ndefokou (commented):
Inline findings on the store/service/models changes. These complement the full review summary posted above.

Hermann-Core (approved):
@Blindspot22 optimistic guard rollback error recovery, pre-transaction list_id contract assertions, exec_without_returning optimization, and model Json annotation verified. Can you fix the conflicts on your branch so we can merge?

Blindspot22:
@ndefokou Please check this again

martcpp (approved):

Hermann-Core (approved):
@Blindspot22  Verified rebase on main with clean conflict resolution. Optimistic guard rollback recovery, pre-transaction list_id assertions, and exec_without_returning optimizations remain verified.

PR #322, opened 2026-07-31
test: add optimistic locking conflict resolution tests for update_one  method

### LINKED ISSUE #274: No test covers retry-after-conflict or transaction shape
### Problem
 
Two gaps remain on main after #244, neither in scope for #267.
 
Retry-after-conflict has zero tests. The client contract is "409 - re-read - retry," and nothing asserts the retry succeeds - a guard that becomes permanently unmatchable would turn every conflict into an infinite client loop, invisibly.
 
Transaction shape is unasserted on default features. `cargo test` compiles none of the atomicity tests (defaults are `server` - `postgres`; the tests are gated on `sqlite`/`mysql`/`postgres-tests`), and `into_transaction_log` appears nowhere in the repo. Nothing prevents a refactor of `&txn` back to `&*self.db`.
 
Minor: the MySQL and Postgres atomicity tests never call `find_valid_at`, so the JSON round-trip is asserted on sqlite only.
 
### Scope boundary
 
#267 covers two-connection contention and is being worked in #266. This does not duplicate it, and nothing here needs a container.
 
### Proposed fix
 
- Retry test: B loses the guard, re-reads `updated_at`, retries, asserts success and final state.
- `MockDatabase` test asserting `BEGIN → UPDATE → INSERT → COMMIT` and `BEGIN → UPDATE → ROLLBACK` via `into_transaction_log()`. Default features, no Docker.
- One `find_valid_at` assertion in the MySQL and Postgres atomicity tests.

### Acceptance criteria
 
- [ ] Retry test exists and fails if the guard becomes unmatchable
- [ ] Transaction-shape test asserts both commit and rollback sequences
- [ ] Transaction-shape test runs under a plain `cargo test`
- [ ] MySQL and Postgres tests assert a snapshot read-back

1 day

Ngha-Boris:
### Problem
 
Two gaps remain on main after #244, neither in scope for #267.
 
Retry-after-conflict has zero tests. The client contract is "409 - re-read - retry," and nothing asserts the retry succeeds - a guard that becomes permanently unmatchable would turn every conflict into an infinite client loop, invisibly.
 
Transaction shape is unasserted on default features. `cargo test` compiles none of the atomicity tests (defaults are `server` - `postgres`; the tests are gated on `sqlite`/`mysql`/`postgres-tests`), and `into_transaction_log` appears nowhere in the repo. Nothing prevents a refactor of `&txn` back to `&*self.db`.
 
Minor: the MySQL and Postgres atomicity tests never call `find_valid_at`, so the JSON round-trip is asserted on sqlite only.
 
### Scope boundary
 
#267 covers two-connection contention and is being worked in #266. This does not duplicate it, and nothing here needs a container.
 
### Proposed fix
 
- Retry test: B loses the guard, re-reads `updated_at`, retries, asserts success and final state.
- `MockDatabase` test asserting `BEGIN → UPDATE → INSERT → COMMIT` and `BEGIN → UPDATE → ROLLBACK` via `into_transaction_log()`. Default features, no Docker.
- One `find_valid_at` assertion in the MySQL and Postgres atomicity tests.

### Acceptance criteria
 
- [x] Retry test exists and fails if the guard becomes unmatchable
- [x] Transaction-shape test asserts both commit and rollback sequences
- [x] Transaction-shape test runs under a plain `cargo test`
- [x] MySQL and Postgres tests assert a snapshot read-back


Christiantyemele (dismissed):
.

src/outbound/mod.rs
```diff
@@ -6,5 +6,5 @@ pub mod cert;
 pub mod memory;
 #[cfg(feature = "redis")]
 pub mod redis;
-#[cfg(feature = "history")]
+#[cfg(any(feature = "history", test))]
```
martcpp, line 9: The `test` arm compiles this module under default features (`memory`), where
 `history` is off and the optional `dep:sea-orm` is inactive — `sea_orm` and
`sea_orm_migration` come from `[dev-dependencies]` instead. That keeps the
 transaction-shape test runnable under a plain `cargo test`; `cargo build`
still never links sea-orm. Note the dev-dep resolves sea-orm's *default*
features, a superset of production's, so the module is never compiled under
production's exact feature set.

Cargo.toml
```diff
@@ -174,6 +174,9 @@ features = ["alloc", "ecdsa", "pem", "pkcs8"]
 sealed_test = "1.1.0"
 wiremock = "0.6"
 sea-orm = { version = "1.1", features = ["mock"] }
+sea-orm-migration = { version = "1.1.20", default-features = false, features = [
```
martcpp, line 177: Mirrors the optional dependency above. Needed because the `test` arm on
 `outbound::sql` (see src/outbound/mod.rs) compiles the module without the
 `history` feature: this supplies `MigratorTrait` and propagates
 `sea-orm/runtime-tokio-rustls`. Keep the version pin in sync with the
 optional dep.

src/outbound/sql/store.rs
```diff
@@ -1280,6 +1358,138 @@ mod test {
         assert!(matches!(backwards, Err(RepositoryError::UpdateError(_))));
     }
 
+    #[tokio::test]
+    async fn test_update_one_with_snapshot_transaction_log_shape() {
```
martcpp, line 1362: Pins the statement *shape* of `update_one_with_snapshot` on default features: `BEGIN → UPDATE → INSERT → COMMIT` on the guarded path, and `BEGIN → UPDATE → ROLLBACK` on the guard miss.

 Scope, deliberately: `MockDatabase` keeps a single connection-wide transaction slot and routes every statement into it while one is open, so  this cannot distinguish `.exec(&txn)` from `.exec(&*self.db)`. Transactional *membership* is proven behaviourally by  `test_sqlite_update_with_snapshot_is_atomic` (a failing snapshot INSERT must  roll the row UPDATE back), which needs `--features sqlite`.  `begin`/`commit`/`rollback` synthesize their markers without calling  `execute`, so they consume no `MockExecResult` — only the UPDATE and INSERT do. The expected SQL is rendered by sea-query and is version-sensitive: a mismatch after a dependency bump is a re-copy, not a regression.

src/outbound/sql/store.rs
```diff
@@ -1280,6 +1358,138 @@ mod test {
...
+
+    #[cfg(feature = "sqlite")]
+    const TEST_EC_JWK: &str = r#"{
+        "kty": "EC",
+        "crv": "P-256",
+        "x": "NeyFv_2L67OEplNbJpR02IFis4_lFW9HYmhfF5Or6m8",
+        "y": "eAH2qe8Pg3GQ28uxA8-qNAqdwQ_zfV2uKAvJ2sLpY9M"
+    }"#;
+
     /// failure rollback (no partial snapshot), and the conflict path — against
```
martcpp, line 1493: Proves the transaction boundary end to end: the commit path, the snapshot failure rollback (no partial snapshot), and the conflict path  against real SQLite, since `MockDatabase` cannot model rollback.

martcpp (changes requested):
Collapse the ten inlined P-256 JWK literals in store.rs onto the shared fixture.

src/outbound/sql/mod.rs
martcpp, line 1: `exec_without_returning` to match `insert_one_with_snapshot`: theInsertResult is discarded, and on Postgres `RETURNING snapshot_id` costs a round-tripped column inside the window where the guarded UPDATE holds its exclusive row lock.

martcpp (changes requested):

Hermann-Core (changes requested):
@Ngha-Boris please address @martcpp's review by collapsing the inlined P-256 JWK literals in src/outbound/sql/store.rs onto the shared TEST_EC_PUBLIC_JWK fixture.

martcpp (dismissed):

src/outbound/sql/store.rs
Hermann-Core: Appreciate the end-to-end transaction test. One consistency note against the "collapse inlined literals onto the shared fixture" changes-requested: this test still introduces a brand-new inlined P-256 JWK literal — `const TEST_EC_JWK` at line 1486 — which is byte-for-byte the same **public** key as `crate::test_fixtures::TEST_EC_PUBLIC_JWK` (same `kty/crv/x/y`, no private `d`). The collapse is therefore only partial: the JWT readers at lines 1503/1681/1871/2629 now use the shared fixture, but `test_update_one_conflict_loser_can_reread_and_retry` (line 1156) still parses `TEST_EC_JWK`, and the new local const re-introduces the duplication. Since the constant is public-key-only, simplest is to drop the literal and reference the shared fixture directly — e.g. `const TEST_EC_JWK: &str = crate::test_fixtures::TEST_EC_PUBLIC_JWK;` (or use it inline) — leaving a single source of truth.

Hermann-Core:
@Ngha-Boris one more comment left here https://github.com/adorsys/status-list-server/pull/322#discussion_r3756507183.

Hermann-Core:
@martcpp Comments have been posted on PR #322 for your attention — please review the thread on the end-to-end transaction test regarding the duplicated TEST_EC_JWK literal, and the accompanying notes. Thanks!

Ngha-Boris:
> @martcpp Comments have been posted on PR #322 for your attention — please review the thread on the end-to-end transaction test regarding the duplicated TEST_EC_JWK literal, and the accompanying notes. Thanks!

Done @Hermann-Core please recheck

src/outbound/sql/store.rs
```diff
@@ -591,10 +589,9 @@ mod test {
     use super::*;
     use crate::outbound::sql::models::StatusList;
     use jsonwebtoken::jwk::Jwk;
-    use sea_orm::{DatabaseBackend, MockDatabase, MockExecResult};
-    // `Migrator::up` is only called from `sqlite_connection` below; the MySQL and
-    // Postgres fixtures run their own migrations inside `test_containers`.
-    #[cfg(feature = "sqlite")]
+    use sea_orm::{DatabaseBackend, MockDatabase, MockExecResult, Statement, Transaction};
+    // `Migrator::up` is only called from the real-backend helpers below.
+    #[cfg(any(feature = "sqlite", feature = "mysql", feature = "postgres-tests"))]
```
Hermann-Core: The `sea_orm_migration` library is already brought in when any of the SQL database features are enabled. Do we really need to manually enable the feature flags again in this internal module, which is already gated by those same features? If not, could you align this with the existing feature-gating pattern? Otherwise, the feature configuration becomes unnecessarily duplicated and inconsistent with the module's gating logic.

src/outbound/sql/store.rs
Hermann-Core, line 1: This file is getting too large. Could we consider refactoring both the implementation and the tests to consolidate repetitive logic and logically duplicated test cases? The current size makes the file harder to navigate and maintain, and I don't think this implementation needs to be this large. We could track that in a follow-up ticket.

src/outbound/mod.rs
Hermann-Core: This should normally be available in tests when the appropriate feature (`history` in this case) is enabled as needed.

Cargo.toml
```diff
@@ -174,6 +174,9 @@ features = ["alloc", "ecdsa", "pem", "pkcs8"]
...
+  "runtime-tokio-rustls",
+] }
```
Hermann-Core: We already have these two dependencies declared at the project level, so they are also available in test dependencies too. We don't need to maintain separate dependency declarations for the test and normal paths.

Hermann-Core (changes requested):

src/outbound/sql/store.rs
Ngha-Boris: You are right @Hermann-Core, I will create a follow up ticket for this and work on it.

src/outbound/mod.rs
Ngha-Boris: The test arm is required because default = ["memory"], so plain cargo test doesn't enable history. Without it, the whole sql module (and the ticket's transaction-shape/retry/atomicity tests) wouldn't compile under default features directly failing the ticket's acceptance criterion. Gating on history alone would only run them with an explicit --features.

Cargo.toml
Ngha-Boris: The dev-dependency is load-bearing, not redundant. The project-level sea-orm-migration is optional = true and only activates via history (off by default), so it isn't linked into the default test build. The dev-dependency is what makes the crate available so migrations.rs/sql compiles for tests under plain cargo test.

src/outbound/sql/store.rs
Hermann-Core: Can you confirm that the ticket have been created already?

src/outbound/mod.rs
Hermann-Core: That should not be the intended behavior. Implementations gated behind a feature should remain within the corresponding feature context. Could you adjust this so that the feature boundary is properly respected?

Cargo.toml
Hermann-Core: That should not be the intended behavior. Implementations gated behind a feature should remain within the corresponding feature context. Could you adjust this so that the feature boundary is properly respected? We should not the same dep in both dependencies and dev-dependencies sections.

Hermann-Core:
@Ngha-Boris can you check the comments I left ?

src/outbound/mod.rs
Ngha-Boris: Okay I will do that.

src/outbound/sql/store.rs
Ngha-Boris: Yes when I do that I will add the link here

Ngha-Boris:
@Hermann-Core please check again all your comments were addressed

src/outbound/sql/store.rs
Ngha-Boris: Here is the follow up ticket: https://github.com/adorsys/status-list-server/issues/358

src/outbound/sql/store.rs
```diff
@@ -1292,6 +1371,133 @@ mod test {
         assert!(matches!(backwards, Err(RepositoryError::UpdateError(_))));
     }
 
+    #[tokio::test]
+    async fn test_update_one_with_snapshot_transaction_log_shape() {
```
Christiantyemele, line 1375: **Important:** This test is intended to satisfy the PR acceptance criterion that transaction-shape coverage runs under a plain `cargo test`, but it is currently unreachable in the default-feature build. The latest head has `src/outbound/mod.rs` back to `#[cfg(feature = "history")] pub mod sql;`, while the crate defaults to `memory`, so `cargo test --package status-list-server` never compiles `outbound::sql` and the filtered check confirms it: `cargo test --package status-list-server test_update_one_with_snapshot_transaction_log_shape` reports `running 0 tests`. That means the test can silently stop guarding the default CI path this PR is specifically trying to protect.

One way to make the guard real again is to expose the SQL module for tests and provide the test-only dependency needed by migrations, or otherwise move this mock transaction-shape assertion into a default-feature test module that is actually compiled:

```rust
#[cfg(any(feature = "history", test))]
pub mod sql;
```

```toml
[dev-dependencies]
sea-orm-migration = { version = "1.1.20", default-features = false, features = ["runtime-tokio-rustls"] }
```

Christiantyemele (changes requested):
.

Christiantyemele (approved):

Christiantyemele (approved):
PR #322 adds three test categories addressing the stated acceptance criteria: (1) retry-after-conflict test verifying the HTTP 409/re-read/retry contract, (2) transaction-shape test asserting BEGIN/UPDATE/INSERT/COMMIT and BEGIN/UPDATE/ROLLBACK via MockDatabase, and (3) find_valid_at assertions in MySQL and Postgres container tests for JSON round-trip verification. All tests compile under default features (cargo test) and follow project conventions. Prior feedback from @martcpp and @Hermann-Core on JWK literal cleanup is addressed: the new test uses the shared TEST_EC_PUBLIC_JWK fixture directly. No correctness, security, or spec-compliance issues found. cargo check passes.

Hermann-Core (approved):

martcpp (approved):

Issue #248, opened 2026-07-21
Replace imperative `kubectl set image` deploy with GitOps/Helm-based deployment


Hermann-Core:
## Problem

`.github/workflows/deploy.yml` builds and pushes a multi-arch image to GHCR, then deploys with a raw `kubectl set image deployment/statuslist-status-list-server-deployment status-list-server=ghcr.io/adorsys/status-list-server:sha-<short_sha> -n statuslist`. This bypasses the Helm chart (`helm/chart/`) entirely at deploy time. Concretely, today:

- Trigger is broader than it looks. The workflow triggers on `push: branches: [main]` and on `v*.*.*` tags — i.e. it deploys to the live EKS cluster (`datev-wallet-cluster`) on every merge to `main`, not only on tagged releases. Tag pushes don't change the deploy behavior at all: the deployed image is always the short-sha tag, never the semver or `latest` tag also built by the same job. Any GitOps/Helm migration needs to explicitly decide/preserve or change this "every merge to main = prod deploy" behavior, and (per companion ticket #247) decide how a real semver release tag should relate to what gets deployed.
- OIDC is already half-wired but unused. The `deploy-to-eks` job already declares `permissions: { id-token: write }`, but then authenticates with `aws-actions/configure-aws-credentials@v6` using static `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` secrets — the OIDC permission is requested and never used. This is the exact problem tracked by #236; this ticket is blocked by it.
- The chart itself is not deploy-ready yet, in a very specific way. `helm/chart/templates/deployment.yaml` currently renders no `securityContext` and does not reference `.Values.statuslist.resources` anywhere — even though `values.yaml` defines `statuslist.resources.requests`/`limits`, the template silently drops them, so today the chart wouldn't even enforce the resource limits an operator thinks they set. There are also no liveness/readiness probes and no `NetworkPolicy` template. This is tracked by #239; this ticket is blocked by it, and should not be started until #239's template fixes land, otherwise "helm upgrade --install" will just codify the same gaps `kubectl set image` currently hides.
[...]

PR #330, opened 2026-08-04
Automate versioning & changelog generation (release-plz-style release PRs)

### LINKED ISSUE #247: Automate versioning & changelog generation (release-plz-style release PRs)
## Problem

Versioning today is manual and, worse, split across two independent, uncoordinated workflows that both touch the version number without ever talking to each other:

1. **`.github/workflows/bump_version.yml`** — human-triggered (`workflow_dispatch`/`workflow_call`) with a manually-typed version string input. Runs `cargo-bump <version>` and commits the result directly to whatever branch triggered it. No PR, no review step, no changelog file is touched at all, and no git tag is created.
2. **`.github/workflows/semver_checks.yml`** — also human-triggered (`workflow_dispatch`/`workflow_call`). Independently computes a baseline from existing git tags and also runs `cargo bump` internally (purely to build a comparison artifact for `cargo-semver-checks`), separate from and uncoordinated with (1).

Because these are two separate manual entry points, a human has to supply the same version number twice, in two different workflow runs, with nothing enforcing they agree — a real, concrete source of drift (e.g. `bump_version.yml` bumps to `0.5.0` but `semver_checks.yml` is run later against a stale/different baseline).

On top of the duplication, the current process has no generated changelog, no PR-based review/audit trail before a version bump lands, and critically no connection to the actual deploy pipeline: `.github/workflows/deploy.yml` deploys on every push to `main` (and separately on `v*.*.*` tags) using a **short-sha** image tag (`ghcr.io/adorsys/status-list-server:sha-<short_sha>`), regardless of whether a version bump happened at all. In other words, today, cutting a "version" via `bump_version.yml` has zero effect on what actually gets deployed — the deployed artifact is always identified by commit sha, not by the crate/release version. Any automated versioning solution needs to close this gap explicitly, or the two systems (release process vs. deploy process) will keep silently diverging exactly as they do today.

## Proposal

1. Introduce `release-plz` (or equivalent: `release-please`, `cargo-release` + `git-cliff`) configured for Conventional Commits.
2. Replace both `bump_version.yml` and the `cargo bump`-running portion of `semver_checks.yml` with a single "release PR" job: every push to `main` opens/updates one PR containing the version bump + generated `CHANGELOG.md` entry, computed from one source of truth.
4. Merging the release PR is the single trigger for: creating the Git tag (`v*.*.*`), publishing a GitHub Release with the changelog body, and kicking off image build + deploy (see companion ticket #248).
5. Decide explicitly how the resulting semver tag maps onto the deploy pipeline's image tagging: today `deploy.yml` tags images as semver (only on `v*.*.*` push), `latest` (only on `main`), and short-sha (always), but deploys using the short-sha tag specifically — this ticket's automation should make the semver tag (from the merged release PR) the one that actually gets deployed, or #248 needs to explicitly change the deploy step to consume it instead of short-sha.
6. Remove both the old manual `bump_version.yml` and `semver_checks.yml` dispatch jobs once the automated flow is validated.

## Workflow After This Ticket

```
Developer commits with Conventional Commit message
    │
    ▼
Merge to main (after CI passes)
    │
    ▼
release-plz opens/updates Release PR
 ├─ Cargo.toml version: 0.4.1 → 0.5.0
 ├─ CHANGELOG.md: new entry with all commits since last release
    │
    ▼
Maintainer reviews and merges Release PR
    │
    ▼
Automatic on merge:
 ├─ Git tag created: v0.5.0
 ├─ GitHub Release published with changelog
 ├─ Docker image built and tagged: ghcr.io/adorsys/status-list-server:0.5.0
 └─ Triggers production deploy workflow (#248) with semver tag
```

## Acceptance Criteria

- [x] A release-PR-generating tool is integrated and runs on every push to `main`.
- [x] The release PR contains a correct semver bump (derived from commit history/Conventional Commits) and a generated `CHANGELOG.md` entry.
- [x] Merging the release PR is the sole trigger for tagging/publishing the release (no more direct-to-main version-bump commits from `bump_version.yml`).
- [x] Explicit decision + implementation on how the release version tag relates to the image tag actually deployed by `deploy.yml`/#248 (currently short-sha, decoupled from version). Document this decision in the PR.
- [x] Old `bump_version.yml` manual workflow and the redundant `cargo bump` step in `semver_checks.yml` are removed or clearly deprecated in favor of the new flow.
- [x] Documented in `README.md` / `CONTRIBUTING.md` / `docs/` how releases are cut going forward, including:
  - Conventional Commit format requirements
  - How to trigger a release (merge to main)
  - How to review a release PR
  - What happens automatically when release PR merges
- [x] Release process supports intermittent development: works correctly if no releases happen for 6 months, then multiple features land in rapid succession

## References

- release-plz: https://release-plz.dev/
- Conventional Commits: https://www.conventionalcommits.org/

Estimated time: 3 days

Hermann-Core:
Automates versioning and changelog generation using [release-plz](https://release-plz.dev/) and [git-cliff](https://git-cliff.org/), eliminating manual version bumping and drift between workflows.

## Acceptance Criteria

- [x] A release-PR-generating tool is integrated and runs on every push to `main` (`release-plz.yml`).
- [x] The release PR contains a correct semver bump (derived from Conventional Commits) and a generated `CHANGELOG.md` entry.
- [x] `cargo-semver-checks` replaced/removed (`semver_checks.yml` deleted as this is a binary crate with no public library API contract).
- [x] Merging the release PR is the sole trigger for tagging/publishing the release (`v*.*.*` tag creation and GitHub Release).
- [x] Explicit decision + implementation on how the release version tag relates to the image tag actually deployed by `deploy.yml` / #248 documented below.
- [x] Old `bump_version.yml` manual workflow and redundant `cargo bump` step removed.
- [x] Documented in `README.md` and `CONTRIBUTING.md` how releases are cut going forward.
- [x] Release process supports intermittent development (release-plz aggregates all commits since last release tag).

## Explicit Decision on Release Version Tag vs. Deployment Image Tag (#248)

As requested by issue #247 (Proposal Item 5 & Acceptance Criteria Item 5):
1. **Tag Creation**: Merging a release-plz Release PR creates a semver tag `v*.*.*` on `main`.
2. **Docker Image Build**: The existing `deploy.yml` triggers on `push` of `v*.*.*` tags and automatically builds and pushes Docker image tags matching `type=semver` (e.g. `ghcr.io/adorsys/status-list-server:1.0.2` and `v1.0.2`).
3. **Deployment Strategy**: Today `deploy.yml` deploys Kubernetes manifests using the short-sha image tag (`sha-<short_sha>`). For companion ticket [#248](https://github.com/adorsys/status-list-server/issues/248), the deploy workflow should be updated to deploy the semver image tag (`${{ github.ref_name }}`) when triggered by a `v*.*.*` release tag push, ensuring production deployments map directly to released semver tags.

cc @martcpp @Christiantyemele @Awambeng @ndefokou


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1
+        with:
+          toolchain: stable
+      - name: Run release-plz
+        uses: release-plz/action@2eb1d8bcb770b4c48ccfaad919734b38b51958c9 # v0.5
+        with:
+          command: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ndefokou: **BLOCKING — This breaks the tag→deploy chain.**

The `release` command runs with the default `GITHUB_TOKEN`. GitHub documented that events caused by `GITHUB_TOKEN` (except `workflow_dispatch`/`repository_dispatch`) do **not** trigger new workflow runs. So when release-plz creates the `v*.*.*` tag, the `push: tags: v*.*.*` trigger in `deploy.yml` will silently not fire. Fix: supply a PAT/GitHub App token with `Contents: write` as `RELEASE_PLZ_TOKEN` (per official release-plz docs), and fail loudly if it's missing.

ndefokou (commented):
Deep review placeholder

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,67 @@
...
+
+  # Create a PR with the new versions and changelog, preparing the next release.
+  release-plz-pr:
+    name: Release-plz PR
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'adorsys' }}
+    concurrency:
```
ndefokou, line 52: **Concurrency gap: `concurrency` is only on `release-plz-pr`, not on `release-plz-release`.** Both jobs run on every `main` push. If two pushes land close together (a normal commit, then a merged release PR), the `release` and `pr` jobs can race — the `release` job may not observe the tag/state created by a parallel `pr` run. Use a **shared** concurrency group covering both jobs (e.g. `group: release-plz-${{ github.ref }}` on both, or a workflow-level concurrency with `cancel-in-progress: false`) so tag creation and release-PR updates are serialized per ref.

ndefokou (commented):
Concurrency between the two release-plz jobs

ndefokou (commented):
Release job gating

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,67 @@
...
+  # Release unpublished packages: creates git tags and GitHub Releases
+  # when a release PR is merged.
+  release-plz-release:
+    name: Release-plz release
+    environment: release
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'adorsys' }}
```
ndefokou, line 23: **The `if: repository_owner == 'adorsys'` gate only prevents forks from running releases; it does not compensate for the missing PAT in comment #1.** Also note the asymmetry: `deploy.yml` has no such owner gate. This is a fork-safety net, not a substitute for the `RELEASE_PLZ_TOKEN` needed to trigger the deploy workflow from the created tag.

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,67 @@
...
+    concurrency:
+      group: release-plz-${{ github.ref }}
+      cancel-in-progress: false
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
ndefokou, line 57: **Good**: `fetch-depth: 0` + `persist-credentials: false` on both jobs is correct for release-plz, and the minimal `permissions: {}` at workflow level is a security plus. Nit: verify the pinned checkout SHA `3d3c42e...# v7.0.1` truly corresponds to v7.0.1 so the comment and SHA don't drift.

ndefokou (commented):
Checkout SHA pin nit

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "status-list-server"
-version = "0.1.0"
+version = "1.0.1"
```
ndefokou, line 3: **HIGH — the hand-set `1.0.1` baseline collides with an existing git tag and skips history.** The repo already has tags `v1.0.0` and `v1.0.1`, while `main`'s Cargo.toml was `0.1.0`. Jumping the crate to `1.0.1`:
1. Matches the existing `v1.0.1` tag — the next release-plz run computes the bump *from* `v1.0.1` (likely `1.0.2`), or may try to re-create `v1.0.1`, depending on how release-plz reconciles crate version vs latest tag.
2. Jumps `0.1.0` → `1.0.1`, a **major** bump by semver rules, silently discarding the changelog for all 0.x history.
Recommendation: don't hand-bump inside a machinery PR. Set the baseline to the latest real tag and let release-plz compute the next version from Conventional Commits (or pick a baseline deliberately and record the tag-reconciliation reasoning). Otherwise the first auto-created release can be a no-op or duplicate.

ndefokou (commented):
Version baseline / existing tag collision

release-plz.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+changelog_path = "./CHANGELOG.md"
+changelog_config = "cliff.toml"
+release_always = false
+semver_check = false
```
ndefokou, line 5: **`semver_check = false` + `publish = false` is reasonable but the tradeoff should be explicit.** This is a binary crate with no public library API (confirmed by `crate_type.yml` checking for `lib`), so `cargo-semver-checks` has nothing meaningful to validate — removing `semver_checks.yml` is justified, and `publish = false` is correct since the crate isn't on crates.io. But it means there is now *no* automated between-release guard; worth a short in-file comment stating why these are disabled so intent survives. Also confirm no other workflow still references `semver_checks.yml`/`bump_version.yml` after deletion (none found in `.github/` at review time).

ndefokou (commented):
semver_check/publish disabled tradeoff

release-plz.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+changelog_path = "./CHANGELOG.md"
```
ndefokou, line 2: **Criterion 5 is only half-implemented — the deploy path is deferred to #248, and docs overstate it.** The PR correctly documents the decision (semver tag on merge; `deploy.yml` still deploys `sha-<short_sha>`; switching to `${{ github.ref_name }}` on `v*.*.*` is #248). That is an honest, explicit decision. But this ticket *by itself* does not close the deploy gap — production is unchanged until #248 lands. Suggest: (1) make the #248 dependency a visible guardrail (a comment/check in `deploy.yml` that must be flipped when #248 merges) so tags/releases can't silently stop driving deploys; (2) tighten README wording so readers don't conclude the release now drives production — it still doesn't until #248.

ndefokou (commented):
Deploy path half-implemented (defers to #248)

cliff.toml
```diff
@@ -0,0 +1,143 @@
...
+postprocessors = [
+  { pattern = ' +\n', replace = "\n" },
+]
+
+[git]
+# parse the commits based on https://www.conventionalcommits.org
+conventional_commits = true
+# filter out the commits that are not conventional
+filter_unconventional = true
+# process each line of a commit as an individual commit
```
ndefokou, line 98: **`filter_unconventional = true` vs `filter_commits = false` is contradictory.** `filter_unconventional = true` drops non-conventional commits from the changelog, while `filter_commits = false` keeps commits not matched by the parsers. Two filters with opposite intent will confuse maintainers. Pick one behavior: if you want *only* conventional commits (recommended, matches the CONTRIBUTING.md 'every commit must be Conventional' policy), keep `filter_unconventional = true` and set `filter_commits = true`, plus add a CI check rejecting non-conventional commits on `main`. Otherwise set `filter_unconventional = false`.

ndefokou (commented):
Contradictory commit filters in cliff.toml

cliff.toml
```diff
@@ -0,0 +1,143 @@
...
+  { message = "^build\\(release\\)", skip = true },
+  { message = "^build", group = "<!-- 08 -->Build" },
+  { body = ".*security", group = "<!-- 09 -->Security" },
+  { message = "^ci", group = "<!-- 10 -->Continuous Integration" },
+  { message = "^revert", group = "<!-- 11 -->Reverted Commits" },
+]
+
+# protect breaking changes from being skipped due to matching a skipping commit_parser
+protect_breaking_commits = false
+# filter out the commits that are not matched by commit parsers
```
ndefokou, line 132: **`protect_breaking_commits = false` is a footgun for release correctness.** Several `skip = true` parsers exist (e.g. `^chore(deps)`, `^chore: release`, `^build(deps)`). A `feat!:` / `BREAKING CHANGE:` commit that also matches a skip pattern would be skipped, silently dropping its **major** bump → a `minor` release where `major` was required, recreating the drift this ticket targets. Enable `protect_breaking_commits = true` so breaking commits are never swallowed by skip parsers.

ndefokou (commented):
protect_breaking_commits disabled is a footgun

cliff.toml
```diff
@@ -0,0 +1,143 @@
...
+filter_commits = false
+# glob pattern for matching git tags
+tag_pattern = "v[0-9]*"
+# regex for skipping tags
```
ndefokou, line 136: **`tag_pattern = "v[0-9]*"` is right for the `v*.*.*` scheme, but pre-release handling is inconsistent**: `skip_tags = "beta|alpha"` yet `ignore_tags = "rc"` (plus `release_always = false` in release-plz.toml). If a `v1.2.0-rc.1` or `-beta.1` is ever shipped, `rc` vs `beta`/`alpha` will behave differently. Pre-releases currently aren't a supported path (given `release_always = false`) — simplify to one consistent policy or document them as intentionally unsupported. Also prefer `topo_order = true` so tags aren't mis-ordered when multiple release PRs merge in succession.

ndefokou (commented):
Pre-release tag handling inconsistent; topo_order

cliff.toml
```diff
@@ -0,0 +1,143 @@
+# https://git-cliff.org/docs/configuration
+
+[remote.github]
+owner = "adorsys"
+repo = "status-list-server"
+
+[changelog]
+# changelog header
+header = """
+# Changelog
```
ndefokou, line 10: **The `# Changelog` header is duplicated** — it appears here in this `header` block *and* statically in the committed `CHANGELOG.md` (which already contains 'All notable changes...' + the markdownlint-disable comment). release-plz writes generated content into the existing `CHANGELOG.md` using this header, so the first generated release will likely produce duplicate headers/intro text. Pick a single owner for the header (empty `header = """"""` here with a static header in the file, or remove the header from `CHANGELOG.md`). Also confirm your markdownlint/CI actually honors the disable comment — the disabled rules are exactly the ones the generated body will trigger.

ndefokou (commented):
Duplicated Changelog header

Christiantyemele (commented):
..

.github/workflows/release-plz.yml
Awambeng, line 1: Since `release-plz` derives semantic version bumps from Conventional Commits, do we already enforce Conventional Commit messages and PR titles via CI/branch protection? If not, we should add enforcement to ensure release versioning remains reliable.

Christiantyemele: @Hermann-Core i will approve onces Bolan's review is addressed https://github.com/adorsys/status-list-server/pull/330#discussion_r3721950901. i guess we shall create a follow up ticket for it

Christiantyemele (approved):

CONTRIBUTING.md
martcpp, line 1: AC 4 is unmet and CONTRIBUTING.md documents behavior that does not exist. Either implement the semver→deploy wiring or explicitly defer to #248  but correct the documentation either way. Shipping docs that describe a non-existent :1.1.0 image is worse than shipping no docs

martcpp (changes requested):

.github/workflows/release-plz.yml
Hermann-Core: Confirmed — the `if: github.repository_owner == 'adorsys'` guard is a fork-safety net to prevent release jobs running in forks. It does not compensate for the `RELEASE_PLZ_TOKEN` issue. Both serve distinct purposes and both are needed. The `RELEASE_PLZ_TOKEN` fix (switching `GITHUB_TOKEN` → `RELEASE_PLZ_TOKEN` in the release job) addresses the downstream workflow-triggering concern separately.

.github/workflows/release-plz.yml
Hermann-Core: SHA `3d3c42e5aac5ba805825da76410c181273ba90b1` is `actions/checkout@v7.0.1`. The comment and SHA are in sync no drift.

Cargo.toml
Hermann-Core: Setting `version = "1.0.1"` is the correct way to bootstrap release-plz into an existing repo that already has git tags `v1.0.0` and `v1.0.1`.

release-plz's behavior:
1. Finds the latest git tag: `v1.0.1`
2. Reads Cargo.toml version: `1.0.1`
3. They match → considers `v1.0.1` the last published release
4. Computes the next bump from commits since `v1.0.1` → e.g. `v1.0.2` or `v1.1.0`
5. Will **not** try to recreate `v1.0.1` — it checks tag existence before acting

The concern about the historical jump from `0.1.0` → `1.0.1` is understood, but tags `v1.0.0` and `v1.0.1` already existed in the repo before this PR. The prior version history is preserved in those git tags. Setting Cargo.toml to match the latest existing tag is the release-plz-recommended bootstrapping approach.

release-plz.toml
Hermann-Core: This is a binary crate with no public library API, so `semver_check = false` is correct (cargo-semver-checks has nothing meaningful to validate). `publish = false` is correct since we distribute as a Docker image, not via crates.io.

release-plz.toml
Hermann-Core: Fixed in CONTRIBUTING.md. The diagram now accurately separates the two steps:

- **Docker image**: `deploy.yml` triggered by the `v*.*.*` tag **does** build and push a semver-tagged image (e.g. `ghcr.io/adorsys/status-list-server:1.2.0`) — this is correct today.
- **EKS deployment**: the `deploy-to-eks` job still rolls out using `sha-<short_sha>`, not the semver tag. This is explicitly called out with a note referencing #248.

The prose and diagram no longer overstate what is wired up today.

cliff.toml
Hermann-Core: These two settings are complementary, not contradictory:

- `filter_unconventional = true`: drops non-conventional commits (first pass, before parsers run)
- `filter_commits = false`: among the conventional commits that remain, keeps ALL of them even if no parser group matches them (second pass)

The net result: only conventional commits appear in the changelog, but a conventional commit with a type not yet in the parsers still shows up rather than being silently dropped. This is intentionally defensive — if a new conventional type is used before being added to the parsers, it surfaces in the changelog as an ungrouped entry instead of vanishing.

Changing `filter_commits = true` would be *stricter* but also worse: it would silently discard any conventional commit that doesn't match a parser, which is harder to notice and debug. Keeping `false` here is the right call.

cliff.toml
Hermann-Core: `CHANGELOG.md` is now empty — the previous placeholder content was removed in the last commit. When release-plz runs for the first time, it will write the `cliff.toml` header + the generated body into the file from scratch. There is no duplication risk in the current state.

.github/workflows/release-plz.yml
Hermann-Core: Good point — currently enforcement relies on developer discipline + PR review. Conventional Commit messages are documented as mandatory in CONTRIBUTING.md but not gated by CI. Adding a commitlint check (or PR title linting via a GitHub Action) would make the enforcement automatic. Suggesting we track this as a follow-up ticket rather than blocking this PR, since the release automation itself is already valuable without CI enforcement of commit format.

CONTRIBUTING.md
Hermann-Core: Fixed. CONTRIBUTING.md now accurately documents what is wired up today vs. what is deferred:

- The semver Docker image **is** built and pushed when the `v*.*.*` tag fires `deploy.yml` (this part was already correct in `deploy.yml`)
- The EKS rollout step still uses `sha-<short_sha>`, not the semver tag — now clearly called out with a note and a reference to #248

The diagram no longer implies a semver Kubernetes deployment that doesn't exist yet.

Hermann-Core:
Thank you @martcpp @Awambeng @ndefokou for reviews. Can you review back?

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1
+        with:
+          toolchain: stable
+      - name: Run release-plz
+        uses: release-plz/action@2eb1d8bcb770b4c48ccfaad919734b38b51958c9 # v0.5
```
Awambeng, line 66:
```suggestion
        uses: release-plz/action@2eb1d8bcb770b4c48ccfaad919734b38b51958c9 # v0.5.131
```

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,70 @@
...
```
Awambeng, line 35:
```suggestion
        uses: release-plz/action@2eb1d8bcb770b4c48ccfaad919734b38b51958c9 # v0.5.131
```

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,70 @@
...
+          # events do not trigger other workflow runs (e.g. deploy.yml on v*.*.*)
+          # See: https://release-plz.dev/docs/github/output#workflow-not-triggered
+          GITHUB_TOKEN: ${{ secrets.RELEASE_PLZ_TOKEN }}
+
+  # Create a PR with the new versions and changelog, preparing the next release.
+  release-plz-pr:
+    name: Release-plz PR
+    permissions:
+      pull-requests: write
+      contents: write
```
Awambeng, line 49: Isn't `contents: write` sufficient for this job? why do we need `pull-requests: write`?

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,70 @@
...
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  # Release unpublished packages: creates git tags and GitHub Releases
+  # when a release PR is merged.
+  release-plz-release:
+    name: Release-plz release
+    environment: release
```
Awambeng, line 18: Is this already configured in the repo? If no, please do so.

Awambeng (changes requested):

.github/workflows/release-plz.yml
Hermann-Core: Fixed — concurrency is intentionally absent from `release-plz-release`. The concurrency group lives only on `release-plz-pr` to serialise PR updates. The release job is idempotent (creating the same tag twice is a no-op) and must not be cancelled mid-run. Cancelling a release mid-flight could push a tag without the matching GitHub Release, leaving the pipeline in an inconsistent state.

.github/workflows/release-plz.yml
Hermann-Core: Fixed — the `release` job now uses `RELEASE_PLZ_TOKEN` (a PAT) rather than `GITHUB_TOKEN`. Events created by a PAT do trigger downstream workflow runs (e.g. `deploy.yml` on `v*.*.*` tags), so the tag→deploy chain is preserved.

cliff.toml
Hermann-Core: Fixed — `protect_breaking_commits = true` is now set in `cliff.toml`. Breaking commits matching a `skip = true` parser will no longer be silently dropped from the changelog; git-cliff will force-include them regardless of the skip rule.

.github/workflows/release-plz.yml
Hermann-Core: Not yet configured, the repo currently has only a `production` environment. A `release` environment needs to be created in **Settings -> Environments** with the `RELEASE_PLZ_TOKEN` secret scoped to it (and optionally a required-reviewer protection rule for added safety). Without the environment, the job will run without any protection gate but `RELEASE_PLZ_TOKEN` will need to be a repo-level secret as a fallback. Creating the environment properly is the right approach.

cliff.toml
Hermann-Core: Fixed — merged `skip_tags` and `ignore_tags` into a single consistent policy:
```toml
skip_tags  = ""              # cleared — skip_tags still anchors the comparison range, wrong for pre-releases
ignore_tags = "rc|beta|alpha" # all pre-release variants fully excluded: no changelog block, no range anchor
```
`skip_tags` generates a changelog block for the matching tag and still uses it to bound the commit range, which means `v1.0.0-beta.1` would appear as its own release section in the changelog. `ignore_tags` removes the tag completely — commits are attributed to the next stable release, which is the intended behaviour for pre-releases.

Hermann-Core:
Hello @Awambeng can you please recheck?

.github/workflows/release-plz.yml
Awambeng: Okay thanks. Hope you don't forget to create it.

.github/workflows/release-plz.yml
Awambeng: Please donot forget to create a follow-up ticket for this enforcement.

Awambeng (approved):
@Hermann-Core thanks for addressing my comments. However, please don’t forget to create the follow-up ticket for enforcing PR titles and commit messages, as well as configuring the environment required by this workflow.

Hermann-Core:
@Awambeng All requested follow-up tickets mentioned in the review comments have been created:

- https://github.com/adorsys/status-list-server/issues/351
- https://github.com/adorsys/status-list-server/issues/352

Thank you!

Hermann-Core:
@martcpp hello you need to approve before we can merge.

martcpp (approved):

PR #331, opened 2026-08-04
Migrate OTel Collector Helm deployment to official opentelemetry-collector chart

### LINKED ISSUE #298: Migrate OTel Collector Helm deployment to official opentelemetry-collector chart
## Context

PR #258 ships a custom-built Helm chart for deploying the OpenTelemetry Collector as a sidecar or standalone Deployment. During review ([discussion](https://github.com/adorsys/status-list-server/pull/258#discussion_r3646033691)), it was suggested to switch to the official [`opentelemetry-collector` Helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts) instead of maintaining a custom one.

## Why deferred

The switch changes:
- `values.yaml` key layout (chart-owned vs. application-owned keys)
- Helm dependency management (`Chart.yaml` `dependencies:`)
- Upgrade and rollback semantics
- ConfigMap naming and volume mount conventions

Expanding scope mid-PR would have blocked merging the core OTel fix. The current custom chart is now validated in CI.

## Proposed work

- [x] Add `opentelemetry-collector` as a Helm chart dependency
- [x] Migrate `helm/chart/templates/otel-collector-*.yaml` templates to use the upstream chart's values
- [x] Update `helm/chart/values.yaml` to follow upstream chart structure
- [x] Validate rendered Helm Collector config in CI against the upstream chart image
- [x] Update Helm chart documentation

## References
- Official chart: https://github.com/open-telemetry/opentelemetry-helm-charts

Estimated time: 1 day

Hermann-Core:
Migrates the OpenTelemetry Collector Helm deployment from custom templates (`otel-collector-configmap.yaml` and `otel-collector-deployment.yaml`) to the official `open-telemetry/opentelemetry-collector` subchart (`v0.166.0` / appVersion `0.157.0`).

## Acceptance Criteria

- [x] Add `opentelemetry-collector` as a Helm chart dependency (`Chart.yaml`).
- [x] Migrate `helm/chart/templates/otel-collector-*.yaml` templates to use the upstream chart's values.
- [x] Update `helm/chart/values.yaml` to follow upstream chart structure (`opentelemetry-collector` key).
- [x] Validate rendered Helm Collector config in CI against the upstream chart image (`0.157.0`).
- [x] Update Helm chart documentation (`helm/README.md`).

cc @martcpp @Awambeng @forkimenjeckayang


.github/workflows/CI.yml
```diff
@@ -339,7 +341,7 @@ jobs:
         run: |
           docker run --rm \
             -v "$PWD/otel-collector.yaml:/etc/otelcol/config.yaml:ro" \
-            otel/opentelemetry-collector-contrib:0.156.0 \
+            otel/opentelemetry-collector-contrib:0.157.0 \
```
Awambeng:
```suggestion
            otel/opentelemetry-collector-contrib:0.158.0 \
```

.github/workflows/CI.yml
```diff
@@ -385,7 +389,7 @@ jobs:
         run: |
           docker run --rm \
             -v "/tmp/helm-otel-collector.yaml:/etc/otelcol/config.yaml:ro" \
-            otel/opentelemetry-collector-contrib:0.156.0 \
+            otel/opentelemetry-collector-contrib:0.157.0 \
```
Awambeng:
```suggestion
            otel/opentelemetry-collector-contrib:0.158.0 \
```
Please update this version in the docker compose as well.

helm/chart/values.yaml
```diff
@@ -246,38 +246,86 @@ redis-ha:
...
-    resources:
-      requests:
-        memory: "128Mi"
-        cpu: "100m"
-      limits:
-        memory: "256Mi"
-        cpu: "200m"
+  mode: deployment
+  image:
+    repository: otel/opentelemetry-collector-contrib
```
Awambeng, line 234: Could you please add a tag here?

.trivyignore.yaml
```diff
@@ -3,6 +3,8 @@ misconfigurations:
     paths:
       - "*/charts/postgres*/**"
       - "*/charts/redis-ha*/**"
+      - "*/charts/opentelemetry-collector*/**"
```
Awambeng: Why don't we just reference the entire charts directory?

.trivyignore.yaml
```diff
@@ -3,6 +3,8 @@ misconfigurations:
...
   - id: AVD-KSV-0118
     paths:
       - "*/charts/redis-ha*/**"
+      - "*/charts/opentelemetry-collector*/**"
```
Awambeng: same comment applies here

docker-compose.yml
```diff
@@ -127,7 +127,7 @@ services:
       - status-list-network
 
   otel-collector:
-    image: otel/opentelemetry-collector-contrib:0.156.0
+    image: otel/opentelemetry-collector-contrib:0.157.0
```
Awambeng:
```suggestion
    image: otel/opentelemetry-collector-contrib:0.158.0
```

supply-chain/config.toml
Awambeng, line 1: I noticed this commit https://github.com/adorsys/status-list-server/pull/331/changes/b2083374ded4414a3c0af337216ce9dfca10fc1f has the message "revert updated dependencies", but the PR still contains dependency updates. Was this intentional?

helm/chart/templates/otel-collector-deployment.yaml
Awambeng, line 1: Could you as well remove all reference to the custom `otelCollector` explanations in the `observability.md` documentation?

helm/chart/templates/network-policy.yaml
```diff
@@ -21,7 +21,8 @@ spec:
       from:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- if .Values.otelCollector.enabled }}
+    {{- $otel := index .Values "opentelemetry-collector" | default dict }}
+    {{- if $otel.enabled }}
     - ports:
         - port: 4317
           protocol: TCP
```
Awambeng: `4317/4318` are collector ports, but the app is the OTLP client, not the server. This should only require egress to the collector. Can we remove this ingress rule?

helm/chart/Chart.yaml
```diff
@@ -11,3 +11,7 @@ dependencies:
   - name: redis-ha
     version: 4.35.0
     repository: https://dandydeveloper.github.io/charts
+  - name: opentelemetry-collector
+    version: 0.166.0
+    repository: https://open-telemetry.github.io/opentelemetry-helm-charts
+    condition: opentelemetry-collector.enabled
```
Awambeng, line 18: This line can be dropped as the default condition of the dependency is `condition: opentelemetry-collector.enabled`

.github/workflows/CI.yml
```diff
@@ -256,6 +256,7 @@ jobs:
       - name: Build Helm dependencies
         run: |
           helm repo add dandydeveloper https://dandydeveloper.github.io/charts
+          helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
```
Awambeng, line 261: Could you please share reference to the default values of this chart? Github link will be enough.

helm/chart/templates/deployment.yaml
```diff
@@ -85,25 +85,13 @@ spec:
...
-                {{- end }}
-              {{- else }}
-                {{- fail "otelCollector.enabled=true requires either sidecar.enabled=true or standalone.enabled=true" }}
+            {{- $otel := index .Values "opentelemetry-collector" | default dict }}
+            {{- if $otel.enabled }}
+              {{- if not (hasKey $env "APP_TELEMETRY__ENABLED") }}
+                {{- $_ := set $env "APP_TELEMETRY__ENABLED" "true" }}
+              {{- end }}
+              {{- if not (hasKey $env "APP_TELEMETRY__OTLP_ENDPOINT") }}
+                {{- $_ := set $env "APP_TELEMETRY__OTLP_ENDPOINT" (printf "http://%s-opentelemetry-collector.%s.svc.cluster.local:4317" .Release.Name .Release.Namespace) }}
```
Awambeng: Could you please derive the chart `name` and `port` from the `values.yaml`?

helm/chart/values.yaml
```diff
@@ -246,38 +246,86 @@ redis-ha:
...
+        limit_percentage: 75
+        spike_limit_percentage: 15
+      batch:
+        timeout: 5s
+        send_batch_size: 1024
     exporters:
-      otlpEndpoint: ""
-      otlpTlsInsecure: true
+      debug:
+        verbosity: basic
```
Awambeng, line 306: The old custom chart supported a backend (otelCollector.config.exporters.otlpEndpoint ) and a  prometheus  exporter on :8889. The compose config ( otel-collector.yaml ) still exports  traces → otlp/jaeger  and  metrics → prometheus:8889 . After this migration, traces, metrics and logs only go to the collector's stdout — Jaeger/Prometheus/Grafana get nothing in production.

Either re-add a configurable  otlp/backend  +  prometheus  exporter (like compose), or confirm the intended backend was "logs only" — but then the chart's Prometheus annotations on the app are also misleading.

helm/chart/values.yaml
```diff
@@ -246,38 +246,86 @@ redis-ha:
...
+    jaeger-compact:
+      enabled: false
+    jaeger-thrift:
+      enabled: false
+    jaeger-grpc:
+      enabled: false
+    zipkin:
+      enabled: false
+    metrics:
+      enabled: true
```
Awambeng, line 284: The subchart's default for  ports.metrics.containerPort  is 8888 (I checked the chart's  values.yaml ) — the old setup and compose both use 8889. Also, the custom templates that carried  prometheus.io/scrape  annotations on the collector pod were deleted, and no  serviceMonitor / podMonitor  is enabled in the new values. 

Result: the metrics port is dead — even the collector's own self-metrics aren't scraped. If you intend to expose collector metrics, pin  containerPort: 8889  and add a  podMonitor / serviceMonitor ; otherwise just set  enabled: false

Awambeng (changes requested):

.trivyignore.yaml
forkimenjeckayang:
```yaml
      - "*/charts/opentelemetry-collector*/**"
```
> These new Collector exclusions are not needed by the rendered chart. I rendered the default values and ran Trivy `0.69.3` without the repository ignore file; the Collector ConfigMap, Deployment, Service, and ServiceAccount all reported zero misconfigurations, including for `KSV-0014` and `KSV-0118`. The configured pod/container security contexts already satisfy those checks.
>
Please remove both `opentelemetry-collector` paths instead of broadening the exemption. Keeping an unused path-level suppression could hide a future regression in the upstream subchart.

helm/chart/values.yaml
```diff
@@ -246,38 +246,86 @@ redis-ha:
 redis:
   enabled: true
 
-otelCollector:
+opentelemetry-collector:
   enabled: true
```
forkimenjeckayang, line 231: > The official chart defaults `serviceAccount.automountServiceAccountToken` to `true`, and the rendered Collector pod therefore receives a Kubernetes API token. This configuration uses only OTLP receivers, memory/batch processors, debug exporters, and Downward API environment fields; it creates no ClusterRole and has no component that needs Kubernetes API access.
>

Please set `serviceAccount.automountServiceAccountToken: false` under the subchart values. This keeps the migrated deployment least-privileged and avoids introducing a pod credential that the removed custom Deployment did not mount explicitly.

forkimenjeckayang (changes requested):
The existing exporter, endpoint naming, documentation, network-policy, and dependency-scope comments by @Awambeng  remain valid, so they do not need to be duplicated.

Added two smaller inline findings: the new Collector Trivy exclusions are unnecessary under the rendered security contexts, and the Collector should disable automatic ServiceAccount token mounting because this configuration does not access the Kubernetes API.

.github/workflows/CI.yml
Hermann-Core: Bumped to `0.158.0` — coordinated across `values.yaml` (`image.tag: "0.158.0"`), `docker-compose.yml`, and both CI validation steps. Note: the current chart dependency stays at `0.166.0` (the Helm chart for appVersion `0.158.0` would be `0.167.0`, which hasn't been published yet). The `image.tag` override ensures the correct binary is pulled regardless of the chart's bundled default.

.trivyignore.yaml
Hermann-Core: The fix is actually narrower than broadening to the whole `charts/` directory — per @forkimenjeckayang's verification (see their comment below), Trivy `0.69.3` reports **zero misconfigurations** for the rendered OTel subchart manifests without any ignore entries. The `opentelemetry-collector` exemptions have been removed entirely rather than broadened.

docker-compose.yml
Hermann-Core: Updated to `0.158.0`. The three touch points (values.yaml image.tag, docker-compose.yml, CI validation images) are now all in sync.

.github/workflows/CI.yml
Hermann-Core: Both CI validation steps bumped to `0.158.0` as well — now consistent with `values.yaml` and `docker-compose.yml`.

.trivyignore.yaml
Hermann-Core: Same as the comment above — removed rather than broadened. The entries are no longer needed per @forkimenjeckayang's Trivy verification.

helm/chart/values.yaml
Hermann-Core: Added `tag: "0.158.0"` to the image block. Pinning the tag explicitly is safer than relying on the chart's implicit appVersion default, especially when the image version and chart version don't align 1-to-1.

supply-chain/config.toml
Hermann-Core: Intentional. The `revert updated dependencies` commit reverted changes to `supply-chain/config.toml` (cargo-vet audit exemptions for new Rust crate transitive dependencies) that were added during development but are unrelated to this Helm migration. Including those would have mixed Rust supply-chain audit changes into a Helm chart PR. The OTel subchart dependency addition in `Chart.yaml`/`Chart.lock` is separate from Rust supply-chain auditing and is retained.

helm/chart/templates/otel-collector-deployment.yaml
Hermann-Core: Updated in `docs/observability.md`. The old `otelCollector.sidecar` and `otelCollector.standalone` custom-chart references have been removed and replaced with accurate documentation for the `opentelemetry-collector.*` subchart values, including a production exporter override example.

helm/chart/values.yaml
Hermann-Core: Fixed. Added `serviceAccount.automountServiceAccountToken: false` under the `opentelemetry-collector` key in `helm/chart/values.yaml` to maintain least-privilege security posture.

helm/chart/templates/network-policy.yaml
Hermann-Core: Fixed. Removed the ingress rule for ports 4317/4318 from `helm/chart/templates/network-policy.yaml`. The status-list-server application pod is an OTLP client (sending telemetry to the collector via egress), not a server receiving OTLP traffic.

helm/chart/templates/deployment.yaml
Hermann-Core: Updated `helm/chart/templates/deployment.yaml` to derive the OTLP port dynamically using `dig "ports" "otlp" "servicePort" 4317 $otel` from `values.yaml` instead of hardcoding 4317.

.trivyignore.yaml
Hermann-Core: Removed both `opentelemetry-collector` exemption paths from `.trivyignore.yaml`. Confirmed that Trivy scan passes with zero misconfigurations.

helm/chart/values.yaml
Hermann-Core: This is by design for the baseline migration. The subchart default configures a `debug` exporter (emits telemetry to collector stdout logs). Production backend exporter targets (e.g. Jaeger gRPC, Prometheus scrape endpoint) are environment-specific and belong in environment-specific Helm values overrides, as documented in `docs/observability.md` under *Configuring Exporters*.

helm/chart/Chart.yaml
Hermann-Core: Helm subcharts do not have an implicit default condition matching their dependency name. Without `condition: opentelemetry-collector.enabled`, Helm would unconditionally render the dependency templates regardless of the `opentelemetry-collector.enabled` setting in `values.yaml`. This line is required to make the subchart deployment conditional.

helm/chart/values.yaml
Hermann-Core: Acknowledged — for this initial migration PR, we stick with the standard OTLP receivers and `debug` exporter in default values. Enabling a Prometheus metrics exporter/port or PodMonitor/ServiceMonitor for production self-metrics/application metrics can be configured via environment overrides as detailed in `docs/observability.md`.

.github/workflows/CI.yml
Hermann-Core: Added link to the upstream chart's default `values.yaml` in `.github/workflows/CI.yml` as well as in `docs/observability.md`:\nhttps://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml

Hermann-Core:
Hello @forkimenjeckayang @Awambeng thank you for the reviews. Can you recheck?

helm/chart/templates/deployment.yaml
```diff
@@ -87,25 +87,14 @@ spec:
...
-                {{- fail "otelCollector.enabled=true requires either sidecar.enabled=true or standalone.enabled=true" }}
+            {{- $otel := index .Values "opentelemetry-collector" | default dict }}
+            {{- if $otel.enabled }}
+              {{- if not (hasKey $env "APP_TELEMETRY__ENABLED") }}
+                {{- $_ := set $env "APP_TELEMETRY__ENABLED" "true" }}
+              {{- end }}
+              {{- if not (hasKey $env "APP_TELEMETRY__OTLP_ENDPOINT") }}
+                {{- $otlpPort := dig "ports" "otlp" "servicePort" 4317 $otel }}
+                {{- $_ := set $env "APP_TELEMETRY__OTLP_ENDPOINT" (printf "http://%s-opentelemetry-collector.%s.svc.cluster.local:%d" .Release.Name .Release.Namespace ($otlpPort | int)) }}
               {{- end }}
```
forkimenjeckayang, line 98: > The OTLP Service port is now derived, but the Service name is still constructed as `<release>-opentelemetry-collector`. The official subchart exposes `fullnameOverride` and `nameOverride`; rendering with `opentelemetry-collector.fullnameOverride=telemetry-gateway` creates Service `telemetry-gateway`, while the application still receives `http://statuslist-opentelemetry-collector.<namespace>.svc.cluster.local:<port>`.
>

Please either derive the same name that the subchart renders or expose one explicit parent-chart Collector endpoint/name value and use it for application wiring. If name overrides are intentionally unsupported by this parent chart, reject or document that limitation instead of silently producing a broken endpoint.

docs/observability.md
```diff
@@ -19,27 +19,61 @@ Telemetry behavior is determined primarily by the deployment environment (`APP_E
...
+          insecure: true
+      prometheus:
+        endpoint: "0.0.0.0:8889"
+    service:
+      pipelines:
+        traces:
+          exporters: [otlp/backend, debug]
+        metrics:
+          exporters: [prometheus, debug]
+```
```
forkimenjeckayang, line 67: > The application itself still exposes metrics through its in-process `/metrics` registry, so the migration does not remove the existing direct Prometheus path. However, this production override example adds a Collector Prometheus exporter on `0.0.0.0:8889` without adding a matching `opentelemetry-collector.ports` entry or any ServiceMonitor, PodMonitor, or scrape annotations. Rendering the example leaves only Collector self-metrics exposed on `8888`; the documented `8889` exporter is not discoverable through the generated Service.
>
Please either complete the example with the required port and scrape-discovery configuration, or remove the Collector Prometheus exporter example and state clearly that Prometheus should scrape the application's existing `/metrics` endpoint. Collector self-metrics on `8888` should be documented separately.

forkimenjeckayang (commented):
a few more comments @Hermann-Core 
Please check when free

helm/chart/templates/deployment.yaml
Hermann-Core: Fixed — the template now follows the same `fullnameOverride → nameOverride → default` naming logic as the subchart itself:

```yaml
{{- $collectorSvcName := printf "%s-opentelemetry-collector" .Release.Name }}
{{- if $otel.fullnameOverride }}
  {{- $collectorSvcName = $otel.fullnameOverride | trunc 63 | trimSuffix "-" }}
{{- else if $otel.nameOverride }}
  {{- $collectorSvcName = printf "%s-%s" .Release.Name ($otel.nameOverride | trunc 63 | trimSuffix "-") }}
{{- end }}
```

Verified with `helm template --set 'opentelemetry-collector.fullnameOverride=telemetry-gateway'` → endpoint is correctly `http://telemetry-gateway.<namespace>.svc.cluster.local:4317`.

docs/observability.md
Hermann-Core: Fixed in `docs/observability.md` — the production override example now includes the required `ports.prometheus` block alongside the exporter config:

```yaml
opentelemetry-collector:
  ports:
    prometheus:
      enabled: true
      containerPort: 8889
      servicePort: 8889
      protocol: TCP
  config:
    exporters:
      prometheus:
        endpoint: "0.0.0.0:8889"
    ...
```

Without this entry, the Service does not include port 8889 and Prometheus cannot reach the exporter even though the container is listening. Also added a note that either `prometheus.io/scrape` pod annotations or a `ServiceMonitor`/`PodMonitor` is still needed for Prometheus auto-discovery.

Hermann-Core:
Hello @forkimenjeckayang can you please recheck?

helm/chart/templates/deployment.yaml
```diff
@@ -87,25 +87,21 @@ spec:
...
+              {{- end }}
+              {{- if not (hasKey $env "APP_TELEMETRY__OTLP_ENDPOINT") }}
+                {{- $otlpPort := dig "ports" "otlp" "servicePort" 4317 $otel }}
+                {{- /* Derive service name following subchart fullnameOverride → nameOverride → default */}}
+                {{- $collectorSvcName := printf "%s-opentelemetry-collector" .Release.Name }}
+                {{- if $otel.fullnameOverride }}
+                  {{- $collectorSvcName = $otel.fullnameOverride | trunc 63 | trimSuffix "-" }}
+                {{- else if $otel.nameOverride }}
+                  {{- $collectorSvcName = printf "%s-%s" .Release.Name ($otel.nameOverride | trunc 63 | trimSuffix "-") }}
                 {{- end }}
```
forkimenjeckayang: The `fullnameOverride` case is fixed, but the generated endpoint still differs from the official subchart for some valid names.

The upstream chart uses the release name directly when it already contains the effective chart name. The current code always appends the name, producing incorrect endpoints in these reproducible cases:

- Release `otel` with `nameOverride=otel`:
  - Service: `otel`
  - App endpoint: `otel-otel`
- Release `prod-opentelemetry-collector` with default values:
  - Service: `prod-opentelemetry-collector`
  - App endpoint: `prod-opentelemetry-collector-opentelemetry-collector`

Please mirror the upstream helper's naming logic: determine the effective name, use the release name when it already contains that name, otherwise concatenate them, and apply the final 63-character truncation. Alternatively, expose an explicit parent-chart Collector endpoint instead of duplicating the subchart helper.

You can check with commands
```bash
helm template otel helm/chart \
  --namespace telemetry \
  --set opentelemetry-collector.nameOverride=otel \
  > /tmp/otel-name-override.yaml

sed -n \
  '/# Source: .*opentelemetry-collector\/templates\/service.yaml/,/^---$/p' \
  /tmp/otel-name-override.yaml | rg '^(kind:|  name:)'

rg -A1 'name: APP_TELEMETRY__OTLP_ENDPOINT' \
  /tmp/otel-name-override.yaml
```

```bash
helm template otel helm/chart \
  --namespace telemetry \
  --set opentelemetry-collector.nameOverride=otel \
  > /tmp/otel-name-override.yaml

sed -n \
  '/# Source: .*opentelemetry-collector\/templates\/service.yaml/,/^---$/p' \
  /tmp/otel-name-override.yaml | rg '^(kind:|  name:)'

rg -A1 'name: APP_TELEMETRY__OTLP_ENDPOINT' \
  /tmp/otel-name-override.yaml
```

forkimenjeckayang (commented):
one last comment @Hermann-Core

helm/chart/templates/deployment.yaml
Hermann-Core: Thanks for the catch, now fixed. The template now mirrors the upstream `opentelemetry-collector.fullname` helper exactly, including the `contains` check that was missing:

```yaml
{{- $effectiveName := $otel.nameOverride | default "opentelemetry-collector" }}
{{- if contains $effectiveName .Release.Name }}
  {{- $collectorSvcName = .Release.Name | trunc 63 | trimSuffix "-" }}
{{- else }}
  {{- $collectorSvcName = printf "%s-%s" .Release.Name $effectiveName | trunc 63 | trimSuffix "-" }}
{{- end }}
```

Verified with `helm template` against all the cases you raised:

| Release | Override | Service | OTLP endpoint |
|---|---|---|---|
| `otel` | `nameOverride=otel` | `otel` ✅ | `http://otel.<ns>.svc.cluster.local:4317` ✅ |
| `prod-opentelemetry-collector` | defaults | `prod-opentelemetry-collector` ✅ | matches ✅ |
| `statuslist` | defaults | `statuslist-opentelemetry-collector` ✅ | matches ✅ |
| `statuslist` | `fullnameOverride=telemetry-gateway` | `telemetry-gateway` ✅ | matches ✅ |

Hermann-Core:
@forkimenjeckayang can you please check again?

forkimenjeckayang (approved):
Thanks for addressing my comments.

helm/chart/Chart.yaml
```diff
@@ -12,3 +12,7 @@ dependencies:
     version: 4.35.0
     repository: https://dandydeveloper.github.io/charts
     condition: redis-ha.enabled
+  - name: opentelemetry-collector
+    version: 0.166.0
```
Awambeng:
```suggestion
    version: 0.169.0
```
This will make chart version compatible with the appVersion.
see screenshot below:

Image

helm/chart/values.yaml
```diff
@@ -224,38 +224,97 @@ redis-ha:
         external-dns.alpha.kubernetes.io/hostname: redis.eudi-adorsys.com
         external-dns.alpha.kubernetes.io/delete: "true"
 
-otelCollector:
+redis:
   enabled: true
```
Awambeng: This looks like dead code can we remove it?

helm/chart/values-local.yaml
Awambeng, line 1: The `redis-ha`  block is dead weight now that `redis-ha.enabled`  defaults to `false`. Either delete it or explicitly set  `redis-ha.enabled: true`  if local Redis is still wanted.

Awambeng (changes requested):

Hermann-Core:
@Awambeng can you recheck?

Awambeng (approved):

PR #334, opened 2026-08-05
feat: update Redis integration and documentation for optional caching

### LINKED ISSUE #306: Decide and document Redis role; keep it optional
## Objective

Decide and document the role of Redis in the server, then make any implementation changes needed so Redis remains optional.

## Current State

Redis is not on the hot path for status-list reads: status-list caching is handled by `MokaStatusListCache` in `src/outbound/cache.rs`. Redis is currently used as a `Storage` implementation in `src/outbound/redis.rs`, mainly as an optional cache layer for certificate material loaded from object storage.

The Helm chart, Docker Compose, and docs still make Redis feel more central than the code does.

## Recommendation

Do not remove Redis outright. Keep it as an optional distributed certificate-material cache for multi-replica deployments, but make the default path in-memory/local and document when Redis is worth the operational cost.

## Deliverables

- [ ] Audit all Redis uses in code, Docker Compose, Helm, and docs.
- [ ] Confirm Redis is not required for core status-list persistence or reads.
- [ ] Document the tradeoff: cross-replica cache sharing and reduced backend reads vs. extra operational dependency, TLS/credential management, and HA complexity.
- [ ] Ensure builds without the `redis` feature do not require Redis settings or Helm dependencies.
- [ ] Update Helm values so Redis can be disabled cleanly without broken env vars, init containers, or mounted secrets.

## Acceptance Criteria

- [ ] A no-Redis deployment path is documented and works.
- [ ] Redis-enabled deployment remains documented and tested for the certificate-cache use case.
- [ ] README and Helm docs clearly state Redis is optional.
- [ ] Any default that implies Redis is mandatory is removed or justified.

## Verification

- [ ] `cargo check --no-default-features --features memory`
- [ ] `cargo check --workspace --all-targets --features memory,acme`
- [ ] `helm template` succeeds with Redis disabled.

1 day

Ngha-Boris:
## Summary

- Make Redis explicitly optional across runtime defaults, Docker, Helm, and docs.
- Keep Redis as an opt-in distributed certificate-material cache for AWS/S3 + ACME multi-replica deployments.
- Update Helm so `redis-ha.enabled=false` renders a clean no-Redis deployment path with no Redis env vars, secret keys, init containers, cert-sync resources, or network-policy ports.

## Changes

- Changed default `redis.uri` to empty so Redis is disabled unless `APP_REDIS__URI` is set.
- Removed Redis from default Docker build/Compose feature set and put the Compose Redis service behind the `redis` profile.
- Added `condition: redis-ha.enabled` to the Helm Redis HA dependency and defaulted `redis-ha.enabled` to `false`.
- Guarded Helm Redis app env vars, ExternalSecret `redis-password`, Redis network-policy ports, and Redis TLS cert-sync resources behind Redis HA enablement.
- Updated README, Helm docs, local deployment docs, database backend docs, and Redis TLS docs to explain Redis’ role and tradeoffs.

## Verification

- `cargo check --no-default-features --features memory`
- `cargo check --workspace --all-targets --features memory,acme`
- `cargo check --workspace --all-targets --features memory,aws,acme,redis`
- `cargo test --no-default-features --features memory config::tests::test_default_config`
- `helm template statuslist ./helm/chart --namespace statuslist`
- `helm template statuslist ./helm/chart --namespace statuslist --set redis-ha.enabled=true`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,6 +44,7 @@ services:
   redis:
     image: redis:8.6
     container_name: redis
+    profiles: ["redis"]
     ports:
       - 6379:6379
     networks:
@@ -91,7 +92,7 @@ services:
       context: .
       dockerfile: Dockerfile
       args:
-        FEATURES: ${FEATURES:-postgres,redis,aws,acme}
+        FEATURES: ${FEATURES:-postgres,aws,acme}
     container_name: status-list-server
     env_file:
       - path: .env.template
@@ -108,13 +109,12 @@ services:
       APP_TELEMETRY__OTLP_ENDPOINT: "http://otel-collector:4317"
       APP_TELEMETRY__ENABLED: ${APP_TELEMETRY__ENABLED:-true}
       APP_TELEMETRY__SAMPLER_RATIO: ${APP_TELEMETRY__SAMPLER_RATIO:-1.0}
+      APP_REDIS__URI: ${APP_REDIS__URI:-}
     ports:
       - 8000:8000
     depends_on:
       db:
         condition: service_healthy
-      redis:
-        condition: service_started
       pebble:
         condition: service_started
       localstack:
--- a/helm/README.md
+++ b/helm/README.md
@@ -13,9 +13,9 @@ This guide provides instructions for deploying the Status List Server using the 
 This chart has the following dependencies:
 
 - **PostgreSQL**: A relational database for storing application data.
-- **Redis HA**: A high-availability Redis cluster for caching.
+- **Redis HA**: Optional high-availability Redis cluster for the distributed certificate-material cache.
 
-These dependencies are managed by the Helm chart and will be installed automatically.
+These dependencies are managed by the Helm chart. PostgreSQL is enabled by default; Redis HA is disabled by default and is installed only when `redis-ha.enabled=true`.
 
 ## Configuration
 
@@ -30,6 +30,26 @@ The following files are used to configure the deployment:
 - **`statuslist.image.tag`**: The Docker image tag.
 - **`postgres.persistence.enabled`**: Enable or disable persistent storage for PostgreSQL.
 - **`redis-ha.persistentVolume.enabled`**: Enable or disable persistent storage for Redis.
+- **`redis-ha.enabled`**: Enable the optional Redis HA dependency and application Redis environment variables.
+
+## Redis Role
+
+Redis is optional. The server does not use Redis for status-list persistence or status-list reads; those use the configured repository backend and the in-process Moka status-list cache. In this chart, Redis is intended only as a distributed certificate-material cache for multi-replica deployments that use the AWS S3 certificate storage path and compile the application with the `redis` feature.
+
+Keep Redis disabled for a simpler deployment. Enable it when sharing certificate cache entries across replicas and reducing object-storage reads is worth the extra Redis dependency, credentials, TLS configuration, monitoring, and HA operations.
+
+### No-Redis Deployment
+
+The default `chart/values.yaml` path keeps `redis-ha.enabled=false`, does not render Redis application env vars, does not render the Redis TLS sync CronJob, and does not require a `redis-password` secret key.
+
+```bash
+helm dependency update ./chart
+helm template statuslist ./chart --namespace statuslist
+```
+
+### Redis-Enabled Certificate Cache
+
+To enable Redis, use an application image built with `redis,aws,acme` features, set `redis-ha.enabled=true`, provide a `redis-password` in the configured secret or ExternalSecret, and keep `APP_REDIS__URI` unset in `statuslist.env` so the chart can generate it from the Redis HA service. Set `APP_REDIS__REQUIRE_CLIENT_AUTH` and Redis HAProxy TLS values only when your Redis endpoint requires them.
 
 ## Production Deployment Instructions
 
@@ -39,7 +59,7 @@ The following files are used to configure the deployment:
    kubectl create namespace statuslist
    ```
 
-2. **Create TLS secrets:**
+2. **Create TLS secrets, if Redis HAProxy TLS is enabled:**
 
    Refer to the [Redis TLS Setup Guide](../docs/REDIS_TLS_SETUP.md) for detailed instructions on creating the necessary TLS secrets for Redis and HAProxy.
 
--- a/helm/chart/templates/external-secrets.yaml
+++ b/helm/chart/templates/external-secrets.yaml
@@ -7,6 +7,8 @@ metadata:
     {{- include "status-list-server-chart.labels" . | nindent 4 }}
   namespace: {{ .Release.Namespace }}
 spec:
+  {{- $redisHA := index .Values "redis-ha" | default dict }}
+  {{- $redisHAEnabled := eq (default false (index $redisHA "enabled")) true }}
   refreshInterval: {{ .Values.externalSecret.spec.refreshInterval }}
   secretStoreRef:
     name: {{ .Values.externalSecret.spec.secretStoreRef.name }}
@@ -16,9 +18,11 @@ spec:
     creationPolicy: {{ .Values.externalSecret.spec.target.creationPolicy }}
   data:
   {{- range .Values.externalSecret.spec.data }}
+    {{- if or (ne .secretKey "redis-password") $redisHAEnabled }}
     - secretKey: {{ .secretKey }}
       remoteRef:
         key: {{ .remoteRef.key }}
         property: {{ .remoteRef.property }}
+    {{- end }}
   {{- end }}
 {{- end }}
--- a/src/config.rs
+++ b/src/config.rs
@@ -546,6 +546,8 @@ impl DnsConfig {
 
 #[derive(Debug, Clone, Deserialize)]
 pub struct RedisConfig {
+    /// Redis connection URI. Leave empty to disable the optional Redis-backed
+    /// certificate-material cache, even when the `redis` feature is compiled.
     pub uri: SecretString,
     pub require_client_auth: bool,
     /// Cache TTL for Redis TLS certificates in seconds.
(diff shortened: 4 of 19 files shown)


src/config.rs
```diff
@@ -785,7 +787,7 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("database.pool.connect_timeout_secs", 10u64)?
         .set_default("database.pool.idle_timeout_secs", 600u64)?
         .set_default("database.pool.max_lifetime_secs", 1800u64)?
-        .set_default("redis.uri", "redis://localhost:6379")?
+        .set_default("redis.uri", "")?
```
ndefokou, line 790: Empty default correctly disables Redis, but the consuming logic in setup.rs (`if let Ok(conn) = config.redis.start(...) else { None }`) swallows connection failures with no logging. If a deployment sets APP_REDIS__URI but Redis is unreachable, the cert cache silently degrades to direct S3 on every boot — invisible to operators. Please log a `tracing::warn!` when start() fails so the degradation is observable.

Related: `start(None, None, None)` never supplies cert/blob/root, so `require_client_auth=true` always returns Err ("Client authentication required but no certificates provided") and disables the cache. The mTLS path documented in redisTLS docs and helm/README is therefore inoperative via the main path. Either wire certs from config/env into the call, or drop the guidance to set APP_REDIS__REQUIRE_CLIENT_AUTH. Add a unit test asserting `start()` behavior with require_client_auth=true and no certs.

docker-compose.yml
```diff
@@ -108,13 +109,12 @@ services:
       APP_TELEMETRY__OTLP_ENDPOINT: "http://otel-collector:4317"
       APP_TELEMETRY__ENABLED: ${APP_TELEMETRY__ENABLED:-true}
       APP_TELEMETRY__SAMPLER_RATIO: ${APP_TELEMETRY__SAMPLER_RATIO:-1.0}
+      APP_REDIS__URI: ${APP_REDIS__URI:-}
```
ndefokou, line 112: The explicit `environment` override of APP_REDIS__URI neutralizes the stale value in `.env.template` (`APP_REDIS__URI=redis://redis:6379`) for compose runs, but that template was not audited by this PR. Any non-compose run that sources `.env.template` still defaults to a non-empty redis URI and blacks a nonexistent localhost Redis — directly contradicting the issue's "Remove any default that implies Redis is mandatory" and the deliverable to audit all Redis uses in docs/templates. Please set APP_REDIS__URI to empty and annotate the '# Redis configuration' block as optional.

Minor: under the `redis` profile there is no `depends_on` between the server and the redis service, so startup ordering isn't guaranteed. Harmless today (failure is non-fatal), but worth a comment or a profile-scoped dependency so the documented redis path is deterministic.

helm/README.md
```diff
@@ -30,6 +30,26 @@ The following files are used to configure the deployment:
...
+The default `chart/values.yaml` path keeps `redis-ha.enabled=false`, does not render Redis application env vars, does not render the Redis TLS sync CronJob, and does not require a `redis-password` secret key.
+
+```bash
+helm dependency update ./chart
+helm template statuslist ./chart --namespace statuslist
+```
+
+### Redis-Enabled Certificate Cache
+
+To enable Redis, use an application image built with `redis,aws,acme` features, set `redis-ha.enabled=true`, provide a `redis-password` in the configured secret or ExternalSecret, and keep `APP_REDIS__URI` unset in `statuslist.env` so the chart can generate it from the Redis HA service. Set `APP_REDIS__REQUIRE_CLIENT_AUTH` and Redis HAProxy TLS values only when your Redis endpoint requires them.
```
ndefokou, line 52: This instructs users to "Set APP_REDIS__REQUIRE_CLIENT_AUTH and Redis HAProxy TLS values only when your Redis endpoint requires them." However the app calls `config.redis.start(None, None, None)` and never supplies client certs/root at runtime, so enabling client auth makes `start()` return an error and the cache silently stays disabled. Either wire the certs (config/env -> start) so this documented path actually works, or remove this sentence to avoid documenting an inoperative configuration each the issue explicitly requires this path be "documented and tested".

helm/chart/templates/external-secrets.yaml
```diff
@@ -16,9 +18,11 @@ spec:
     creationPolicy: {{ .Values.externalSecret.spec.target.creationPolicy }}
   data:
   {{- range .Values.externalSecret.spec.data }}
+    {{- if or (ne .secretKey "redis-password") $redisHAEnabled }}
```
ndefokou, line 21: The redis-disabled filter only omits entries whose `secretKey` is exactly the string `redis-password`. If an operator names the key differently (e.g. `redis_password`, or already has it under another name), the no-Redis path still renders the remoteRef and breaks on the missing secret — exactly what this guard is meant to prevent. Consider making the skip key configurable (e.g. a `redis.secretKey` value) rather than a hard-coded literal.

ndefokou (commented):
Reviewed against the issue (make Redis optional / keep as distributed cert-material cache). Overall the direction is correct and the Helm gating is thorough. Full structured review below; inline comments on specific diffs.

## 1. High-level architectural review
The refactor is sound. Redis is confirmed off the hot status-list read path (that's `MokaStatusListCache`); Redis remains a `Storage` adapter wrapping S3 for certificate material. Moving the default `redis.uri` to empty and gating the Helm dependency on `redis-ha.enabled` is the right mechanism. The README/Helm/TLS docs now consistently describe Redis as optional.

## 2. Correctness review
- The empty-URI default is wired correctly in `setup.rs`: `if !config.redis.uri.expose_secret().is_empty() { if let Ok(conn) = start(...) { ... } else { None } }`. No-Redis builds compile and run without any Redis settings.
- **Silent degradation (see inline):** when `APP_REDIS__URI` is set but Redis is unreachable/misconfigured, `start()` returns `Err` and it is swallowed (`else { None }`) with no logging. A configured cache silently drops to direct S3 on every boot. Should `warn!`.
- **mTLS path is exercised yet dead (see inline):** `start(None, None, None)` never supplies certs, so `require_client_auth=true` always errors and disables the cache, contradicting the README/TLS docs.

## 3. Performance / scalability
No performance regression on the read path. Redis only affects cert-material storage wake-up at ACME-provision time. Fine. The 60s connection timeout in `start` is on a non-hot path, acceptable.

## 4. API contract review
No public API changes. Config surface changes (default `redis.uri`) are documented. Constant "Redis is not a database backend" line in `database-backends.md` is accurate.

## 5. Security / privacy
NetworkPolicy correctly drops Redis ports when disabled. ExternalSecret omits `redis-password` when disabled (good). One fragility: the filter only matches the literal `secretKey "redis-password"` (see inline).

## 6. Testing review
Only the config default test was updated. No test covers the behavior that actually matters: that an unreachable/misconfigured Redis is logged and degrades gracefully, or that the mTLS path's `start(None,None,None)` is broken. Recommend a test for `start()` with `require_client_auth=true` and no certs (it currently returns Err, silently). Also no helm-render test asserting zero redis objects with redis disabled (a `.github` action rendering both templates would harden the acceptance criterion).

## 7. Maintainability
The triple `index .Values "redis-ha" | default dict` + `default false ... eq ... true` boilerplate is repeated in 4+ templates. A single `{{- $redisHAEnabled := ... }}` shared helper (e.g. `include "status-list-server-chart.redisHAEnabled" .`) would reduce drift risk.

## 8. Edge cases / regressions
- **`.env.template` not audited (accepted-criteria gap)** — still defaults `APP_REDIS__URI=redis://redis:6379`; compose neutralizes it via the explicit `environment` override, but any non-compose run sourcing the template silently tries a nonexistent Redis. Deliverable "audit all Redis uses in docs" is incomplete.
- **Backward-compat default flip** — `redis-ha.enabled: true→false` plus removal of `wait-for-redis` initContainers is a silent behavior change for existing installs (they lose auto Redis URI/cache on upgrade). Intent is correct, but needs a release note / migration callout.
- **Compose ordering** — under the `redis` profile there is no `depends_on` between server and redis, so no startup ordering; harmless today (failure is non-fatal) but the README example implies Redis is used, and readiness would just never reflect it.
- No Redis feature-gating gaps: `outbound/redis.rs`, `health` wiring, and `RedisConfig::start` are all `#[cfg(feature="redis")]`; `redis.uri` remains parsed unconditionally (fine).

## 9. Blocking issues
None that prevent merge. The silent-degradation-no-log issue is the closest to blocking but is non-fatal (system still serves correctly). I'd require the `.env.template` fix before merge to actually claim the audit deliverable and to honor the issue's "no default implies Redis is mandatory" criterion.

## 10. Non-blocking suggestions
- Add `tracing::warn!` on Redis connect failure in `setup.rs`.
- Resolve the mTLS gap: either wire cert/key/root into `start` from config/env or drop the `APP_REDIS__REQUIRE_CLIENT_AUTH` guidance in docs.
- Update `.env.template` to empty `APP_REDIS__URI` and remove/annotate the "# Redis configuration" block.
- DRY the `$redisHAEnabled` computation into a shared helper.
- Bump chart version given dependency-condition + lock digest change.
- Add a helm-render smoke check (disabled + enabled) to CI.

## 11. Final assessment
Directionally excellent and largely correct. The Helm work is careful and complete; docs are accurate on the big picture. The main weaknesses are (1) unlogged silent cache degradation, (2) a documented-but-inoperative mTLS path, and (3) the incomplete `.env.template` audit. Address these and the PR meets the issue's acceptance criteria.

ndefokou (approved):

martcpp (changes requested):
require the .env.template fix before merge to actually claim the audit deliverable and to honor the issue's "no default implies Redis is mandatory" criterion.

martcpp (approved):

Hermann-Core:
Going to check this

Hermann-Core (approved):
@Christiantyemele Approved - verified optional Redis configuration defaults, tracing::warn! fallback logging, and concise documentation alignment.

PR #335, opened 2026-08-05
adding kubernetes readiness liveness probes and dependency aware heal…

### LINKED ISSUE #172: Add Kubernetes readiness/liveness probes and dependency-aware health checks
## Objective

Split the current `/health` endpoint into `/health/live` and `/health/ready`, and make readiness reflect the actual availability of Postgres, Redis/cache backend, and signing material.

## Context

The Helm deployment currently defines no probes. A readiness check should fail when a dependency the app cannot function without is down, so Kubernetes can stop routing traffic to the pod. This is especially important after introducing Redis-backed caches.

## Deliverables

- [x] Add `/health/live` and `/health/ready` endpoints.
- [x] Liveness: a lightweight check that the HTTP server is running.
- [x] Readiness checks for:
  - PostgreSQL connectivity (read query);
  - configured cache backend connectivity (in-memory always ready, Redis ping);
  - certificate/key store availability (can reach S3/secrets manager or local store).
- [x] Update `helm/chart/templates/deployment.yaml` with `livenessProbe` and `readinessProbe`.
- [x] Avoid loading/storing private keys in memory solely for the health check; only verify that the store is reachable.

## Acceptance Criteria

- `/health/ready` returns `200` only when all critical dependencies are reachable.
- `/health/live` returns `200` while the process is up.
- Kubernetes probes are configured in Helm.
- Tests verify failure of each readiness dependency.

ndefokou:
# PR: Split `/health` into `/health/live` and `/health/ready` with real dependency readiness

## Ticket / Motivation

The Helm deployment defines **no functional probes** — both `livenessProbe` and `readinessProbe` pointed at the old monolithic `/health` endpoint, which returned a static `"OK"` regardless of whether the service could actually serve traffic. After introducing Redis-backed caches, a readiness check must reflect the real availability of the dependencies the app cannot function without, so Kubernetes stops routing traffic to a pod whose Postgres, cache backend, or signing store is down.

**Partial fix (Helm-only) rejection:** Kubernetes has no built-in probe that can target an arbitrary dependency — `httpGet`/`tcpSocket`/`grpc` probes only reach the pod's own ports. A sidecar `exec` probe (e.g. `pg_isready`/`redis-cli`) is possible but cannot verify the S3/Secrets signing store, duplicates TLS/mTLS auth, and adds tooling overhead. The idiomatic and complete solution is an application readiness endpoint — which is what this PR delivers.

### 1. Split the health endpoint (`src/server/health.rs` — new module)
- **`GET /health/live`** — lightweight check that the HTTP process is up; returns `200 "OK"` and is deliberately detached from downstream dependencies so a healthy process is never restarted merely because a dependency is down.
- **`GET /health/ready`** — returns `200` **only when every critical dependency is reachable**, otherwise `503`, with a JSON body naming each dependency and the failure reason.
- Introduced a `ReadinessCheck` trait and a `Readiness` aggregator that runs all checks concurrently, matching the existing port/adapter pattern in `domain/ports.rs`.

### 2. Real dependency checks (no app can function without these)
Each check uses the **existing adapters** captured in the composition root (`setup.rs`) — no duplicated connection logic:
- **PostgreSQL**: `SELECT 1` via `SeaOrmStore::ping()` (new method), exercising a real connection round-trip. In-memory backend is always-ready.
- **Cache backend**: in-memory status-list cache is always-ready; the Redis cache is checked with `PING` via `Redis::ping()` (new method on the existing `ConnectionManager`).
- **Signing material store**: verifies reachability **without loading or parsing the private key**:
  - ACME/S3/Secrets backend → new `Storage::reachable()` (S3 `head_bucket`, Secrets Manager `list_secrets`, memory always-ready).
  - Filesystem store → checks the configured cert/key **files exist** (`metadata` only).

### 3. Plumbing
- Added `readiness: Readiness` to `AppState` and wired the probes into the router (`startup.rs`).
- `setup.rs` now retains the database `Arc<DatabaseConnection>` and Redis `ConnectionManager` needed by the probes and assembles the correct checkers per configured backend.

### 4. Helm probes
- `helm/chart/values.yaml`: `livenessProbe` → `/health/live`, `readinessProbe` → `/health/ready` (the template already supported rendering them).

### 5. Docs & load-test references to the removed endpoint
- `docs/openapi.yaml` — documents both new endpoints (incl. readiness response schema).
- `docs/LOCAL_DEPLOYMENT.md` — updated curl examples.
- Artillery `load-test.yml`, `spike-test.yml`, `stress-test.yml` — health check now uses `/health/live` (preserving their `200`/`"OK"`/`text/plain` assertions).

## Acceptance criteria
- ✅ `/health/ready` returns `200` only when all critical dependencies are reachable.
- ✅ `/health/live` returns `200` while the process is up.
- ✅ Kubernetes probes configured in Helm.
- ✅ Tests verify failure of each readiness dependency: database, `redis_cache`, and `cert_store` each flip `/health/ready` to `503` (`src/server/health.rs`).

## Testing
- `cargo test --lib` — **164 tests pass** (default features), including new handler + per-dependency failure tests.
- `cargo clippy --all-targets` — clean on default and full feature set.
- Build verified with `--features "acme,aws,postgres,sqlite,mysql,redis,memory"`.
- `helm template` renders `livenessProbe=/health/live` and `readinessProbe=/health/ready`.

## Files changed
```
M Cargo.toml / Cargo.lock                 (add `futures` dep)
A src/server/health.rs                    (ReadinessCheck, Readiness, live/ready handlers + tests)
M src/server/mod.rs                       (register health module, add readiness to AppState)
M src/startup.rs                          (register /health/live, /health/ready)
M src/setup.rs                            (capture handles, assemble probes)
M src/outbound/sql/store.rs               (SeaOrmStore::ping)
M src/outbound/redis.rs                   (Redis::ping, connection accessor)
M src/outbound/aws.rs                     (Storage::reachable for S3 / Secrets Manager)
M src/utils/cert_manager/storage.rs       (Storage::reachable trait + default/box impl)
M src/test_utils.rs                       (add readiness field)
M helm/chart/values.yaml                  (probe paths)
M docs/openapi.yaml, docs/LOCAL_DEPLOYMENT.md (endpoint docs)
M artillery-tests/tests/*.yml             (health URL → /health/live)
```


Ngha-Boris (commented):
@ndefokou please consider the comments below

Cargo.toml
```diff
@@ -58,6 +58,7 @@ opt-level = 1
 # Async runtime and utilities
 tokio = { version = "1.5", features = ["full"] }
 async-trait = "0.1"
+futures = "0.3"
```
Ngha-Boris, line 62: **Issue**: Unnecessary `futures` crate dependency

The `futures` crate is added solely for `futures::future::join_all`. Tokio already provides `tokio::join!` macro and `JoinSet` for concurrent execution.

**Suggestion**: Remove this dependency and use Tokio's built-in utilities:
```rust
// Alternative using tokio::join! for a small number of checks
let (db_result, cache_result, cert_result) = tokio::join!(
    db_check.check(),
    cache_check.check(), 
    cert_check.check()
);
```

Or use `futures-lite` if you need a lightweight `join_all` alternative.

helm/chart/values.yaml
```diff
@@ -103,15 +103,15 @@ statuslist:
     APP_TELEMETRY__SAMPLER_RATIO: "1.0"
   livenessProbe:
     httpGet:
-      path: /health
+      path: /health/live
       port: http
     initialDelaySeconds: 10
```
Ngha-Boris, line 88: **Issue**: Missing `failureThreshold` in readinessProbe

The readiness probe is missing the `failureThreshold` field that is present in the livenessProbe. While Kubernetes uses a default of 3, it's better to be explicit and consistent.

**Suggestion**: Add `failureThreshold: 3` to match the livenessProbe configuration:
```yaml
readinessProbe:
  httpGet:
    path: /health/ready
    port: http
  initialDelaySeconds: 5
  periodSeconds: 10
  timeoutSeconds: 5
  failureThreshold: 3  # Add this line
```

src/outbound/aws.rs
```diff
@@ -133,6 +133,18 @@ impl Storage for AwsSecretsManager {
         }
         Ok(())
     }
+
+    /// Verify the Secrets Manager API is reachable without touching any
+    /// specific secret, by issuing a paginated `MaxResults(1)` list call.
+    async fn reachable(&self) -> Result<(), StorageError> {
+        self.client
+            .list_secrets()
```
Ngha-Boris: **Issue**: Secrets Manager `reachable()` requires broad IAM permissions

Using `list_secrets()` requires `secretsmanager:ListSecrets` permission, which may be broader than necessary for a health check.

**Suggestion**: Consider using `describe_secret` on a known non-existent secret name instead:

```rust
async fn reachable(&self) -> Result<(), StorageError> {
    use aws_sdk_secretsmanager::error::SdkError;
    
    match self.client
        .describe_secret()
        .secret_id("__health_probe_test__")
        .send()
        .await 
    {
        Ok(_) => Ok(()),
        Err(SdkError::ServiceError(err)) if err.err().is_resource_not_found_exception() => {
            Ok(()) // API is reachable, secret just doesn't exist
        }
        Err(e) => Err(StorageError::AwsSdk(e.into())),
    }
}
```

This only requires `secretsmanager:DescribeSecret` permission.

src/outbound/aws.rs
```diff
@@ -133,6 +133,18 @@ impl Storage for AwsSecretsManager {
...
+            .max_results(1)
+            .send()
+            .await
+            .map_err(|e| StorageError::AwsSdk(e.into()))?;
+        Ok(())
+    }
 }
```
Ngha-Boris, line 159: **Issue**: S3 `reachable()` has side effects - it can CREATE the bucket

The `reachable()` implementation calls `ensure_bucket_exists()` which has a side effect: if the bucket doesn't exist, it will **create** it. This is unexpected behavior for a health check.

**Suggestion**: Create a proper `reachable()` method that only checks without side effects:

```rust
async fn reachable(&self) -> Result<(), StorageError> {
    use aws_sdk_s3::error::SdkError;
    
    // Simple HEAD check without creation side effects
    self.client
        .head_bucket()
        .bucket(&self.bucket)
        .send()
        .await
        .map(|_| ())
        .map_err(|e| match e {
            SdkError::ServiceError(err) => {
                StorageError::AwsSdk(eyre!("S3 bucket unavailable: {}", err.err()))
            }
            other => StorageError::AwsSdk(other.into()),
        })
}
```

src/outbound/redis.rs
```diff
@@ -17,6 +17,21 @@ impl Redis {
         Self { conn, ttl: None }
     }
 
+    /// Return a clone of the underlying connection manager, for use where a
+    /// separate handle on the same backend is needed (e.g. readiness probing).
+    pub fn connection(&self) -> ConnectionManager {
+        self.conn.clone()
+    }
+
```
Ngha-Boris, line 25: **Suggestion**: The `ping()` method clones the connection for each call. Since `ConnectionManager` is already cloneable and designed for concurrent use, consider whether this clone is necessary or if you can use `&mut self.conn` directly.

Also, consider adding a timeout wrapper in the `RedisCheck` to prevent indefinite hangs if Redis becomes unresponsive.

src/outbound/sql/store.rs
```diff
@@ -25,6 +25,16 @@ impl<T> SeaOrmStore<T> {
...
     }
+
+    /// Verifies the database is reachable by issuing a lightweight read query.
+    ///
+    /// Used by the `/health/ready` probe. Matching the existing `SELECT 1`
+    /// pattern used in the pool-timeout test, this exercises a real connection
+    /// round-trip so a downed database surfaces as a readiness failure.
+    pub async fn ping(&self) -> Result<(), sea_orm::DbErr> {
+        use sea_orm::ConnectionTrait as _;
+        (*self.db).execute_unprepared("SELECT 1").await.map(|_| ())
```
Hermann-Core: There is an existing `ping` method in sea-orm library, you should use that one instead of a custom one.

Hermann-Core (dismissed):

src/server/health.rs
```diff
@@ -0,0 +1,394 @@
...
+            set.spawn(async move {
+                let res = tokio::time::timeout(Duration::from_secs(5), check.check())
+                    .await
+                    .unwrap_or_else(|_| Err("check timed out".to_string()));
+                (check, res)
+            });
+        }
+
+        let mut checks: Vec<CheckResult> = Vec::with_capacity(self.checks.len());
+        while let Some(Ok((check, res))) = set.join_next().await {
```
Hermann-Core: **Bug:** `while let Some(Ok(...))` drops any task that panics or returns `Err` from `join_next`. A failed join omits that dependency from `checks` entirely, so `ready` can become `true` even though not every registered check ran.

Handle `join_next()` errors explicitly (treat as `ok: false` with a reason) and assert `checks.len() == self.checks.len()` before computing `ready`, or switch to a join helper that cannot lose entries.

src/server/health.rs
```diff
@@ -0,0 +1,394 @@
...
+
+#[cfg(any(feature = "sqlite", feature = "postgres", feature = "mysql"))]
+#[async_trait::async_trait]
+impl ReadinessCheck for DbCheck {
+    fn name(&self) -> &'static str {
+        "database"
+    }
+
+    async fn check(&self) -> Result<(), String> {
+        crate::outbound::sql::SeaOrmStore::<()>::new(self.db.clone())
```
Hermann-Core: Constructing `SeaOrmStore::<()>` just to ping is structural drift: this is connection health, not entity-store behaviour. `DbCheck` already holds `Arc<DatabaseConnection>` — call `self.db.ping().await` directly (as `SeaOrmStore::ping` already delegates). Drop the unit-type store wrapper here; if nothing else needs `SeaOrmStore::ping`, remove that method rather than parking connection probes on the generic store.

src/setup.rs
```diff
@@ -351,6 +366,46 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+            None => readiness = readiness.with_check(AlwaysReady::new("database")),
+        }
+    }
+    #[cfg(not(any(feature = "sqlite", feature = "postgres", feature = "mysql")))]
+    {
+        readiness = readiness.with_check(AlwaysReady::new("database"));
+    }
+
+    // The status-list cache is always in-process (Moka); it is always ready.
+    readiness = readiness.with_check(AlwaysReady::new("cache"));
```
Hermann-Core: `AlwaysReady("cache")` is a permanent no-op that always appears green in the readiness JSON beside `"redis_cache"`. Issue #172 asks to treat the in-memory cache as implicitly ready, not to emit a fake dependency row. Either omit the Moka cache from the report, or rename consistently (e.g. `status_list_cache` + `cert_redis_cache`) so operators can tell which backend failed.

src/outbound/redis.rs
```diff
@@ -17,6 +17,21 @@ impl Redis {
         Self { conn, ttl: None }
     }
 
+    /// Return a clone of the underlying connection manager, for use where a
+    /// separate handle on the same backend is needed (e.g. readiness probing).
+    pub fn connection(&self) -> ConnectionManager {
```
Hermann-Core: `connection()` is unused — `setup.rs` passes the `ConnectionManager` straight into `RedisCheck`, and `ping()` clones internally. Remove the dead accessor unless a caller is planned; it adds surface area without earning a use site.

src/server/health.rs
```diff
@@ -0,0 +1,394 @@
...
+#[cfg(feature = "redis")]
+#[async_trait::async_trait]
+impl ReadinessCheck for RedisCheck {
+    fn name(&self) -> &'static str {
+        "redis_cache"
+    }
+
+    async fn check(&self) -> Result<(), String> {
+        let conn = self.conn.clone();
+        tokio::time::timeout(Duration::from_secs(5), async move {
```
Hermann-Core: Double timeout: `Readiness::run` already wraps every check in 5s, so the inner `tokio::time::timeout` here is redundant. Keep the timeout at the aggregator boundary only — that matches a single probe budget and avoids duplicating policy in each adapter.

Hermann-Core (changes requested):

src/utils/cert_manager/storage.rs
```diff
@@ -49,6 +57,9 @@ impl<T: Storage + ?Sized> Storage for Box<T> {
     async fn delete(&self, key: &str) -> Result<(), StorageError> {
         (**self).delete(key).await
     }
+    async fn reachable(&self) -> Result<(), StorageError> {
```
Hermann-Core, line 61: Add a space between these 2 blocks of code (between lines `59` and `60`).

src/setup.rs
```diff
@@ -218,7 +228,12 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
                     let cache_opt = if !config.redis.uri.expose_secret().is_empty() {
                         match config.redis.start(None, None, None).await {
                             Ok(redis_conn) => {
-                                Some(Redis::new(redis_conn).with_ttl(config.redis.cert_cache_ttl))
+                                redis_manager = Some(redis_conn);
+                                let redis_conn = redis_manager.as_ref().expect("just set");
```
Hermann-Core: Avoid panicking code and return or propagate error directly.

Hermann-Core (changes requested):

Ngha-Boris:
Hello @Hermann-Core please re-check

Hermann-Core (dismissed):
@Ngha-Boris all prior review threads are correctly addressed.

Non-blocking notes for operators
- `/health` is removed (breaking for external monitors not yet on Helm); migration is live vs ready split
- When Redis URI is empty or startup connection fails, no `redis_cache` row is registered — readiness stays green by design (graceful degradation)
- `cert_manager_opt.unwrap()` in the acme return path is unreachable with current wiring; could be flattened later for clarity

Issue #172 acceptance criteria met. LGTM.

Ngha-Boris (dismissed):

Hermann-Core:
@forkimenjeckayang can you take a look at this so we can merge?

forkimenjeckayang:
currently checking

src/server/health.rs
```diff
@@ -0,0 +1,435 @@
...
+    async fn check(&self) -> Result<(), String> {
+        let cert_store = self
+            .manager
+            .cert_storage()
+            .map_err(|e| format!("cert storage not configured: {e}"))?;
+        crate::cert_manager::storage::Storage::reachable(cert_store)
+            .await
+            .map_err(|e| format!("cert store unreachable: {e}"))
+    }
+}
```
forkimenjeckayang, line 248: > This checks only `CertManager::cert_storage()`. In the production `postgres,aws,acme` path, `setup.rs` injects S3 as `cert_storage` and AWS Secrets Manager as the separate `secrets_storage`. Status-list token generation calls `signing_key_pem()`, which reads `secrets_storage` at runtime, so a Secrets Manager outage can leave `/health/ready` at 200 while JWT/CWT generation fails. The new `AwsSecretsManager::reachable()` implementation is currently never reached by this checker.
>

Please verify both certificate and secrets storage under the existing outer timeout, and add a regression test where certificate storage is reachable but secrets storage is not and `/health/ready` returns 503. The check should remain reachability-only as required by the ticket; it does not need to load private-key material.

src/setup.rs
```diff
@@ -357,6 +368,43 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+    #[cfg(not(any(feature = "sqlite", feature = "postgres", feature = "mysql")))]
+    {
+        readiness = readiness.with_check(AlwaysReady::new("database"));
+    }
+
+    #[cfg(feature = "redis")]
+    {
+        if let Some(conn) = redis_manager {
+            readiness = readiness.with_check(crate::server::health::RedisCheck::new(conn));
+        }
```
forkimenjeckayang: > Redis is still implemented and documented as an optional certificate-material cache. Startup catches a Redis connection failure and falls back to direct S3, and `AwsS3::load()` also logs Redis cache errors and continues to S3. But if Redis connected at startup and becomes unavailable later, this check returns 503 and Kubernetes removes the otherwise functional pod from service.
>

 Please make the dependency semantics consistent. If Redis remains an optional optimization, omit it from readiness or report it as degraded without gating `ready`. If configured Redis is intended to be critical, startup and the S3 adapter must stop silently falling back. Add a runtime-outage test proving the selected policy. The ticket's acceptance criterion says readiness gates dependencies the app cannot function without, which the current Redis path is not.

src/server/health.rs
```diff
@@ -0,0 +1,435 @@
...
+/// Readiness: `200` only when every critical dependency is reachable, else
+/// `503`. Returns a JSON body naming each dependency so humans and CI can see
+/// exactly which one is failing.
+pub async fn ready(State(state): State<AppState>) -> Response {
+    let report = state.readiness.run().await;
+    if report.ready {
+        (StatusCode::OK, Json(report)).into_response()
+    } else {
+        (StatusCode::SERVICE_UNAVAILABLE, Json(report)).into_response()
+    }
```
forkimenjeckayang, line 156: > This endpoint is outside the API governor/auth middleware, and the default Helm ingress exposes `/` with `pathType: Prefix`, so `/health/ready` is publicly callable. Every request spawns live DB, Redis, and S3 checks, and the response returns raw backend error strings. An external caller can therefore amplify requests into internal dependencies, consume the small DB pool, generate AWS API traffic, and learn internal failure details.
>

Kubernetes needs an unauthenticated pod-local probe, but it does not require that probe to be publicly exposed or to execute a fresh fan-out for every internet request. Please isolate health endpoints on a management port not routed by ingress, or otherwise coalesce/cache and bound the checks while returning only generic public status. Log detailed reasons internally.

src/startup.rs
```diff
@@ -73,7 +70,8 @@ impl HttpServer {
 
         let mut router = Router::new()
             .route("/", get(welcome))
-            .route("/health", get(health_check))
+            .route("/health/live", get(health::live))
+            .route("/health/ready", get(health::ready))
```
forkimenjeckayang, line 75: > The split removes `/health`, but repository-owned consumers still call it: all four `demo/workflows/*.ipynb` notebooks and `docs/blog/status-list-server.adoc` reference `GET /health`. Those examples now receive 404 even though the PR updates the load tests and local deployment guide.
>
Please update all repository references, or retain `/health` temporarily as a documented liveness alias if backward compatibility is desired. A repository-wide route-reference check would keep the endpoint migration complete.

forkimenjeckayang (changes requested):
a few comments
Please check when free

Ngha-Boris:
Hello @forkimenjeckayang please take a look again

Ngha-Boris (dismissed):

forkimenjeckayang (dismissed):
LGTM
But please check CI faulre

Hermann-Core (approved):

Ngha-Boris (approved):

forkimenjeckayang (approved):

PR #340, opened 2026-08-06
Implement SQL-backed certificate storage with SeaORM integration

### LINKED ISSUE #307: Add SQL-backed certificate storage
## Objective

Add a SQL-backed storage implementation for certificate chains as a portable fallback where object storage is not available.

## Context

The server already supports PostgreSQL, MySQL, and SQLite through SeaORM. A SQL-backed implementation of the certificate-manager `Storage` trait would give operators a simple provider-neutral option, especially for smaller deployments or environments where adding separate object storage is unnecessary.

## Deliverables

* [x] Design a small table schema for certificate storage with certificate name, value, timestamps, and optional metadata.
* [x] Implement a SeaORM-backed `Storage` adapter for certificates.
* [x] Add migrations for supported SQL backends.
* [x] Document when SQL-backed certificate storage is appropriate.

## Acceptance Criteria

* [x] SQL storage can be selected for certificate storage.
* [x] The adapter works with PostgreSQL, MySQL, and SQLite test configurations where those features are enabled.
* [x] Deletion and update semantics match the `Storage` trait contract.

## Verification

* [x] SeaORM unit/integration tests for each enabled backend.
* [x] `cargo nextest run --workspace --all-targets --features sqlite,acme`

**Estimate:** **1 day**

Ngha-Boris:
Implemented and verified the SQL-backed certificate storage path.

What’s in place:
- SeaORM certificate storage adapter: `certificate_storage.rs`
- `certificate_storage` migration/table schema with name, value, timestamps, optional metadata: `migrations.rs`
- Config selection via `server.cert.store.source = "sql"` in startup wiring: `setup.rs`
- Operator docs in `README.md` and `database-backends.md`
- Fixed SeaORM entity visibility in `models.rs` so the `sqlite,acme` build compiles cleanly around the generated entity API.

Verification:
- `cargo nextest run --workspace --all-targets --features sqlite,acme` all 247 test passed
- Targeted SQL cert storage and startup selection tests also passed.


src/setup.rs
```diff
@@ -421,6 +363,110 @@ fn empty_to_none(value: Option<String>) -> Option<String> {
...
+    if config.server.cert.store.source.eq_ignore_ascii_case("sql") {
+        let db = sql_db.ok_or_else(|| {
+            eyre!(
+                "server.cert.store.source = 'sql' requires a SQL database backend; \
+                 configure database.backend as postgres, mysql, or sqlite"
+            )
+        })?;
+        return Ok((
+            Box::new(SqlCertificateStorage::new(db)),
+            Box::new(MemoryStorage::default()),
```
Christiantyemele, line 383: Bug: the `sql` source wires `secrets_storage` to a fresh `MemoryStorage::default()`. The CertManager routes the **signing key** (`keys-<domain>` via `signing_key_from_storage()` / `persist_signing_key()` / `signing_key_pem()`) and ACME account credentials to `secrets_storage`, not `cert_storage`. Meanwhile `store_certificate_strategy` maps `sql` to `StoreProvisioningStrategy::storage(...)` (line 524), whose `load_material` reads the signing key from `secrets_storage.load(signing_key_key)` (strategy.rs:141). With an empty in-memory store, store provisioning fails on startup with "store signing key key '...' was not found", and even if it succeeded the signing key would not persist across restarts (regenerated each start, invalidating previously issued token signatures). The signing key must be persisted in SQL too (or a dedicated secrets table), not `MemoryStorage`. Suggested fix:

```rust
return Ok((
    Box::new(SqlCertificateStorage::new(db.clone())),
    Box::new(SqlCertificateStorage::new(db)), // persist signing key in SQL as well
));
```

src/setup.rs
```diff
@@ -680,6 +753,40 @@ mod general_tests {
         });
     }
 
+    #[cfg(all(feature = "acme", feature = "sqlite"))]
+    #[tokio::test]
```
Christiantyemele, line 757: Suggestion: `build_state_selects_sql_certificate_storage` only exercises `persist_certificate_data` and `certificate()`, both of which hit `cert_storage` (SQL). It never drives `StoreProvisioningStrategy::provision()` or the signing-key path (`signing_key_from_storage`/`persist_signing_key`), so it cannot catch the signing-key-routing problem above. Consider extending it to run the full store-provisioning flow (or assert signing-key persistence/survival across a simulated restart) so the SQL source is actually end-to-end verified.

src/outbound/sql/certificate_storage.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[tracing::instrument(skip(self), fields(db.system = "sea-orm"))]
+    async fn load(&self, key: &str) -> Result<Option<String>, StorageError> {
+        let row = certificate_storage::Entity::find_by_id(key)
+            .one(&*self.db)
+            .await?;
+        Ok(row.map(|row| row.value))
+    }
+
+    #[tracing::instrument(skip(self, value), fields(db.system = "sea-orm"))]
+    async fn update(&self, key: &str, value: &str) -> Result<(), StorageError> {
```
Christiantyemele: Nit: the `update` override only delegates to `store`, which is exactly what the `Storage` trait provides as a default method (storage.rs:37-39). The override is redundant. If the intent is to document that both `store`/`update` are upserts, prefer dropping the override and relying on the trait default, or add a `#[doc]` note on the impl rather than reimplementing it.

Christiantyemele (changes requested):
.

Ngha-Boris:
Hello @Christiantyemele please check again

src/setup.rs
```diff
@@ -421,6 +363,110 @@ fn empty_to_none(value: Option<String>) -> Option<String> {
...
+    if config.server.cert.store.source.eq_ignore_ascii_case("sql") {
+        let db = sql_db.ok_or_else(|| {
+            eyre!(
+                "server.cert.store.source = 'sql' requires a SQL database backend; \
+                 configure database.backend as postgres, mysql, or sqlite"
+            )
+        })?;
+        return Ok((
+            Box::new(SqlCertificateStorage::new(db.clone())),
+            Box::new(SqlCertificateStorage::new(db)),
```
Christiantyemele, line 382: Important (security): binding `secrets_storage` to `SqlCertificateStorage` means the TLS/signing private key (`keys-<domain>` from `signing_key_pem()`/`persist_signing_key`) and the ACME account credentials (`acme_accounts-<domain>`, which include the ACME account private key) are now stored verbatim as plaintext in the `certificate_storage.value` column (`certificate_storage.rs:37-45` writes the raw string, no encryption at rest). The certificate chain itself is public material, but the private key is not — a DB read, a backup leak, or a dump of this table would expose the key that signs status list tokens. This is consistent with the existing plaintext `storage`/S3 and `filesystem` backends and the docs already steer high-scale/production toward dedicated secret stores, but because it now lives in the same application DB as all other data, it materially widens exposure. Recommendation: at minimum document explicitly on the `sql` source that the private signing key is stored unencrypted in the `certificate_storage` table and must be protected at the DB layer (encrypted volumes, restricted access, careful backup handling); ideally consider limiting `sql` to certificate-chain storage and keep the signing key in a dedicated secrets backend.

Christiantyemele (changes requested):
..

src/setup.rs
Ngha-Boris: @Christiantyemele thanks, you're right, and I think the cleaner resolution is to stay within the ticket's actual scope.

The ticket asks for SQL storage of the **certificate chain**, not secrets. My earlier fix wired `secrets_storage` to `SqlCertificateStorage` too, which is exactly what your security comment flags — it pulls the private signing key and ACME account credentials into the same plaintext table. Those two concerns are two sides of the same root issue: the `sql` source shouldn't be backing `secrets_storage` at all.

Rather than keep the signing key in SQL (persistence) and accept plaintext exposure, I'm reverting to the ticket-faithful design: `sql` stores **certificate chains only** (`cert_storage` → SQL), and `secrets_storage` stays on its dedicated secrets backend. That keeps the implementation within scope, removes the plaintext private-key exposure, and makes the earlier persistence concern moot. I'll drop the restart-persistence test that relied on the secrets-in-SQL assumption and keep the small `update` cleanup.

src/setup.rs
Ngha-Boris: See comment https://github.com/adorsys/status-list-server/pull/340#discussion_r3760512269

Hermann-Core:
We are going to move toward a consolidated cryptographic-material backend by default: the same selected secrets/material backend should manage both the signing key and the certificate chain associated with it. Closing this now.

Issue #249, opened 2026-07-21
Support additional secrets-storage backends (HashiCorp Vault, GCP Secret Manager, Azure Key Vault)


Hermann-Core:
## Problem

The certificate manager's `Storage` trait (`src/utils/cert_manager/storage.rs`) is the abstraction used for **both** certificate blob storage and signing-key/secret storage — `AwsS3` and `AwsSecretsManager` are its only two implementations today (`src/utils/cert_manager/storage/aws.rs`), and both are constructed unconditionally in `src/utils/state.rs` using AWS SDK config. There is no way to run this server against a non-AWS secrets backend, which is a hard blocker for any deployment on GCP, Azure, on-prem/Kubernetes-native (Vault), or any multi-cloud/vendor-neutral setup.

This is a gap in the infrastructure abstraction work tracked by #166: that EPIC's checklist lists `SecretsBackend` / env-based and in-memory implementations as done (`[x]`), but in the current codebase only the two AWS-backed implementations of `Storage` exist — there is no `EnvSecretsBackend`, no `InMemoryKeyStore`, and no non-AWS production backend. This ticket specifically covers the **production, non-AWS secrets-manager backends** most commonly requested/used in the wild; the env/in-memory dev backends are covered by #166's own remaining scope.

## Proposal: model this on the existing DNS-provider abstraction

The codebase already has a mature, proven pattern for exactly this kind of "pick one of N vendor backends at runtime" problem: `DnsProvider` (`src/utils/cert_manager/challenge/dns01/`), documented end-to-end in `docs/dns-providers.md`, with 6 implementations (Route53, Cloudflare, GCloud DNS, Azure DNS, ACME-DNS, Pebble) selected via `APP_SERVER__CERT__DNS__PROVIDER`.

Apply the same shape to secrets storage:

1. Keep (or extract a more precisely-named) `Storage`/`SecretsBackend` trait — the existing `Storage` trait in `storage.rs` already has the right shape (`store`/`load`/`update`/`delete`), so this is primarily about adding implementations, not redesigning the trait.
2. Add new implementations under `src/utils/cert_manager/storage/`:
   - `vault.rs` — HashiCorp Vault (KV v2 secrets engine, token or AppRole auth).
   - `gcp.rs` — GCP Secret Manager.
   - `azure.rs` — Azure Key Vault.
[...]

PR #343, opened 2026-08-07
feat(secrets): implement Vault & OpenBao secrets backend (#312, #249)

### LINKED ISSUE #312: Implement Vault and OpenBao secrets backend for signing materials
## Objective

Implement HashiCorp Vault and OpenBao support for signing-key/secret storage.

## Context

Issue #249 tracks additional secrets backends broadly. This sub-issue carves out the Vault/OpenBao slice because it is valuable for Kubernetes-native and on-prem deployments and can be implemented independently from cloud-vendor secret managers.

## Deliverables

- [x] Implement a Vault/OpenBao KV v2 `Storage` adapter.
- [x] Support token auth first; document AppRole/Kubernetes auth as follow-up or add if straightforward.
- [x] Add configurable mount path, secret path prefix, namespace, and timeout.
- [x] Add optional in-process cache using TTL semantics aligned with `aws.secrets_cache_ttl`.
- [x] Add tests.
- [x] Add docs with Vault and OpenBao examples.

## Acceptance Criteria

- [x] Vault and OpenBao can store and retrieve signing key material through the existing certificate manager.
- [x] Cache can be disabled with TTL `0`.
- [x] Missing/forbidden secrets produce clear startup/runtime errors.
- [x] No secret values appear in logs or error messages.

## Verification

- [x] Adapter tests

Estimated time: 2 days

Hermann-Core:
Implement HashiCorp Vault and OpenBao support for signing-key/secret storage.

## Acceptance Criteria

- [x] Vault and OpenBao can store and retrieve signing key material through the existing certificate manager.
- [x] Cache can be disabled with TTL `0`.
- [x] Missing/forbidden secrets produce clear startup/runtime errors.
- [x] No secret values appear in logs or error messages.


Christiantyemele (commented):
placeholder

Christiantyemele (commented):
## Summary

This PR implements Vault & OpenBao KV v2 as a secrets backend for signing-key storage, refactors `StorageError::AwsSdk` to `StorageError::Backend`, removes the `sealed_test` dependency, and adds integration tests for both Vault and OpenBao. The implementation is generally well-structured: token auth is handled via `SecretString`, cache TTL respects the 0=disabled contract, and error responses do not leak secret values.

### Findings by severity
- **Suggestion**: 5 — key URL-encoding, 403 handling consistency, cache capacity, missing unit tests, feature coupling
- **Nit**: 2 — doc comment formatting
- **Bug**: 0
- **Important**: 0

### Key themes
- **Spec compliance**: Vault KV v2 HTTP API is correctly implemented (data/metadata endpoints, X-Vault-Token, X-Vault-Namespace headers)
- **Security**: Token is handled via `SecretString` and never logged; error messages omit secret values. No `unwrap()` in production paths.
- **Code quality**: Clean builder pattern, consistent error mapping, all existing tests pass.
- **Gaps**: No unit tests for the vault adapter (only Docker-based integration tests); keys are not URL-encoded before being placed in request paths.

src/outbound/vault.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    /// Build the full URL for KV v2 `metadata` (used by hard-delete).
+    fn metadata_url(&self, key: &str) -> String {
+        let path = self.qualify_key(key);
+        format!("{}/v1/{}/metadata/{}", self.addr, self.mount, path)
+    }
+
+    /// Qualify a key by prepending the configured path prefix.
+    fn qualify_key(&self, key: &str) -> String {
```
Christiantyemele, line 61: **Suggestion**: Keys are interpolated directly into URL path segments without URL-encoding. If a key contains characters like `?`, `#`, or spaces, the HTTP request URL will be malformed. Consider percent-encoding each path segment — e.g. via `url::form_urlencoded::Serializer` or a manual `replace` for `'?'`, `'#'`, and `' '` — or at minimum document that keys must be URL-safe.

src/outbound/vault.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        let status = resp.status();
+        if status.is_success() {
+            // Update cache on successful write
+            if let Some(cache) = &self.cache {
+                cache.insert(key.to_string(), value.to_string()).await;
+            }
+            Ok(())
+        } else {
+            Err(StorageError::Backend(eyre!(
```
Christiantyemele, line 230: **Suggestion**: `load()` (line 261) handles HTTP 403 with a specific error message (`"vault access denied for path '{}'"`), but `store()` and `delete()` lump 403 into the generic `"vault store/delete failed"` message. Add a specific 403 branch to these methods for consistency and debuggability.

src/outbound/vault.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// Prefix prepended to all secret paths.
+    path_prefix: String,
+    /// Optional Vault Enterprise / OpenBao namespace.
+    namespace: Option<String>,
+    /// Optional in-process TTL cache.
+    cache: Option<Cache<String, String>>,
+}
+
+impl VaultClient {
+    const CACHE_MAX_CAPACITY: u64 = 10;
```
Christiantyemele, line 40: **Suggestion**: `CACHE_MAX_CAPACITY = 10` is small for multi-domain deployments. With 20+ domains, every other read would miss the cache. Consider increasing this or documenting the rationale.

src/outbound/vault.rs
```diff
@@ -0,0 +1,299 @@
+//! Vault / OpenBao KV v2 secrets-storage adapter implementing [`Storage`].
+//!
+//! Both HashiCorp Vault and OpenBao expose the same KV v2 HTTP API, so this
+//! single adapter covers both.  Only **token auth** is supported in this
+//! initial implementation;
```
Christiantyemele, line 5: **Suggestion**: The module doc string ends with a semicolon (`;`) instead of a period.

Cargo.toml
```diff
@@ -8,11 +8,11 @@ default = ["memory"]
...
 server = ["history"]
-postgres = ["sea-orm?/sqlx-postgres", "history"]
-sqlite = ["sea-orm?/sqlx-sqlite", "history"]
-mysql = ["sea-orm?/sqlx-mysql", "history"]
-# Optional distributed certificate-material cache driver.
+postgres = ["history", "sea-orm?/sqlx-postgres"]
+sqlite = ["history", "sea-orm?/sqlx-sqlite"]
+mysql = ["history", "sea-orm?/sqlx-mysql"]
 redis = ["dep:redis"]
+vault = ["acme"]
```
Christiantyemele, line 15: **Suggestion**: `vault = ["acme"]` forces the ACME feature whenever vault is enabled. The `VaultClient` module itself has no ACME dependency — the coupling is only in `setup.rs` where the vault backend is constructed inside `#[cfg(feature = "acme")]`. Consider whether Vault might be used for non-ACME secret storage in the future; this may need refactoring to decouple the features.

src/config.rs
```diff
@@ -599,6 +600,26 @@ pub struct AwsConfig {
     pub s3_key_prefix: String,
 }
 
+#[derive(Debug, Clone, Deserialize)]
+pub struct VaultConfig {
+    /// Vault / OpenBao API address (e.g. `http://vault:8200`).
+    pub addr: String,
+    /// Authentication token.
+    pub token: SecretString,
+    /// KV v2 engine mount path..
```
Christiantyemele, line 609: **Nit**: Double period in doc comment — `"mount path.."` should be `"mount path."`.

src/outbound/vault.rs
```diff
@@ -0,0 +1,299 @@
...
+            .send()
+            .await?;
+
+        let status = resp.status();
+        if status.is_success() || status == StatusCode::NOT_FOUND {
+            if let Some(cache) = &self.cache {
+                cache.invalidate(key).await;
+            }
+            Ok(())
+        } else {
```
Christiantyemele, line 286: **Suggestion**: Same as the `store()` comment — `delete()` also lacks a distinct 403 error message.

Cargo.toml
Hermann-Core: Valid point — the coupling exists because `VaultClient` is currently constructed only inside the `#[cfg(feature = "acme")]` cert-manager setup block in `setup.rs`. Enabling `vault` without `acme` would compile the adapter but never wire it, making it a dead feature.

When vault is extended to cover non-ACME secrets (e.g. database credentials), the structural refactoring of `setup.rs` to lift `VaultClient` outside the ACME gate will be deferred to that implementation.

Hermann-Core:
@Christiantyemele thanks for the reviews left here, comments have been addressed can you recheck?

src/setup.rs
```diff
@@ -198,17 +204,56 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
         let app_env = std::env::var("APP_ENV").unwrap_or(ENV_DEVELOPMENT.to_string());
         let cert_domains = [config.server.domain.as_str()];
-        let (cert_storage, secrets_storage): (Box<dyn Storage>, Box<dyn Storage>) = {
-            #[cfg(feature = "aws")]
+        let secrets_storage: Box<dyn Storage> = {
+            #[cfg(feature = "vault")]
+            {
+                tracing::info!("Using Vault KV v2 as secrets backend");
+                Box::new(
+                    VaultClient::builder(&config.vault.addr, config.vault.token.clone())
```
Ngha-Boris, line 226: Vault is selected unconditionally whenever the vault feature is compiled. That makes backend choice depend on build flags instead of runtime config, and it overrides AWS Secrets Manager or memory secrets storage even when server.cert.store.source is configured differently.

Suggestion: Add an explicit secrets backend selector, or extend server.cert.store.source to include vault/openbao, and only build VaultClient when that value is selected.

src/outbound/vault.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub fn timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = timeout;
+        self
+    }
+
+    /// Build the [`VaultClient`] adapter instance.
+    pub fn build(self) -> Result<VaultClient, StorageError> {
+        let client = Client::builder()
+            .timeout(self.timeout)
+            .build()
```
Ngha-Boris, line 240: The builder accepts an empty token, empty mount, and default localhost address without validation. Docs say the token is mandatory when Vault is enabled, but startup can succeed and fail later with less actionable connection/auth errors.

Suggestion: Validate selected Vault config at build/startup: reject empty token, empty mount, empty address, and invalid URLs with config-specific messages that do not expose secret values.

tests/cert_provisioning.rs
```diff
@@ -283,3 +315,139 @@ async fn test_certificate_renewal_with_existing_cert() {
     assert_eq!(initial_cert.valid_from, current_cert.valid_from);
     assert_eq!(initial_cert.expires_at, current_cert.expires_at);
 }
+
+#[cfg(feature = "vault")]
+#[tokio::test]
+async fn test_cert_provisioning_with_hashicorp_vault() {
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```
Ngha-Boris, line 322: The Vault/OpenBao tests cover certificate provisioning, but there are no focused adapter tests for KV v2 behavior: mount/path prefix construction, namespace header, TTL cache disabled with 0, cache hit behavior, missing secret handling, forbidden handling, or secret redaction in errors.

Suggestion: Add wiremock tests around VaultClient itself for store/load/delete, 404 -> Ok(None), 403 -> clear error, namespace header, path prefix/mount URL shape, and cache TTL 0 causing repeated HTTP reads.

Ngha-Boris (commented):
Hello @Hermann-Core please take a look at the comments below

src/setup.rs
Hermann-Core: Backend selection stays feature-flag-only (not runtime config) so that unused backend dependencies are never pulled in at compile time. The `server.cert.store.source` runtime selector will be dropped as part of the config consolidation tracked in #168.

src/outbound/vault.rs
Hermann-Core: Addressed in commit 417fc6c — `VaultClientBuilder::build()` now validates all three fields before touching the HTTP client

tests/cert_provisioning.rs
Hermann-Core: Unit tests for the adapter's HTTP behaviour (store/load/delete, 404 → Ok(None), 403 handling, cache hit, namespace header, path/mount URL shape) are already covered by the Docker-based integration tests in this file. The new `#[cfg(test)]` module added in commit 417fc6c focuses exclusively on `VaultClientBuilder` configuration validation — the boundary that wasn't tested and that benefits most from fast, dependency-free unit tests.

src/outbound/vault.rs
Christiantyemele: Fixed. Keys are now percent-encoded per RFC 3986 via encode_path_segment() in qualify_key(). Committed in `ae29dcb`.

src/outbound/vault.rs
Christiantyemele: Fixed. store() now has a distinct FORBIDDEN branch matching load()'s 403 handling. Committed in `ae29dcb`.

src/outbound/vault.rs
Christiantyemele: Fixed. Increased CACHE_MAX_CAPACITY from 10 to 100 with rationale comment. Committed in `ae29dcb`.

src/outbound/vault.rs
Christiantyemele: Fixed. Module doc trailing semicolon replaced with period. Committed in `ae29dcb`.

src/config.rs
Christiantyemele: Fixed. Double period in VaultConfig.mount doc comment corrected to single period. Committed in `ae29dcb`.

src/outbound/vault.rs
Christiantyemele: Fixed. delete() now has a distinct FORBIDDEN branch matching store() and load() for consistent debuggability. Committed in `ae29dcb`.

Christiantyemele (commented):
.

tests/cert_provisioning.rs
```diff
@@ -283,3 +315,139 @@ async fn test_certificate_renewal_with_existing_cert() {
...
+#[cfg(feature = "vault")]
+#[tokio::test]
+async fn test_cert_provisioning_with_openbao() {
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+
+    let infra = TestInfra::start("openbao-provision").await;
+    let _openbao_container = GenericImage::new(OPENBAO_IMAGE, OPENBAO_TAG)
+        .with_exposed_port(8200.tcp())
+        .with_wait_for(WaitFor::message_on_stdout(
+            "Development mode should NOT be used in production",
```
Christiantyemele, line 393: **Nit**: OpenBao test sets both `VAULT_DEV_*` and `BAO_DEV_*` env vars. OpenBao only reads `BAO_*` vars since the fork — the `VAULT_DEV_ROOT_TOKEN_ID` and `VAULT_DEV_LISTEN_ADDRESS` are dead config. Consider removing them to avoid confusion about which code path is being exercised.

docs/secrets-backends.md
```diff
@@ -0,0 +1,78 @@
...
+
+The server supports storing sensitive cryptographic materials (such as ACME account keys and server signing keys) in secure secrets management backends.
+
+The supported secrets backends include:
+
+- **HashiCorp Vault / OpenBao** (KV v2 engine)
+- **AWS Secrets Manager**
+- **In-Memory** (for development and testing)
+
+## HashiCorp Vault & OpenBao (KV v2)
```
Christiantyemele, line 11: **Nit**: The doc says Vault/OpenBao are "supported natively when the `vault` feature flag is enabled" but doesn't mention the build command. Consider adding a one-liner like `cargo build --features vault` in the prerequisites, so operators know how to produce a vault-enabled binary.

Hermann-Core: The build command does not belong in this document. This document is focused on the supported secrets backends and their configuration, not on how the server is built. Build and development commands are already documented in their respective location, so including them here introduces unnecessary duplication and increases the risk of documentation drift.

src/config.rs
```diff
@@ -812,6 +833,13 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("server.cert.store.certificate_key", Option::<String>::None)?
         .set_default("server.cert.store.signing_key_key", Option::<String>::None)?
         .set_default("aws.region", "us-east-1")?
+        .set_default("vault.addr", "http://localhost:8200")?
+        .set_default("vault.token", "root")?
```
Christiantyemele: **Important**: `vault.token` defaults to `root`, so a vault-enabled binary no longer detects a missing `APP_VAULT__TOKEN`. The docs say this value is mandatory when Vault is enabled, and the builder's empty-token validation cannot fire because config injects a valid-looking token first. In production that can silently aim requests at localhost with a dev/root token value, or produce misleading runtime auth failures instead of a clear startup config error. Please keep the default empty and set `root` only in tests/dev examples that intentionally run a dev server.

```rust
.set_default("vault.token", "")?
```

src/outbound/vault.rs
```diff
@@ -0,0 +1,489 @@
...
+
+/// The actual key-value pairs stored in Vault.
+/// We store a single field named `value`.
+#[derive(Serialize, Deserialize)]
+struct KvDataPayload {
+    value: String,
+}
+
+#[async_trait]
+impl Storage for VaultClient {
```
Christiantyemele, line 286: **Important**: `VaultClient` inherits `Storage::reachable()`, whose default implementation returns `Ok(())`, so readiness will report the secrets backend as healthy even when Vault/OpenBao is down, the token is revoked, or the configured mount is inaccessible. `AppState` health probes call `Storage::reachable()` for the cert and secrets stores, so this backend needs a real, non-secret probe before traffic is routed to an instance that cannot sign or renew certificates. Add an override that performs an authenticated Vault request allowed by the deployment policy and maps unreachable/403 responses to `StorageError::Backend` without including secret values.

```rust
async fn reachable(&self) -> Result<(), StorageError> {
    let resp = self
        .add_auth_headers(self.client.get(format!("{}/v1/{}/config", self.addr, self.mount)))
        .send()
        .await?;

    if resp.status().is_success() {
        Ok(())
    } else {
        Err(StorageError::Backend(eyre!(
            "vault readiness check failed: HTTP {}",
            resp.status()
        )))
    }
}
```

Christiantyemele (changes requested):
looks good except for the few NITS

Christiantyemele (dismissed):

Ngha-Boris (dismissed):

Ngha-Boris (approved):

Christiantyemele (approved):

PR #344, opened 2026-08-07
certificate storage and secrets backend selection

### LINKED ISSUE #308: Split certificate storage and secrets backend selection
## Objective

Introduce explicit, provider-neutral configuration and composition for certificate object storage and secrets storage.

## Current State

`src/setup.rs` currently selects AWS storage implicitly: when the `aws` feature is compiled and `aws.s3_bucket` is non-empty, it constructs `AwsS3` for certificate storage and `AwsSecretsManager` for secrets storage together. This makes it hard to combine, for example, S3-compatible object storage with Vault/OpenBao secrets, or filesystem certificates with a cloud secrets manager.

## Deliverables

- [x] Add explicit config enums for certificate object storage and secrets storage providers.
- [x] Split certificate storage selection from secrets storage selection in `src/setup.rs`.
- [x] Preserve existing AWS behavior through backwards-compatible defaults or a documented migration path.
- [x] Fail fast with clear startup errors when a selected provider is not compiled in or is missing required config.
- [x] Add unit tests for provider resolution and composition errors.

## Acceptance Criteria

- [x] Operators can select certificate object storage independently from signing-key/secret storage.
- [x] `aws.s3_bucket` no longer implicitly decides the secrets backend.
- [x] Store provisioning source names are no longer AWS-specific where a provider-neutral name is more accurate.
- [x] Existing AWS S3 + AWS Secrets Manager deployment still works.

## Verification

- [x] Config tests cover every provider selector.
- [x] `cargo test config`

Estimated time: 1.5 day

Blindspot22:
Introduce explicit, provider-neutral configuration and composition for certificate object storage and secrets storage.


src/config.rs
```diff
@@ -763,6 +807,11 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
     #[cfg(not(feature = "acme"))]
     let default_chain_cache_ttl = 86400;
 
+    #[cfg(feature = "aws")]
+    let (default_cert_storage, default_secrets_storage) = ("aws_s3", "aws_secrets_manager");
```
Hermann-Core: the `aws` umbrella feature (which also gates Route53 DNS etc.) now defaults both cert storage and secrets storage to AWS. This partially reintroduces the coupling the issue (#308) is meant to remove, and it changes startup behavior for partial-AWS builds.

With the old code, an ``aws`` build whose ``aws.s3_bucket`` was empty (and source wasn't ``aws_secrets_manager``) silently fell back to memory/memory. Under this change that same config fails fast with ``aws.s3_bucket is required`` unless the operator explicitly sets both ``cert_storage=memory`` / ``secrets_storage=memory``.

It works for the standard chart deployment (which sets ``s3_bucket`` and ``route53``), so it's backwards-compatible there but any deployment enabling the ``aws`` feature for reasons unrelated to storage (e.g. Route53-only DNS) now inherits two cloud storage defaults it may not want, and cannot independently default one backend to memory. The `aws` feature must be removed.

Also: the new ``server.cert.cert_storage`` / ``server.cert.secrets_storage`` knobs aren't surfaced in the helm chart's documented env vars — operators relying on charts won't know they exist unless they read the README. Consider exposing them in the chart for consistency with the other cert settings.

src/config.rs
```diff
@@ -178,6 +178,46 @@ pub struct ServerConfig {
     pub aggregation_uri: Option<String>,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum CertStorageProvider {
+    #[default]
+    Memory,
+    #[serde(alias = "s3", alias = "aws_s3", alias = "aws-s3")]
+    AwsS3,
```
Hermann-Core, line 187: ``#[serde(rename_all = "lowercase")]`` turns this variant into the canonical serde name ``"awss3"`` (and ``AwsSecretsManager`` → ``"awssecretsmanager"``), which disagrees with both ``as_str()`` and the documented value (``"aws_s3"``/``"aws_secrets_manager"``). Deserialization only works because of the explicit aliases. For internal consistency, prefer an explicit ``#[serde(rename = "aws_s3")]`` on the variant so the canonical name matches ``as_str()`` and the docs, keeping the aliases for backward compatibility.

src/setup.rs
```diff
@@ -240,32 +242,44 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+                    match cache_opt {
                         Some(c) => Box::new(s3.with_cache(c)),
                         None => Box::new(s3),
-                    };
+                    }
+                }
+                #[cfg(not(feature = "aws"))]
+                {
+                    return Err(color_eyre::eyre::eyre!(
+                        "s3 certificate storage provider selected, but 'aws' feature flag was not compiled in."
```
Hermann-Core: **Nit** — the error says ``"s3 certificate storage provider selected"`` but the provider is named ``aws_s3`` everywhere else (``as_str()``, env var docs). Align the wording to ``aws_s3`` for consistency.

src/setup.rs
```diff
@@ -240,32 +242,44 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
                         &config.aws.region,
                         &config.aws.s3_key_prefix,
                     );
-                    let cert_st: Box<dyn Storage> = match cache_opt {
+                    match cache_opt {
```
Hermann-Core, line 253: the ``#[cfg(not(feature = "redis"))] let cache_opt: Option<_> = None;`` earlier in this arm fails to compile under ``--features aws,acme`` (without ``redis``) with ``type annotations needed for Option<_>``. This is pre-existing (same block fails on the base commit), but since this function is being rewritten, a one-token fix (``let cache_opt: Option<_> = None;``) here prevents shipping the latent break. CI's ``--all-features`` won't catch this feature combination.

Hermann-Core (commented):
@Blindspot22 Comments have been posted below for your attention.

Blindspot22:
Okay on it @Hermann-Core :+1:

helm/chart/values.yaml
```diff
@@ -53,6 +53,10 @@ statuslist:
     APP_SERVER__CERT__ACME_DIRECTORY_URL: "https://acme-v02.api.letsencrypt.org/directory"
     APP_SERVER__CERT__EKU: "1,3,6,1,5,5,7,3,30"
     APP_SERVER__CERT__RENEWAL_CRON_SCHEDULE: "0 0 0 * * *"
+    # Certificate material storage backend: memory | aws_s3
+    APP_SERVER__CERT__CERT_STORAGE: "aws_s3"
+    # Secrets (key material) storage backend: memory | aws_secrets_manager
+    APP_SERVER__CERT__SECRETS_STORAGE: "aws_secrets_manager"
```
Hermann-Core: I suggest renaming this to `aws_secrets` to align the upcomming naming convention that will be introduced by the vault support

src/config.rs
```diff
@@ -178,6 +178,46 @@ pub struct ServerConfig {
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum SecretsStorageProvider {
+    #[default]
+    Memory,
+    #[serde(
+        rename = "aws_secrets_manager",
+        alias = "aws-secrets-manager",
+        alias = "secrets_manager"
```
Hermann-Core: `aws` should explicitly be mentioned to avoid any confusion.

src/config.rs
Hermann-Core, line 1: Refactor the backend selection to rely exclusively on Cargo features rather than selecting backends dynamically at runtime. The goal is to make backend support explicit and ensure that only the drivers and dependencies required by the selected backends are compiled into the project.

- Remove the generic `aws` feature entirely. It is too implicit and error-prone because enabling it does not clearly communicate which AWS services are actually being used.
- Keep the feature configuration consistent with the existing feature-gating pattern throughout the project.
- Ensure implementations gated behind a feature do not leak outside that feature's context.
- Verify that the project still compiles and tests successfully with the relevant feature combinations.

README.md
```diff
@@ -211,10 +213,20 @@ The Status List Server is provisioned with a cryptographic certificate that is e
...
+**Certificate and Secrets Storage Backends:**
+
+Certificate object storage (`server.cert.cert_storage`) and secrets storage (`server.cert.secrets_storage`) are selected independently:
+
+- `memory` — in-process ephemeral storage (the default for both backends regardless of compiled features). AWS storage backends require explicit opt-in.
+- `aws_s3` — AWS S3 for certificate storage. Requires `aws.s3_bucket` and the `aws` feature.
+- `aws_secrets_manager` — AWS Secrets Manager for secrets storage. Requires the `aws` feature.
+
+This allows mixed deployments such as S3-compatible object storage for certificates with an in-memory or alternative secrets backend.
+
```
Hermann-Core, line 231: You should not leak internal implementation details in public documentation.

Hermann-Core (changes requested):
@Blindspot22 I left some comments below for you to address.

The main goal here is to remove the generic `aws` feature, which unnecessarily couples AWS Secrets Manager and S3 together, and make backend support explicit. This should also ensure that only the drivers and dependencies required by the selected backends are compiled into the project.

Hermann-Core:
We are going to move toward a consolidated cryptographic-material backend by default: the same selected secrets/material backend should manage both the signing key and the certificate chain associated with it. Closing this.

PR #354, opened 2026-08-11
351 enforce conventional commit messages and pr titles via ci

Blindspot22:
Since `release-plz` derives semantic version bumps automatically from Conventional Commits (feat, fix, refactor, feat!, etc.), commit messages and PR titles on `main` must adhere strictly to the Conventional Commits specification. Without enforcement, unconventional commits (e.g. "fixed stuff") will be ignored by `git-cliff` and `release-plz`, causing inaccurate version bumps and incomplete changelogs.


Ngha-Boris:
There is another pull request that implements this.

PR #375, opened 2026-08-14
feat: Consolidate certificate and signing-key material backend

### LINKED ISSUE #361: Consolidate certificate and signing-key management into one cryptographic-material backend
## Objective

Consolidate certificate-chain and signing-key management behind one cryptographic-material backend by default.

## Decision Direction

The intended direction is that the same selected secrets/material backend manages both:

- the server signing key, and
- the certificate chain associated with that key.

This should be the default and recommended model because the key and certificate are lifecycle-coupled: provisioning, renewal, rotation, cache invalidation, access control, audit policy, and rollback should be reasoned about together.

Split certificate and key backends can be reconsidered later only if real deployment needs justify the added complexity.

## Current State

The code currently has separate `cert_storage` and `secrets_storage` constructor paths in the certificate manager builder/composition root. Earlier tickets also discussed splitting certificate object storage from secrets storage. That flexibility is useful in theory, but it makes the operational model more complex and can create mismatched rotation/cache behavior.

## Deliverables

- [ ] Define a consolidated backend abstraction/name for server cryptographic material.
- [ ] Make the selected secrets/material backend responsible for both signing key and certificate chain.
- [ ] Preserve filesystem store provisioning for local/dev and externally managed cert/key material.
- [ ] Add optional caching for certificate and key reads, with explicit no-cache mode for private-key material if required by policy.
- [ ] Use one invalidation path when key/certificate material is provisioned, renewed, rotated, or reloaded.

## Acceptance Criteria

- [ ] Configuration presents one primary backend choice for server cryptographic material.
- [ ] Certificate and signing key rotation are coordinated through the same backend.
- [ ] Cache invalidation is consistent for both certificate and key material.

## Verification

- [ ] Config tests.
- [ ] Certificate manager tests for provision, load, rotate, cache hit/miss, and invalidation.

Ngha-Boris:
**Description**
This PR consolidates server certificate-chain and signing-key management behind a single cryptographic-material backend. The selected material backend now owns the lifecycle-coupled server signing key, certificate chain, and related ACME account material by default.

**Changes**
- Added `CryptoMaterialStorage` and `CryptoMaterialCachePolicy`.
- Added primary config keys:
  - `server.cert.material_backend`
  - `server.cert.material_cache_ttl`
  - `server.cert.signing_key_cache_ttl`
- Updated `CertManager` and builder wiring to use one `crypto_material_storage(...)` backend.
- Kept legacy `cert_storage` / `secrets_storage` builder methods as compatibility aliases.
- Updated ACME and store provisioning to persist certificate and signing key through the same material backend.
- Preserved filesystem store provisioning for externally managed cert/key material.
- Added optional material read caching, with explicit no-cache support for private key reads via `signing_key_cache_ttl = 0`.
- Reworked cache refresh/invalidation so certificate data, parsed certificate chain, and signing-key cache entries are invalidated through one path.
- Updated readiness checks to probe the single material backend.
- Updated tests, integration harness expectations, and docs to reflect the consolidated backend model.

**Verification**
- `cargo check --all-features`
- `cargo test --all-features --no-run`
- `cargo test --lib --features acme utils::cert_manager::tests:: -- --test-threads=1`
- `cargo test --lib --features acme test_config_loading`
- `cargo test --lib --features acme ready_fails_when_crypto_material_backend_is_unreachable`

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -26,9 +26,7 @@ APP_SERVER__CERT__EKU=1,3,6,1,5,5,7,3,30
 # The cryptographic material backend is selected by enabled Cargo features:
 # vault => Vault/OpenBao; aws-secrets without vault => AWS Secrets Manager;
 # neither feature => in-memory storage for local development and tests.
-# TTL in seconds for certificate material reads from the selected backend.
-# Set to 0 to disable this material cache.
-APP_SERVER__CERT__MATERIAL_CACHE_TTL=300
+# Certificate material stays cached until provisioning or renewal invalidates it.
 # TTL in seconds for private signing-key reads from the selected backend.
 # Set to 0 to force every private-key read to the backend.
 APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL=0
--- a/.env.template
+++ b/.env.template
@@ -23,6 +23,14 @@ APP_SERVER__CERT__ORGANIZATION=example.com
 APP_SERVER__CERT__ACME_DIRECTORY_URL=https://pebble:14000/dir
 # The last number is not explicitly defined in the spec
 APP_SERVER__CERT__EKU=1,3,6,1,5,5,7,3,30
+# Primary backend for server cryptographic material: memory | aws_secrets_manager | vault
+APP_SERVER__CERT__MATERIAL_BACKEND=memory
+# TTL in seconds for certificate material reads from the selected backend.
+# Set to 0 to disable this material cache.
+APP_SERVER__CERT__MATERIAL_CACHE_TTL=300
+# TTL in seconds for private signing-key reads from the selected backend.
+# Set to 0 to force every private-key read to the backend.
+APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL=0
 # Cron schedule for certificate renewal checks (6-field cron: sec min hour day month day-of-week)
 APP_SERVER__CERT__RENEWAL_CRON_SCHEDULE=0 0 0 * * *
 # TTL in seconds for the in-memory parsed certificate chain cache.
@@ -34,8 +42,8 @@ APP_SERVER__CERT__CHAIN_CACHE_TTL=3600
 
 # Store provisioning settings. Required only when APP_SERVER__CERT__PROVISIONING_STRATEGY=store.
 # source=filesystem: load from local files. Files may contain PEM text or raw DER.
-# source=storage: load cert from configured certificate storage and key from configured secrets storage.
-# source=aws_secrets_manager: load both cert and key from configured AWS Secrets Manager backend.
+# source=storage: load cert and key from the configured cryptographic material backend.
+# source=aws_secrets_manager: compatibility alias for storage.
 # Storage-backed values may contain PEM text or base64/base64url-encoded DER.
 # Private keys must be PKCS#8 PEM or DER.
 APP_SERVER__CERT__STORE__SOURCE=filesystem
@@ -103,14 +111,15 @@ APP_DATABASE__POOL__IDLE_TIMEOUT_SECS=600
 APP_DATABASE__POOL__MAX_LIFETIME_SECS=1800
 
 # Redis configuration (optional)
-# Redis is NOT required for status-list persistence or reads. It is only used
-# as a distributed certificate-material cache for multi-replica deployments
-# that use AWS S3 certificate storage with the 'redis' feature compiled.
+# Redis is optional. The server does not use Redis for status-list persistence
+# or status-list reads; those use the configured repository backend and the
+# in-process status-list cache. Certificate and signing-key material are managed
+# together by the selected cryptographic material backend; prefer the built-in
+# material read-cache TTLs before adding Redis.
 # Leave APP_REDIS__URI empty to disable Redis entirely.
 APP_REDIS__URI=
 APP_REDIS__REQUIRE_CLIENT_AUTH=false
-# TTL for the certificate cache in Redis.
-# Set to 0 to disable and always fetch from S3.
+# TTL for Redis-backed certificate cache integrations that explicitly opt in.
 APP_REDIS__CERT_CACHE_TTL=3600
 
 # AWS SDK
@@ -121,14 +130,10 @@ AWS_SECRET_ACCESS_KEY=test
 # TTL for the Secrets Manager cache.
 # Set to 0 to disable and always fetch from AWS API directly.
 APP_AWS__SECRETS_CACHE_TTL=300
-# S3 bucket name for storing certificate data
-APP_AWS__S3_BUCKET=status-list-adorsys
-# Optional prefix prepended to all S3 object keys (empty = no prefix)
-APP_AWS__S3_KEY_PREFIX=
 
 # Vault / OpenBao KV v2 configuration (feature = "vault")
-# When compiled with `vault` feature, Vault/OpenBao is used
-# as the secrets backend for signing-key material.
+# When compiled with the `vault` feature, Vault/OpenBao can be selected as the
+# cryptographic material backend for certificate and signing-key material.
 #
 # APP_VAULT__ADDR=http://vault:8200
 # APP_VAULT__TOKEN=hvs.XXXXXXXXXXXXXXXXXXXXXXXX
--- a/.env.template
+++ b/.env.template
@@ -23,8 +23,9 @@ APP_SERVER__CERT__ORGANIZATION=example.com
 APP_SERVER__CERT__ACME_DIRECTORY_URL=https://pebble:14000/dir
 # The last number is not explicitly defined in the spec
 APP_SERVER__CERT__EKU=1,3,6,1,5,5,7,3,30
-# Primary backend for server cryptographic material: memory | aws_secrets_manager | vault
-APP_SERVER__CERT__MATERIAL_BACKEND=memory
+# The cryptographic material backend is selected by enabled Cargo features:
+# vault => Vault/OpenBao; aws-secrets without vault => AWS Secrets Manager;
+# neither feature => in-memory storage for local development and tests.
 # TTL in seconds for certificate material reads from the selected backend.
 # Set to 0 to disable this material cache.
 APP_SERVER__CERT__MATERIAL_CACHE_TTL=300
--- a/.env.template
+++ b/.env.template
@@ -34,9 +34,7 @@ APP_SERVER__CERT__MATERIAL_CACHE_TTL=300
 APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL=0
 # Cron schedule for certificate renewal checks (6-field cron: sec min hour day month day-of-week)
 APP_SERVER__CERT__RENEWAL_CRON_SCHEDULE=0 0 0 * * *
-# TTL in seconds for the in-memory parsed certificate chain cache.
-# Set to 0 to keep entries forever; in multi-replica deployments, non-provisioning replicas rely on this TTL to pick up renewed chains.
-APP_SERVER__CERT__CHAIN_CACHE_TTL=3600
+# Parsed certificate chains stay cached until certificate provisioning or renewal replaces them.
 # DNS challenge server URL. Optional; only used in development mode (APP_ENV=development).
 # Falls back to http://challtestsrv:8055 when unset.
 # APP_SERVER__CERT__DNS_CHALLENGE_SERVER_URL=http://challtestsrv:8055
--- a/.env.template
+++ b/.env.template
@@ -41,16 +41,14 @@ APP_SERVER__CERT__CHAIN_CACHE_TTL=3600
 # APP_SERVER__CERT__DNS_CHALLENGE_SERVER_URL=http://challtestsrv:8055
 
 # Store provisioning settings. Required only when APP_SERVER__CERT__PROVISIONING_STRATEGY=store.
-# source=filesystem: load from local files. Files may contain PEM text or raw DER.
-# source=storage: load cert and key from the configured cryptographic material backend.
-# source=aws_secrets_manager: compatibility alias for storage.
+# Configure either filesystem paths or storage keys, not both.
+# Filesystem values may contain PEM text or raw DER.
 # Storage-backed values may contain PEM text or base64/base64url-encoded DER.
 # Private keys must be PKCS#8 PEM or DER.
-APP_SERVER__CERT__STORE__SOURCE=filesystem
-# Required for source=filesystem.
+# Filesystem provisioning:
 # APP_SERVER__CERT__STORE__CERTIFICATE_PATH=/etc/status-list/tls.crt
 # APP_SERVER__CERT__STORE__SIGNING_KEY_PATH=/etc/status-list/tls.key
-# Required for source=storage or source=aws_secrets_manager.
+# Storage-backed provisioning from the selected cryptographic material backend:
 # APP_SERVER__CERT__STORE__CERTIFICATE_KEY=status-list/server-certificate
 # APP_SERVER__CERT__STORE__SIGNING_KEY_KEY=status-list/server-signing-key
 
@@ -127,9 +125,6 @@ APP_AWS__REGION=us-east-1
 AWS_ENDPOINT_URL=http://localstack:4566 # remove this in production
 AWS_ACCESS_KEY_ID=test
 AWS_SECRET_ACCESS_KEY=test
-# TTL for the Secrets Manager cache.
-# Set to 0 to disable and always fetch from AWS API directly.
-APP_AWS__SECRETS_CACHE_TTL=300
 
 # Vault / OpenBao KV v2 configuration (feature = "vault")
 # When compiled with the `vault` feature, Vault/OpenBao can be selected as the
@@ -144,8 +139,6 @@ APP_AWS__SECRETS_CACHE_TTL=300
 # APP_VAULT__PATH_PREFIX=
 # Vault Enterprise / OpenBao namespace (default: empty / unset)
 # APP_VAULT__NAMESPACE=
-# TTL for in-memory secrets cache in seconds. Set to 0 to disable caching
-# APP_VAULT__SECRETS_CACHE_TTL=300
 # HTTP request timeout in seconds
 # APP_VAULT__TIMEOUT_SECS=30
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ mysql = ["history", "sea-orm?/sqlx-mysql"]
 redis = ["dep:redis"]
 vault = ["acme"]
 aws-secrets = [
+  "acme",
   "dep:aws-config",
   "dep:aws-sdk-route53",
   "dep:aws-sdk-s3",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ sqlite = ["history", "sea-orm?/sqlx-sqlite"]
 mysql = ["history", "sea-orm?/sqlx-mysql"]
 redis = ["dep:redis"]
 vault = ["acme"]
-aws = [
+aws-secrets = [
   "dep:aws-config",
   "dep:aws-sdk-route53",
   "dep:aws-sdk-s3",
--- a/README.md
+++ b/README.md
@@ -93,15 +93,15 @@ By default, the server will listen on `http://localhost:8000` using in-memory re
 
 The crate uses modular Cargo feature flags to gate optional production backend drivers:
 
-| Feature       | Description                                                                                                                                        | Default    |
-| ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | ---------- |
-| `memory`      | In-memory repositories (`MemoryStatusLists`, `MemoryCredentials`, `MemoryStatusListHistory`), Moka TTL cache, and store-based certificate storage. | ✅ Default |
-| `postgres`    | SeaORM PostgreSQL database driver.                                                                                                                 | ❌ Opt-in  |
-| `sqlite`      | SeaORM SQLite database driver.                                                                                                                     | ❌ Opt-in  |
-| `mysql`       | SeaORM MySQL database driver.                                                                                                                      | ❌ Opt-in  |
-| `redis`       | Redis storage driver for explicit cache/storage integrations.                                                                                      | ❌ Opt-in  |
-| `aws-secrets` | AWS S3 object storage, Route53 DNS-01, and AWS Secrets Manager drivers.                                                                            | ❌ Opt-in  |
-| `acme`        | ACME DNS-01 certificate manager driver.                                                                                                            | ❌ Opt-in  |
+| Feature       | Description                                                             | Default    |
+| ------------- | ----------------------------------------------------------------------- | ---------- |
+| `memory`      | In-memory repositories, TTL cache, and store-based certificate storage. | ✅ Default |
+| `postgres`    | SeaORM PostgreSQL database driver.                                      | ❌ Opt-in  |
+| `sqlite`      | SeaORM SQLite database driver.                                          | ❌ Opt-in  |
+| `mysql`       | SeaORM MySQL database driver.                                           | ❌ Opt-in  |
+| `redis`       | Redis storage driver for explicit cache/storage integrations.           | ❌ Opt-in  |
+| `aws-secrets` | Route53 DNS-01, and AWS Secrets Manager drivers.                        | ❌ Opt-in  |
+| `acme`        | ACME DNS-01 certificate manager driver.                                 | ❌ Opt-in  |
 
 To build with specific backend drivers, pass the matching feature flag(s):
 
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ The Status List Server is provisioned with a cryptographic certificate that is e
 - Certificate issuance and renewal are managed according to the configured renewal strategy.
 - Every day, a cron job checks whether the certificate should be renewed based on this strategy.
 - If the certificate is still considered valid according to the configured strategy, no renewal occurs; renewal is only triggered when necessary.
-- The server signing key and certificate chain are stored in one selected cryptographic-material backend, configured with `server.cert.material_backend`.
+- The server signing key and certificate chain are stored in one cryptographic-material backend selected by enabled Cargo features (`vault`, `aws-secrets`, or in-memory fallback).
 - Certificate and signing-key reads can be cached with `APP_SERVER__CERT__MATERIAL_CACHE_TTL` and `APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL`. Set the signing-key TTL to `0` to force private-key reads to bypass the material cache.
 - Parsed certificate chains are cached in memory for `APP_SERVER__CERT__CHAIN_CACHE_TTL` seconds (default: `3600`). A value of `0` keeps entries indefinitely. In multi-replica deployments, replicas that did not perform renewal rely on this TTL to refresh their in-memory chain.
 
@@ -211,7 +211,7 @@ The Status List Server is provisioned with a cryptographic certificate that is e
 - `server.cert.provisioning_strategy = "acme"` requests and renews certificates through ACME.
 - `server.cert.provisioning_strategy = "store"` loads externally managed certificate material and persists it into the configured cryptographic-material backend.
 - Store provisioning infers filesystem loading when `server.cert.store.certificate_path` and `signing_key_path` are configured.
-- Store provisioning infers storage-backed loading when `server.cert.store.certificate_key` and `signing_key_key` are configured; those keys are read from the selected cryptographic-material backend.
+- Store provisioning infers storage-backed loading when `server.cert.store.certificate_key` and `signing_key_key` are configured; those keys are read from the feature-selected cryptographic-material backend.
 - Filesystem store inputs may be PEM text or raw DER. Storage-backed store inputs may be PEM text or base64/base64url-encoded DER. Private keys must be PKCS#8 in PEM or DER form.
 - The renewal cron schedule is configured with `server.cert.renewal_cron_schedule`. For store provisioning, each scheduled run reloads the configured source and refreshes persisted material only when it changed.
 
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ The Status List Server is provisioned with a cryptographic certificate that is e
 - If the certificate is still considered valid according to the configured strategy, no renewal occurs; renewal is only triggered when necessary.
 - The server signing key and certificate chain are stored in one cryptographic-material backend selected by enabled Cargo features (`vault`, `aws-secrets`, or in-memory fallback).
 - Certificate and signing-key reads can be cached with `APP_SERVER__CERT__MATERIAL_CACHE_TTL` and `APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL`. Set the signing-key TTL to `0` to force private-key reads to bypass the material cache.
-- Parsed certificate chains are cached in memory for `APP_SERVER__CERT__CHAIN_CACHE_TTL` seconds (default: `3600`). A value of `0` keeps entries indefinitely. In multi-replica deployments, replicas that did not perform renewal rely on this TTL to refresh their in-memory chain.
+- Parsed certificate chains stay cached in memory until certificate provisioning or renewal replaces them.
 
 **Provisioning Modes:**
 
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ The crate uses modular Cargo feature flags to gate optional production backend d
 | `postgres` | SeaORM PostgreSQL database driver.                                                                                                                 | ❌ Opt-in  |
 | `sqlite`   | SeaORM SQLite database driver.                                                                                                                     | ❌ Opt-in  |
 | `mysql`    | SeaORM MySQL database driver.                                                                                                                      | ❌ Opt-in  |
-| `redis`    | Redis storage driver for legacy or explicit cache/storage integrations.                                                                            | ❌ Opt-in  |
+| `redis`    | Redis storage driver for explicit cache/storage integrations.                                                                                      | ❌ Opt-in  |
 | `aws`      | AWS S3 object storage and AWS Secrets Manager drivers.                                                                                             | ❌ Opt-in  |
 | `acme`     | ACME DNS-01 certificate manager driver.                                                                                                            | ❌ Opt-in  |
 
@@ -115,9 +115,7 @@ cargo run --features postgres,aws,acme
 
 ## Redis Role
 
-Redis is not required for core status-list persistence or status-list reads. Status-list records are stored by the configured repository backend (`memory`, PostgreSQL, MySQL, or SQLite), and hot status-list reads are cached in-process by `MokaStatusListCache`.
-
-The ACME certificate manager uses the selected cryptographic-material backend for both the certificate chain and signing key. Prefer the built-in material read caches (`APP_SERVER__CERT__MATERIAL_CACHE_TTL` and `APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL`) before adding Redis. Redis remains available only for explicit adapter-level integrations that still opt into it.
+Redis is optional. The server does not use Redis for status-list persistence or status-list reads; those use the configured repository backend and the in-process status-list cache. Certificate and signing-key material are managed together by the selected cryptographic material backend; prefer the built-in material read-cache TTLs before adding Redis.
 
 For single-replica deployments, local development, tests, or deployments where certificate material reads are inexpensive, leave `APP_REDIS__URI` unset and omit the `redis` Cargo feature.
 
@@ -224,7 +222,7 @@ The Status List Server is provisioned with a cryptographic certificate that is e
 - ACME uses `DefaultHttpClient` unless `.acme_http_client(...)` is supplied.
 - Store provisioning does not create ACME HTTP client state unless explicitly configured.
 - `email` defaults to an empty string, `organization` defaults to none, and `eku` defaults to none.
-- `domains` and `crypto_material_storage` must always be provided.
+- `domains` and `crypto_storage` must always be provided.
 - ACME additionally requires `challenge_handler` and `acme_directory_url`.
 
 ```rust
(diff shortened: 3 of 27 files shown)


src/utils/cert_manager.rs
```diff
@@ -724,56 +726,59 @@ impl CertManager {
...
-    async fn persist_certificate_data_with_storage(
-        &self,
-        cert_storage: &dyn Storage,
-        cert_data: &CertificateData,
     ) -> Result<(), CertError> {
         let serialized_cert_data = serde_json::to_string(cert_data)?;
-        cert_storage
-            .store(&self.cert_key(), &serialized_cert_data)
+        self.crypto_material_storage()?
+            .store_certificate_data(&self.cert_key(), &serialized_cert_data)
```
Hermann-Core: **Critical:** `persist_certificate_data` unconditionally calls `store_certificate_data`, which invokes `self.backend.store(...)`.

For storage adapters with "create-if-absent" semantics (such as `AwsSecretsManager`), calling `store` on an existing secret logs a warning (`"Secret already exists. Skipping..."`) and returns `Ok(())` without updating the secret value. Consequently, when a certificate is renewed (via ACME renewal or store refresh), the new certificate chain is silently never persisted to Secrets Manager.

`persist_certificate_data` should check if the certificate already exists and call `update_certificate_data` (which invokes `self.backend.update(...)`), exactly mirroring how `persist_signing_key` handles key updates:

```rust
pub(crate) async fn persist_certificate_data(
    &self,
    cert_data: &CertificateData,
) -> Result<(), CertError> {
    let serialized_cert_data = serde_json::to_string(cert_data)?;
    let material_storage = self.crypto_material_storage()?;
    let cert_key = self.cert_key();
    if material_storage
        .load_certificate_data(&cert_key)
        .await?
        .is_some()
    {
        material_storage
            .update_certificate_data(&cert_key, &serialized_cert_data)
            .await?;
    } else {
        material_storage
            .store_certificate_data(&cert_key, &serialized_cert_data)
            .await?;
    }
    Ok(())
}
```

Please also expose `pub async fn update_certificate_data(&self, key: &str, value: &str) -> Result<(), StorageError>` on `CryptoMaterialStorage` in `storage.rs`.

src/utils/cert_manager/builder.rs
```diff
@@ -130,18 +129,39 @@ impl CertificateManagerBuilder {
...
+            self.cache_policy,
+        ));
         self
     }
 
-    /// Set the signing key and ACME account storage backend.
-    pub fn secrets_storage(mut self, storage: impl Storage + 'static) -> Self {
-        self.secrets_storage = Some(Box::new(storage));
+    /// Set cache policy for certificate and signing-key material reads.
+    pub fn crypto_material_cache_policy(mut self, policy: CryptoMaterialCachePolicy) -> Self {
```
Hermann-Core: Calling `.crypto_material_cache_policy(policy)` after `.crypto_material_storage(storage)` silently discards the provided policy.

In `crypto_material_storage(storage)`, `CryptoMaterialStorage` is immediately instantiated with whatever `self.cache_policy` is currently set to (defaulting to `NO_CACHE`). If a consumer subsequently calls `.crypto_material_cache_policy(policy)`, `self.cache_policy` is updated on the builder, but `self.crypto_material_storage` is not reconfigured, and `build()` takes the previously instantiated struct as-is.

To make builder method ordering robust, either retain the raw backend (`Option<Box<dyn Storage>>`) and construct `CryptoMaterialStorage::with_cache_policy` inside `build()`, or update the inner storage's cache policy when `crypto_material_cache_policy` is called.

src/setup.rs
```diff
@@ -520,6 +429,61 @@ fn acme_dns_credentials(account: &crate::config::AcmeDnsAccount) -> AcmeDnsCrede
     }
 }
 
+#[cfg(feature = "acme")]
+async fn build_crypto_material_storage(config: &AppConfig) -> EyeResult<Box<dyn Storage>> {
+    let backend = config.server.cert.material_backend.as_str();
```
Hermann-Core: In `store_certificate_strategy` (line 556), the match only checks `"filesystem"` and `"aws_secrets_manager"`, rejecting `"storage"` with an error:
`unsupported certificate store source 'storage'; expected 'filesystem' or 'aws_secrets_manager'`.

However, the README documentation and `StoreProvisioningStrategy::storage` explicitly advertise `server.cert.store.source = "storage"` for loading existing material from the configured material backend.

Please update `store_certificate_strategy` in `src/setup.rs` to accept `"storage"` as a supported source alongside `"aws_secrets_manager"`:

```rust
source if source.eq_ignore_ascii_case("aws_secrets_manager")
    || source.eq_ignore_ascii_case("storage") => {
```

src/utils/cert_manager/strategy.rs
```diff
@@ -156,25 +157,23 @@ impl StoreProvisioningStrategy {
                 certificate_key,
                 signing_key_key,
```
Hermann-Core: `StoreProvisioningSource::Storage` and `StoreProvisioningSource::SecretsStorage` contain duplicated logic in `load_material`. Now that certificate and secrets backends are consolidated into `material_storage`, both arms perform identical operations.

Please collapse these arms into a single match pattern:

```rust
StoreProvisioningSource::Storage {
    certificate_key,
    signing_key_key,
}
| StoreProvisioningSource::SecretsStorage {
    certificate_key,
    signing_key_key,
} => {
    let material_storage = manager.crypto_material_storage()?;
    let certificate = material_storage
        .load_secret(certificate_key)
        .await?
        .ok_or_else(|| {
            CertError::Validation(format!(
                "store certificate key '{certificate_key}' was not found"
            ))
        })?;
    let signing_key = material_storage
        .load_secret(signing_key_key)
        .await?
        .ok_or_else(|| {
            CertError::Validation(format!(
                "store signing key '{signing_key_key}' was not found"
            ))
        })?;
    Ok((
        decode_text_material(certificate, "certificate")?,
        decode_text_material(signing_key, "signing key")?,
    ))
}
```

src/utils/cert_manager/tests.rs
```diff
@@ -313,14 +308,56 @@ async fn test_signing_key_generation_and_storage() {
...
+    let stored_key = material_storage
         .load("keys-example.com")
         .await
         .unwrap()
         .unwrap();
     assert_eq!(key, stored_key);
 }
 
+#[tokio::test]
+async fn test_crypto_material_cache_policy_can_skip_private_key_cache() {
```
Hermann-Core: In `setup_test_metrics_registry()` (line 693), `let _ = setup_metrics(...)` immediately drops the returned `SdkMeterProvider`. This causes OpenTelemetry to shut down the provider, leading to intermittent race conditions and test failures (such as in `test_record_successful_renewal_updates_counters_and_gauge`) when running the test suite in parallel (`cargo test --lib --all-features`).

Please return `(prometheus::Registry, opentelemetry_sdk::metrics::SdkMeterProvider)` from `setup_test_metrics_registry()` so test cases retain ownership of `_meter_provider` for their duration, matching the pattern in `src/utils/cache/cert_chain.rs`.

docs/secrets-backends.md
```diff
@@ -1,8 +1,8 @@
-# Secrets Storage Backend Guidance
+# Cryptographic-Material Backend Guidance
 
-The server supports storing sensitive cryptographic materials (such as ACME account keys and server signing keys) in secure secrets management backends.
+The server stores lifecycle-coupled cryptographic material in one selected backend by default: ACME account keys, the server signing key, and the certificate chain associated with that key.
```
Hermann-Core: `docs/secrets-backends.md` and `.env.template` should be updated to reflect the new configuration keys for the consolidated cryptographic material backend:
- Document `APP_SERVER__CERT__MATERIAL_BACKEND` (`memory`, `aws_secrets_manager`, or `vault`), `APP_SERVER__CERT__MATERIAL_CACHE_TTL`, and `APP_SERVER__CERT__SIGNING_KEY_CACHE_TTL` in `docs/secrets-backends.md`.
- In `.env.template`, add these environment variables and clean up obsolete references to S3 certificate storage and split secrets configuration.

Hermann-Core (commented):
@Ngha-Boris review comments have been posted inline for your attention.

docs/secrets-backends.md
```diff
@@ -1,8 +1,8 @@
-# Secrets Storage Backend Guidance
+# Cryptographic-Material Backend Guidance
```
Hermann-Core:
```suggestion
# Cryptographic Material Backend Guidance
```

docs/secrets-backends.md
```diff
@@ -1,8 +1,8 @@
...
 
-The server supports storing sensitive cryptographic materials (such as ACME account keys and server signing keys) in secure secrets management backends.
+The server stores lifecycle-coupled cryptographic material in one selected backend by default: ACME account keys, the server signing key, and the certificate chain associated with that key.
 
-The supported secrets backends include:
+The supported material backends include:
```
Hermann-Core:
```suggestion
The supported backends include:
```

helm/README.md
```diff
@@ -35,9 +35,9 @@ The following files are used to configure the deployment:
 
 ## Redis Role
 
-Redis is optional. The server does not use Redis for status-list persistence or status-list reads; those use the configured repository backend and the in-process Moka status-list cache. In this chart, Redis is intended only as a distributed certificate-material cache for multi-replica deployments that use the AWS S3 certificate storage path and compile the application with the `redis` feature.
+Redis is optional. The server does not use Redis for status-list persistence or status-list reads; those use the configured repository backend and the in-process Moka status-list cache. Certificate and signing-key material are managed together by the selected cryptographic-material backend; prefer the built-in material read-cache TTLs before adding Redis.
```
Hermann-Core:
```suggestion
Redis is optional. The server does not use Redis for status-list persistence or status-list reads; those use the configured repository backend and the in-process status-list cache. Certificate and signing-key material are managed together by the selected cryptographic material backend; prefer the built-in material read-cache TTLs before adding Redis.
```

src/server/health.rs
```diff
@@ -229,21 +229,14 @@ impl ReadinessCheck for CertStoreCheck {
     }
 
     async fn check(&self) -> Result<(), String> {
-        let cert_store = self
+        let material_store = self
```
Hermann-Core: Consider using Rust way namings.
```suggestion
        let crypto_store = self
```

src/server/health.rs
```diff
@@ -229,21 +229,14 @@ impl ReadinessCheck for CertStoreCheck {
...
             .manager
-            .cert_storage()
-            .map_err(|e| format!("cert storage not configured: {e}"))?;
-        crate::cert_manager::storage::Storage::reachable(cert_store)
+            .crypto_material_storage()
```
Hermann-Core: Consider using Rust way namings.
```suggestion
            .crypto_storage()
```

src/utils/cert_manager/builder.rs
```diff
@@ -23,12 +25,11 @@ type ACMEHttpClientFactory = Box<dyn Fn() -> Box<dyn HttpClient> + Send + Sync>;
 /// - organization: none
 /// - EKU: none
 /// - ACME directory URL: empty string, but required when ACME is selected
-/// - storage backends, domains, and ACME challenge handler: not set
+/// - cryptographic-material backend, domains, and ACME challenge handler: not set
```
Hermann-Core:
```suggestion
/// - cryptographic material backend, domains, and ACME challenge handler: not set
```

src/utils/cert_manager/builder.rs
```diff
@@ -46,8 +47,7 @@ type ACMEHttpClientFactory = Box<dyn Fn() -> Box<dyn HttpClient> + Send + Sync>;
 ///     .email("support@example.com")
 ///     .organization(Some("example.com"))
 ///     .acme_directory_url("https://acme-v02.api.letsencrypt.org/directory")
-///     .cert_storage(cert_storage)
-///     .secrets_storage(secrets_storage)
+///     .crypto_material_storage(material_storage)
```
Hermann-Core:
```suggestion
///     .crypto_storage(material_storage)
```

src/utils/cert_manager/builder.rs
```diff
@@ -58,17 +58,16 @@ type ACMEHttpClientFactory = Box<dyn Fn() -> Box<dyn HttpClient> + Send + Sync>;
 /// ```ignore
 /// let manager = CertManager::builder()
 ///     .domains(["statuslist.example.com"])
-///     .cert_storage(cert_storage)
-///     .secrets_storage(secrets_storage)
+///     .crypto_material_storage(material_storage)
```
Hermann-Core:
```suggestion
///     .crypto_storage(material_storage)
```

src/utils/cert_manager/builder.rs
```diff
@@ -58,17 +58,16 @@ type ACMEHttpClientFactory = Box<dyn Fn() -> Box<dyn HttpClient> + Send + Sync>;
...
 ///     .store_strategy(StoreProvisioningStrategy::filesystem(
 ///         "/etc/status-list/tls.crt",
 ///         "/etc/status-list/tls.key",
 ///     ))
 ///     .build()?;
 /// ```
 pub struct CertificateManagerBuilder {
-    cert_storage: Option<Box<dyn Storage>>,
-    secrets_storage: Option<Box<dyn Storage>>,
+    crypto_material_storage: Option<CryptoMaterialStorage>,
```
Hermann-Core:
```suggestion
    crypto_storage: Option<CryptoStorage>,
```

src/utils/cert_manager/builder.rs
```diff
@@ -58,17 +58,16 @@ type ACMEHttpClientFactory = Box<dyn Fn() -> Box<dyn HttpClient> + Send + Sync>;
...
+    cache_policy: CryptoMaterialCachePolicy,
```
Hermann-Core:
```suggestion
    cache_policy: CryptoCachePolicy,
```

src/utils/cert_manager/builder.rs
```diff
@@ -84,8 +83,8 @@ pub struct CertificateManagerBuilder {
 impl Default for CertificateManagerBuilder {
     fn default() -> Self {
         Self {
-            cert_storage: None,
-            secrets_storage: None,
+            crypto_material_storage: None,
+            cache_policy: CryptoMaterialCachePolicy::NO_CACHE,
```
Hermann-Core:
```suggestion
            crypto_storage: None,
            cache_policy: CryptoCachePolicy::NO_CACHE,
```

src/utils/cert_manager/builder.rs
```diff
@@ -130,18 +129,39 @@ impl CertificateManagerBuilder {
         self
     }
 
-    /// Set the certificate storage backend.
-    pub fn cert_storage(mut self, storage: impl Storage + 'static) -> Self {
-        self.cert_storage = Some(Box::new(storage));
+    /// Set the consolidated backend for server certificate and signing-key material.
+    pub fn crypto_material_storage(mut self, storage: impl Storage + 'static) -> Self {
```
Hermann-Core:
```suggestion
    pub fn crypto_storage(mut self, storage: impl Storage + 'static) -> Self {
```

src/utils/cert_manager/builder.rs
```diff
@@ -130,18 +129,39 @@ impl CertificateManagerBuilder {
...
+        self.crypto_material_storage = Some(CryptoMaterialStorage::with_cache_policy(
```
Hermann-Core:
```suggestion
        self.crypto_storage = Some(CryptoStorage::with_cache_policy(
```

src/utils/cert_manager/builder.rs
Hermann-Core:
```suggestion
    pub fn crypto_cache_policy(mut self, policy: CryptoCachePolicy) -> Self {
```

src/utils/cert_manager/builder.rs
```diff
@@ -130,18 +129,39 @@ impl CertificateManagerBuilder {
...
+    /// Compatibility alias for the old split secrets storage setter.
+    ///
+    /// The certificate manager now uses the configured cryptographic-material
+    /// backend for both certificate data and signing keys. When used after
+    /// [`Self::cert_storage`], this backend wins so legacy call chains converge
+    /// on the selected secrets/material backend.
+    pub fn secrets_storage(self, storage: impl Storage + 'static) -> Self {
+        self.crypto_material_storage(storage)
+    }
+
```
Hermann-Core: Since they are now consolidated, we should avoid maintaining 2 separate conf that do the same thing and instead use the same method, unless cache policies must be isolated for key and certificate.

src/utils/cert_manager/builder.rs
```diff
@@ -192,11 +212,8 @@ impl CertificateManagerBuilder {
             ));
         }
 
-        let cert_storage = self.cert_storage.ok_or_else(|| {
-            CertError::Validation("certificate storage backend must be configured".to_string())
-        })?;
-        let secrets_storage = self.secrets_storage.ok_or_else(|| {
-            CertError::Validation("secrets storage backend must be configured".to_string())
+        let crypto_material_storage = self.crypto_material_storage.ok_or_else(|| {
+            CertError::Validation("cryptographic-material backend must be configured".to_string())
```
Hermann-Core:
```suggestion
        let crypto_storage = self.crypto_storage.ok_or(CertError::Validation("cryptographic-material backend must be configured".to_string())
```

src/utils/cert_manager/storage.rs
```diff
@@ -91,3 +93,156 @@ impl Storage for MemoryStorage {
...
     }
 }
+
+/// Cache policy for server cryptographic material.
+///
+/// Certificate material and private-key material are configured separately so
+/// deployments can cache public certificate chains while forcing each private
+/// key read to hit the backing secrets system.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CryptoMaterialCachePolicy {
```
Hermann-Core:
```suggestion
pub struct CryptoCachePolicy {
```

src/utils/cert_manager/storage.rs
```diff
@@ -91,3 +93,156 @@ impl Storage for MemoryStorage {
...
+
+impl Default for CryptoMaterialCachePolicy {
+    fn default() -> Self {
+        Self::NO_CACHE
+    }
+}
+
+/// Consolidated storage backend for the server certificate chain, signing key,
+/// and adjacent ACME account material.
+pub struct CryptoMaterialStorage {
```
Hermann-Core:
```suggestion
pub struct CryptoStorage {
```

src/utils/cert_manager.rs
```diff
@@ -121,10 +120,8 @@ type ACMEHttpClientFactory = Box<dyn Fn() -> Box<dyn HttpClient> + Send + Sync>;
 
 /// Struct representing the certificate manager
 pub struct CertManager {
-    // Certificate storage backend
-    cert_storage: Option<Box<dyn Storage>>,
-    // Secrets storage backend
-    secrets_storage: Option<Box<dyn Storage>>,
+    // Consolidated backend for certificate chain, signing key, and ACME account material.
+    crypto_material_storage: Option<CryptoMaterialStorage>,
```
Hermann-Core:
```suggestion
    crypto_storage: Option<CryptoMaterialStorage>,
```

src/utils/cert_manager.rs
```diff
@@ -173,8 +170,7 @@ impl CertManager {
         let cert_chain_cache = CertChainCache::new(DEFAULT_CHAIN_CACHE_TTL, domain_label);
 
         Ok(Self {
-            cert_storage: None,
-            secrets_storage: None,
+            crypto_material_storage: None,
```
Hermann-Core:
```suggestion
            crypto_storage: None,
```

src/utils/cert_manager.rs
```diff
@@ -192,17 +188,24 @@ impl CertManager {
...
+    /// Compatibility alias for the consolidated cryptographic-material backend.
+    ///
+    /// When used after [`Self::with_cert_storage`], this backend wins so legacy
+    /// call chains converge on the selected secrets/material backend.
+    pub fn with_secrets_storage(self, storage: impl Storage + 'static) -> Self {
+        self.with_crypto_material_storage(storage)
+    }
+
+    /// Set the consolidated backend for server certificate and signing-key material.
+    pub fn with_crypto_material_storage(mut self, storage: impl Storage + 'static) -> Self {
```
Hermann-Core:
```suggestion
    pub fn with_crypto_storage(mut self, storage: impl Storage + 'static) -> Self {
```

src/utils/cert_manager.rs
```diff
@@ -192,17 +188,24 @@ impl CertManager {
...
-    /// **Note:** This method is required.
-    pub fn with_secrets_storage(mut self, storage: impl Storage + 'static) -> Self {
-        self.secrets_storage = Some(Box::new(storage));
+    /// Compatibility alias for the consolidated cryptographic-material backend.
+    ///
+    /// When used after [`Self::with_cert_storage`], this backend wins so legacy
+    /// call chains converge on the selected secrets/material backend.
+    pub fn with_secrets_storage(self, storage: impl Storage + 'static) -> Self {
+        self.with_crypto_material_storage(storage)
+    }
```
Hermann-Core: Do we need these dead codes? I think they should be removed.

src/utils/cert_manager.rs
```diff
@@ -724,56 +726,59 @@ impl CertManager {
...
-        self.cert_storage
-            .as_deref()
-            .ok_or_else(|| CertError::Other(eyre!("Certificate storage not set")))
-    }
-
-    pub(crate) fn secrets_storage(&self) -> Result<&dyn Storage, CertError> {
-        self.secrets_storage
-            .as_deref()
-            .ok_or_else(|| CertError::Other(eyre!("Secrets storage not set")))
+    pub(crate) fn crypto_material_storage(&self) -> Result<&CryptoMaterialStorage, CertError> {
```
Hermann-Core:
```suggestion
    pub(crate) fn crypto_storage(&self) -> Result<&CryptoMaterialStorage, CertError> {
```

README.md
```diff
@@ -99,7 +99,7 @@ The crate uses modular Cargo feature flags to gate optional production backend d
 | `postgres` | SeaORM PostgreSQL database driver.                                                                                                                 | ❌ Opt-in  |
 | `sqlite`   | SeaORM SQLite database driver.                                                                                                                     | ❌ Opt-in  |
 | `mysql`    | SeaORM MySQL database driver.                                                                                                                      | ❌ Opt-in  |
-| `redis`    | Redis storage and certificate cache driver.                                                                                                        | ❌ Opt-in  |
+| `redis`    | Redis storage driver for legacy or explicit cache/storage integrations.                                                                            | ❌ Opt-in  |
```
Hermann-Core: Redis should not be treated as legacy but optional tool.
```suggestion
| `redis`    | Redis storage driver for explicit cache/storage integrations.                                                                            | ❌ Opt-in  |
```

src/setup.rs
Hermann-Core, line 1: Could we get rid of the `source` configuration and rely solely on feature selection together with the environment variables required by the chosen backend? I'm not sure the `source` configuration adds much value, since the backend is already determined by the enabled feature set. Removing it could simplify the configuration model, reduce duplication, and eliminate the possibility of configuration mismatches between the selected feature and the configured source.

Additionally, could you rename the `aws` feature to something more explicit, such as `aws-secrets`? The current name is quite generic and does not clearly communicate which AWS service it enables.

.env.template
Hermann-Core, line 1: Please make sure to update this file as well when addressing the comment in https://github.com/adorsys/status-list-server/pull/375/changes#r3795431618 so that the documentation remains consistent with the implementation changes.

docs/secrets-backends.md
Hermann-Core, line 1: Please make sure to update this file as well when addressing the comment in https://github.com/adorsys/status-list-server/pull/375/changes#r3795431618 so that the documentation remains consistent with the implementation changes.

Could you also consider using the same cache configuration across all backends? It would make the configuration model more consistent and easier to understand.

Hermann-Core (commented):
@Ngha-Boris Please check the additional comments I left below.

Could you also confirm that, for all supported backends, certificates and private keys follow the same storage model and behavior? I would like to ensure that the storage semantics remain consistent regardless of the selected backend.

docs/secrets-backends.md
```diff
@@ -1,27 +1,46 @@
...
+
 ## HashiCorp Vault & OpenBao (KV v2)
 
 Both HashiCorp Vault and OpenBao share the KV v2 REST API interface and are supported natively when the `vault` feature flag is enabled.
 
 ### Configuration Variables
 
+| Variable                                    | Type              | Default    | Description                                                                                                  |
+| ------------------------------------------- | ----------------- | ---------- | ------------------------------------------------------------------------------------------------------------ |
+| `APP_SERVER__CERT__MATERIAL_BACKEND`        | String            | `memory`   | Primary backend for server cryptographic material: `memory`, `aws_secrets_manager`, or `vault`.              |
```
Hermann-Core: What is the use of this variable since the source configuration has been removed?

docs/secrets-backends.md
```diff
@@ -1,27 +1,46 @@
...
+| In-memory             | Stored as a string under the certificate material key | Stored as a string under the signing-key material key | Volatile process-local storage for development and tests                                |
+| AWS Secrets Manager   | Stored as one secret string                           | Stored as one secret string                           | Same create, update, load, delete, cache, and invalidation path as certificate material |
+| Vault / OpenBao KV v2 | Stored as one KV v2 secret value                      | Stored as one KV v2 secret value                      | Same create, update, load, delete, cache, and invalidation path as certificate material |
+
+ACME provisioning writes the issued certificate chain and the signing key through this single abstraction. Store provisioning can seed material either from local files or from existing keys in the selected material backend, but the refreshed server material is persisted back through the same backend and invalidated together.
+
```
Hermann-Core: I dont think this section is useful since it leaks some implementation details. Can we remove it?

docs/secrets-backends.md
```diff
@@ -58,21 +76,22 @@ APP_VAULT__TOKEN=s.xxxxxxxxx
...
 
-When compiled with the `aws` feature flag:
+When compiled with the `aws-secrets` feature flag:
 
 ```env
 AWS_ACCESS_KEY_ID=test
 AWS_SECRET_ACCESS_KEY=test
 APP_AWS__REGION=us-east-1
-APP_AWS__SECRETS_CACHE_TTL=300
+APP_SERVER__CERT__MATERIAL_BACKEND=aws_secrets_manager
```
Hermann-Core: Is this still useful?

src/config.rs
```diff
@@ -189,6 +189,13 @@ pub struct CertConfig {
     #[serde(default)]
     pub eku: Vec<u64>,
     pub acme_directory_url: String,
+    /// Primary backend for server cryptographic material (`memory`,
+    /// `aws_secrets_manager`, or `vault` when compiled in).
+    pub material_backend: String,
```
Hermann-Core: What is the use of this configuration path?

src/config.rs
Hermann-Core, line 1: The goal here was to remove all dynamic and duplicated backend selection logic and rely solely on the enabled features and the environment variables required by the selected backend. Could you align this implementation with that approach?

src/setup.rs
Hermann-Core, line 1: See comment https://github.com/adorsys/status-list-server/pull/375/changes#r3795812315.

.env.template
Hermann-Core, line 1: See comment https://github.com/adorsys/status-list-server/pull/375/changes#r3795812315 and update the references accordingly.

src/config.rs
```diff
@@ -189,9 +189,10 @@ pub struct CertConfig {
     #[serde(default)]
     pub eku: Vec<u64>,
     pub acme_directory_url: String,
-    /// Cache TTL for parsed certificate chains in seconds.
-    /// A value of 0 keeps entries in memory indefinitely without expiration.
-    pub chain_cache_ttl: u64,
+    /// Cache TTL for certificate material reads in seconds. `0` disables this cache.
+    pub material_cache_ttl: u64,
```
Hermann-Core: What this still use for?

src/setup.rs
```diff
@@ -323,9 +231,11 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
             .email(&config.server.cert.email)
             .organization(config.server.cert.organization.as_deref())
             .acme_directory_url(&config.server.cert.acme_directory_url)
-            .cert_storage(cert_storage)
-            .secrets_storage(secrets_storage)
-            .chain_cache_ttl(Duration::from_secs(config.server.cert.chain_cache_ttl))
+            .crypto_cache_policy(CryptoCachePolicy::new(
+                Duration::from_secs(config.server.cert.material_cache_ttl),
```
Hermann-Core: What this still used for? The certificate is always cached and should be refresh/invalidated on renewal.

.env.template
```diff
@@ -23,26 +23,31 @@ APP_SERVER__CERT__ORGANIZATION=example.com
 APP_SERVER__CERT__ACME_DIRECTORY_URL=https://pebble:14000/dir
 # The last number is not explicitly defined in the spec
 APP_SERVER__CERT__EKU=1,3,6,1,5,5,7,3,30
+# The cryptographic material backend is selected by enabled Cargo features:
+# vault => Vault/OpenBao; aws-secrets without vault => AWS Secrets Manager;
+# neither feature => in-memory storage for local development and tests.
+# TTL in seconds for certificate material reads from the selected backend.
+# Set to 0 to disable this material cache.
+APP_SERVER__CERT__MATERIAL_CACHE_TTL=300
```
Hermann-Core: What this still used for? The certificate is always cached and should be refresh/invalidated on renewal.

src/utils/cert_manager/storage.rs
```diff
@@ -91,3 +93,160 @@ impl Storage for MemoryStorage {
...
+    pub async fn delete_secret(&self, key: &str) -> Result<(), StorageError> {
+        self.backend.delete(key).await
+    }
+
+    pub async fn invalidate_material(
+        &self,
+        certificate_key: &str,
+        signing_key_key: &str,
+    ) -> Result<(), StorageError> {
+        self.certificate_cache.invalidate(certificate_key).await;
```
Hermann-Core: **Logic inconsistency — `invalidate_material` evicts the certificate cache that `persist_certificate_data` just populated.**

The call chain for every provisioning/renewal is:

```rust
// persist_certificate_data:  backend.store + certificate_cache.insert  ✅
// persist_signing_key:        backend.store + signing_key_cache.insert  ✅
// refresh_material_cache:
//   invalidate_material  → certificate_cache.invalidate  ❌ (undoes the insert above)
//                        → signing_key_cache.invalidate  ❌ (undoes the insert above)
//   cache_provisioned_chain → cert_chain_cache.replace   ✅
```

After `invalidate_material` runs, `certificate_cache` is **empty** even though we just wrote the fresh cert into it. The next call to `certificate()` (triggered immediately by `renew_cert_if_needed` → `update_time_to_expiry`) therefore hits the backend unnecessarily, and the invariant *"certificate is always cached once provisioned/renewed"* is momentarily broken.

The backing-store write already happened and the cache already holds the freshest value. The correct behaviour for the renewal hook is to **replace** the cert-cache entry (it is already up-to-date) and to invalidate only the signing-key cache so that the next read re-fetches from the backend (or, if the signing key was also updated, to replace it too). There is no reason to evict `certificate_cache` here.

Suggested fix: remove the `certificate_cache.invalidate` from `invalidate_material` (or provide a dedicated `invalidate_signing_key` helper for the renewal hook). The `cert_chain_cache` replacement is the separate, correct action and should be kept.

src/utils/cert_manager.rs
```diff
@@ -799,6 +768,14 @@ impl CertManager {
         Ok(())
     }
 
+    /// One invalidation path for lifecycle-coupled certificate and key material.
+    async fn refresh_material_cache(&self, cert_pem: &str) -> Result<(), CertError> {
+        self.crypto_storage()?
+            .invalidate_material(&self.cert_key(), &self.signing_secret_id())
```
Hermann-Core: **`refresh_material_cache` evicts the `certificate_cache` entry that `persist_certificate_data` just wrote, breaking the "always cached" invariant.**

Tracing the full call chain for every provisioning / renewal:

```
persist_certificate_data  → CryptoStorage::certificate_cache.insert(fresh cert)   ✅
persist_signing_key       → CryptoStorage::signing_key_cache.insert(fresh key)     ✅
refresh_material_cache:
  invalidate_material     → certificate_cache.invalidate   ❌ undoes the insert above
                          → signing_key_cache.invalidate   ❌ undoes the insert above
  cache_provisioned_chain → cert_chain_cache.replace       ✅
```

After this function returns:
- `certificate_cache` is **empty** — the next `certificate()` call hits the backend needlessly.
- `signing_key_cache` is **empty** — even when the signing key did not change (ACME path).
- `cert_chain_cache` holds the fresh parsed chain — the TLS serving hot-path is fine.

The cert-data cache was already updated to the newest value by `persist_certificate_data`; there is nothing to "refresh". Invalidating it here only produces an avoidable round-trip on the very next `certificate()` call (which `renew_cert_if_needed` makes immediately afterwards to update the time-to-expiry gauge).

**Fix:** Remove the `certificate_cache.invalidate` from `invalidate_material`, or introduce a separate `invalidate_signing_key` method and use that here instead. `persist_certificate_data` already keeps `certificate_cache` coherent; `refresh_material_cache` only needs to populate `cert_chain_cache`.

src/utils/cert_manager/storage.rs
```diff
@@ -91,3 +93,160 @@ impl Storage for MemoryStorage {
...
+/// Consolidated storage backend for the server certificate chain, signing key,
+/// and adjacent ACME account material.
+pub struct CryptoStorage {
+    backend: Box<dyn Storage>,
+    certificate_cache: Cache<String, String>,
+    signing_key_cache: Option<Cache<String, String>>,
+}
+
+impl CryptoStorage {
+    const CACHE_MAX_CAPACITY: u64 = 16;
```
Hermann-Core: **`CACHE_MAX_CAPACITY: u64 = 16` is misleadingly large for a single-domain instance.**

One `CertManager` serves exactly one TLD+1 (as documented on `CertChainCache`). Therefore `CryptoStorage` will hold at most **one** `certificate_cache` entry and at most **one** `signing_key_cache` entry for the lifetime of the process. The same constant is reused for both caches (line 249 in `cache_for_ttl`), so the actual upper bound is `2` meaningful entries, but 16 slots are pre-allocated.

Using `16` wastes memory on every deployment, hides the single-entry contract from readers, and makes a future bug (e.g., accidentally keying by something other than the domain) harder to notice because the cache will silently grow.

Suggested values:
- `certificate_cache`: capacity `1` — one cert per domain, matches `CertChainCache::CACHE_CAPACITY`.
- `signing_key_cache`: capacity `1` — one signing key per domain.

Or, share a pair of explicit constants to make the intent clear:
```rust
const CERT_CACHE_CAPACITY: u64 = 1;
const KEY_CACHE_CAPACITY: u64 = 1;
```

Hermann-Core:
**`cert_manager.rs` – stale comment in `cert_chain_parts` (line ~459)**

```rust
// NOTE: There is a benign race between this read-path insert and
// the provisioning-path `replace` in `cache_provisioned_chain`.
// If a concurrent `request_certificate` replaces the cache entry
// between our miss and this insert, we overwrite the fresh chain
// with the (still valid) old chain. The stale entry is bounded by
// the cache TTL. This is acceptable because ...
```

`CertChainCache` is built with `max_capacity(1)` and **no** `time_to_live`. The phrase *"bounded by the cache TTL"* is incorrect — there is no TTL on this cache. A stale chain inserted by the read-path after a concurrent provisioning-path `replace` will stay until the **next** provisioning event. The comment should say *"persists until the next provisioning or renewal cycle"* instead of referencing a TTL that does not exist.

Hermann-Core (dismissed):

src/utils/cert_manager.rs
```diff
@@ -799,6 +768,14 @@ impl CertManager {
         Ok(())
     }
 
+    /// Refresh process-local caches after lifecycle-coupled certificate and key material changes.
+    async fn refresh_material_cache(&self, cert_pem: &str) -> Result<(), CertError> {
+        self.crypto_storage()?
+            .invalidate_signing_key(&self.signing_secret_id())
```
Christiantyemele: **Suggestion** - renewals/provisioning invalidate the private-key cache even when the key is unchanged.

`refresh_material_cache` calls `invalidate_signing_key`, evicting `signing_key_cache`. In the ACME renewal path the signing key is stable and was already populated in the cache, so this forces an avoidable backend round-trip on the very next `signing_key_pem()` read. In the store path, `persist_crypto_material` -> `persist_signing_key` just inserted the fresh key, then `refresh_material_cache` immediately evicts it. This is the same class of wasted-read that was fixed for the certificate/cert-chain caches; consider only evicting (or replacing) the signing-key cache entry when the key actually changed.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ sqlite = ["history", "sea-orm?/sqlx-sqlite"]
 mysql = ["history", "sea-orm?/sqlx-mysql"]
 redis = ["dep:redis"]
 vault = ["acme"]
```
Christiantyemele, line 14: **Question** - feature asymmetry between `vault` and `aws-secrets`.

`vault = ["acme"]` (line 14) forces the acme feature, but `aws-secrets` (line 15) does not imply `acme`. `build_crypto_storage` is gated on `#[cfg(feature = "acme")]` (setup.rs:432), so building with `--features aws-secrets` but without `acme` never actually uses Secrets Manager as the cryptographic-material backend, and the Route53 dependency in `aws-secrets` is likewise only meaningful under ACME. Should `aws-secrets` also imply `acme` for symmetry with `vault` (and so a bare `aws-secrets` build is not silently inert)?

Christiantyemele (commented):
.

Blindspot22 (dismissed):
apart from @Christiantyemele  suggestions...
I think this is good to go

Blindspot22:
@Hermann-Core Please resolve your conflicts

Blindspot22 (dismissed):

Christiantyemele (approved):

Hermann-Core:
@Blindspot22 conflicts have been resolved. Can you check again?

Blindspot22 (approved):
Looks good to me

martcpp:
if the ticket his block why his it in review where the blocker his not merge @Ngha-Boris

Hermann-Core:
> if the ticket his block why his it in review where the blocker his not merge @Ngha-Boris

It was a mistake. The ticket that was blocking it has been merged already so this can safely be merged.

PR #377, opened 2026-08-14
feat: implement gcp secret manager and azure key vault backends

### LINKED ISSUE #310: Implement GCP Secret Manager and Azure Key Vault backends
## Objective

Implement cloud-vendor secrets backends beyond AWS Secrets Manager.

## Context

The server currently supports AWS Secrets Manager only. To reduce AWS lock-in, add widely used managed alternatives while keeping backend selection consistent with the existing DNS-provider pattern.

## Deliverables

- [x] Implement Google Cloud Secret Manager storage.
- [x] Implement Azure Key Vault secret storage.
- [x] Reuse the provider selector introduced for #249 sub-issues.
- [x] Support TTL-based in-process caching with clear invalidation on update/delete.
- [x] Add tests using wiremock or provider emulators where practical.
- [x] Document required IAM/role permissions with least-privilege examples.

## Acceptance Criteria

- [x] GCP Secret Manager and Azure Key Vault are selectable secrets backends.
- [x] Existing AWS Secrets Manager support remains intact.
- [x] Secret lifecycle operations map consistently to `store`, `load`, `update`, and `delete`.
- [x] Docs include full env/config examples.

## Verification

- [x] Adapter tests
- [x] Config tests

Estimated time: 2 days

Hermann-Core:
## Objective

Implemented cloud-vendor secrets backends beyond AWS Secrets Manager.

## Acceptance Criteria

- [x] GCP Secret Manager and Azure Key Vault are selectable secrets backends.
- [x] Existing AWS Secrets Manager support remains intact.
- [x] Secret lifecycle operations map consistently to `store`, `load`, `update`, and `delete`.
- [x] Docs include full env/config examples.


src/config.rs
```diff
@@ -840,6 +878,10 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("vault.namespace", Option::<String>::None)?
         .set_default("vault.secrets_cache_ttl", 300)?
         .set_default("vault.timeout_secs", 30)?
+        .set_default("gcp_secret_manager.project_id", "")?
+        .set_default("gcp_secret_manager.secrets_cache_ttl", 300)?
+        .set_default("azure_keyvault.vault_url", "")?
```
Ngha-Boris: This breaks configuration loading for *all* users, regardless of the `azure-kv` feature.

`AzureKeyVaultConfig.vault_url` is typed `url::Url` (line 646) and is an **unconditional, non-feature-gated** field on `Config` (lines 88–91). Defaulting it to `""` yields a value that cannot be deserialized into `Url` (`Url::parse("")` fails), so every `AppConfig::load_from_overrides` call errors. I reproduced it:

```
test config::tests::test_config_loading ... FAILED
Failed to load default config: relative URL without a base: "" for key `azure_keyvault.vault_url`
```

The PR's verification only ran `cargo check`, which never executes this test, so the regression slipped through.

**Suggestion:** Make the field `Option<url::Url>` with `#[serde(default)]` (and similarly validate it is present only when `azure-kv` is compiled in), instead of defaulting a `Url` field to `""`. Add an assertion to `test_config_loading` covering the new config structs so this is caught in CI.

src/config.rs
```diff
@@ -840,6 +878,10 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("vault.namespace", Option::<String>::None)?
         .set_default("vault.secrets_cache_ttl", 300)?
         .set_default("vault.timeout_secs", 30)?
+        .set_default("gcp_secret_manager.project_id", "")?
```
Ngha-Boris, line 913: **Validation** `project_id` (and `vault_url` above) are documented as "Required when the backend is enabled" but have no enforcement, and silently default to empty.

If `gcp-secrets` is compiled in with an empty `project_id`, the resource-name helpers (`gcp_secret.rs` lines 43–45) produce malformed names like `projects//secrets/foo` and a client that fails only at request time with a confusing error.

**Suggestion:** Add a startup validation step (analogous to `store_certificate_strategy`) that errors out when the compiled-in feature lacks its required field — e.g. error if `gcp-secrets` is active and `project_id` is empty, or `azure-kv` is active and `vault_url` is absent.

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,292 @@
...
+        } else {
+            DeveloperToolsCredential::new(None).map_err(|e| {
+                StorageError::Backend(eyre!(
+                    "failed to create Azure DeveloperToolsCredential: {e}"
+                ))
+            })?
+        };
+
+        let mut options = azure_security_keyvault_secrets::SecretClientOptions {
+            verify_challenge_resource: Some(false),
```
Ngha-Boris: **Security** this disables Key Vault WWW-Authenticate **challenge verification**.

`verify_challenge_resource: Some(false)` turns off the SDK's protection against token-audience mismatch / a spoofed challenge from the vault URL. It is applied unconditionally in the production code path, with no comment explaining why, rather than only for test/emulator scenarios.

**Suggestion:** Keep verification enabled (default `Some(true)`) in production and disable it only in test construction paths (e.g. a `#[cfg(test)]` branch or dedicated test helper that sets a mock `http_client`).

src/outbound/gcp_secret.rs
```diff
@@ -0,0 +1,320 @@
...
+            client,
+            project_id: self.project_id,
+            cache,
+        })
+    }
+}
+
+#[async_trait]
+impl Storage for GcpSecretManagerClient {
+    async fn store(&self, key: &str, value: &str) -> Result<(), StorageError> {
```
Ngha-Boris, line 279: **Consistency** the `store` semantics here diverge from the AWS backend and are internally inconsistent with `update`.

- AWS `store` is a *create-if-not-exists* no-op on existing secrets (`aws.rs:60-63` logs "already exists. Skipping...").
- GCP `store` **always** adds a new secret version (overwrites on every call).
- GCP `update` (line 233) only calls `add_secret_version` and **fails if the secret container does not exist**, whereas `store` handles the create path.

This conflicts with the ticket's "Secret lifecycle operations map consistently to `store`, `load`, `update`, and `delete`."

**Suggestion:** Pick one policy and make `store` and `update` behave identically (both create-or-update). Extract the get → create-if-404 → add-version logic into a private helper used by both methods so they can't drift.

src/outbound/gcp_secret.rs
```diff
@@ -0,0 +1,320 @@
...
+        let secret_name = self.secret_name(key);
+
+        // Check if secret exists
+        let resp = self.client.get_secret().set_name(&secret_name).send().await;
```
Ngha-Boris: **Race condition (TOCTOU)** `get_secret` → `create_secret` on 404 → `add_secret_version` is a check-then-act.

Two concurrent `store` calls (e.g. during parallel cert provisioning) can both observe 404 and both invoke `create_secret`, one failing with an AlreadyExists error.

**Suggestion:** Treat an AlreadyExists error from `create_secret` as benign (match on the specific error instead of only the 404-on-`get`), or retry the create on that status so concurrent stores are idempotent.

src/outbound/gcp_secret.rs
```diff
@@ -0,0 +1,320 @@
...
+        StorageError::Backend(error.into())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_client_builder_and_resource_names() {
```
Ngha-Boris, line 395: **Tests** — the ticket's deliverable "Add tests using wiremock or provider emulators where practical" and the Verification item "Adapter tests" are not met.

Only a builder/field test exists here; there are no tests for `store`/`load`/`update`/`delete`, no emulator tests, and no cache-invalidation coverage for update/delete (an explicit deliverable). The builder already exposes the needed seam — `endpoint` + anonymous credentials (lines 107–111) — that is never exercised.

**Suggestion:** Use the `endpoint`/anonymous-creds seam against a local emulator (or a small token server) to cover not-found → `Ok(None)`, create-on-store, update with cache invalidation, and delete with cache invalidation.

Ngha-Boris (commented):
@Hermann-Core please consider the comments below

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,292 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_client_builder() {
```
Ngha-Boris, line 445: **Tests** — the ticket's "Add tests using wiremock or provider emulators where practical" deliverable and the "Adapter tests" verification item are unmet.

Only `test_client_builder` exists; there are no `store`/`load`/`update`/`delete` behavior tests and no cache-invalidation tests. The builder exposes `custom_credential` and `http_client` seams (lines 121, 105) clearly intended for mocking but never used by any test.

**Suggestion:** Add adapter tests that inject a mock `HttpClient` (e.g. a `wiremock`-backed client that satisfies `azure_core::http::HttpClient`) plus a stub `TokenCredential`, and assert not-found → `Ok(None)`, create-on-store, and cache invalidation on update/delete. This would also let you keep challenge verification enabled while testing.

src/setup.rs
```diff
@@ -232,35 +246,82 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
                         .build()?,
                 )
             }
-            #[cfg(not(feature = "vault"))]
+            #[cfg(all(feature = "gcp-secrets", not(feature = "vault")))]
             {
-                #[cfg(feature = "aws")]
+                tracing::info!("Using GCP Secret Manager as secrets backend");
```
Ngha-Boris: **Design** — the ticket claims to "Reuse the provider selector introduced for #249" and keep selection consistent with the DNS-provider pattern (runtime selection), but the secrets backend is chosen by a **compile-time `#[cfg]` chain** (vault > gcp > azure > aws) keyed off feature flags, not runtime config.

Meanwhile the AWS branch (lines 287–303) reads the runtime `store.source == "aws_secrets_manager"` value, and `store_certificate_strategy` accepts `gcp_secret_manager`/`azure_keyvault` as source strings. The result is inconsistent: setting `source=azure_keyvault` while built with `--features gcp-secrets` silently uses GCP, and the runtime `source` value is effectively ignored for GCP/Azure.

**Suggestion:** Make backend selection runtime-based and consistent with the AWS branch — inspect `server.cert.store.source` and error if the requested backend was not compiled in (mirror the `#[cfg(not(feature = ...))]` error pattern used in `build_dns_challenge_handler`), or explicitly document and enforce the feature-flag-only selection so the two mechanisms can't silently conflict.

Hermann-Core:
@Ngha-Boris can you check this back?

src/outbound/gcp_secret.rs
Hermann-Core: Behavior is tested in integration tests under the `tests` folder.

src/outbound/azure_kv.rs
Hermann-Core: Behavior is tested in integration tests under the tests folder.

src/setup.rs
Hermann-Core: The configuration should be compile time and must only pulls in the required dependencies when required. The `store.source` must be removed and in tracked in another issue.

Hermann-Core:
@Christiantyemele @Blindspot22 can you check this so we merge?

Blindspot22:
@Hermann-Core Please resolve your conflicts here as well

Blindspot22 (dismissed):

src/outbound/gcp_secret.rs
```diff
@@ -0,0 +1,372 @@
...
+                )))
+            }
+        }
+    }
+
+    async fn update(&self, key: &str, value: &str) -> Result<(), StorageError> {
+        self.ensure_secret_and_add_version(key, value).await
+    }
+
+    async fn delete(&self, key: &str) -> Result<(), StorageError> {
```
Christiantyemele, line 326: **Important — readiness probe silently always succeeds for the GCP backend.**

`Storage` (src/utils/cert_manager/storage.rs:41) gives `reachable()` a default `Ok(())` implementation, and this `impl Storage for GcpSecretManagerClient` never overrides it. Every other backend implements a real probe — AWS (aws.rs:141), Vault (vault.rs:385), and Azure (azure_kv.rs:280, a `get_secret("__health_probe_test__")`).

Consequence: on a GCP deployment with bad credentials, wrong project, or no connectivity, the readiness/liveness check (`reachable()`) reports the backend as up, so traffic is routed / the container is kept healthy against a secrets backend that will fail on the first real `load()`. This defeats the purpose of the probe for this new backend.

**Suggestion:** implement `reachable()` on the GCP adapter (e.g. `access_secret_version` on a well-known probe key and treat a 404 as reachable, mirroring Azure's `__health_probe_test__` approach and the emulator `endpoint` seam you already expose for tests).

Christiantyemele (changes requested):
.

Christiantyemele (dismissed):

Hermann-Core:
@Blindspot22 can you recheck and approve if everything Ok?

Christiantyemele (dismissed):

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,326 @@
...
+                    Err(StorageError::Backend(eyre!(
+                        "Azure secret delete failed for key '{key}': {err:?}"
+                    )))
+                }
+            }
+        }
+    }
+
+    async fn reachable(&self) -> Result<(), StorageError> {
+        match self.client.get_secret("__health_probe_test__", None).await {
```
martcpp: `__health_probe_test__` is not a legal Key Vault object name. Per the [KV docs](https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates), names must be "a 1-127 character string, containing only 0-9, a-z, A-Z, and -"  underscores are rejected.

So this returns `400 Bad Request`, not `404`. The `NotFound` branch below never fires, `reachable()` always errors, and `/health/ready` returns 503 permanently — the pod never passes its readiness gate.

Pick a name legal in both GCP and Azure charsets, e.g. `health-probe-do-not-create`.

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,326 @@
...
+                }
+            }
+        }
+    }
+
+    async fn update(&self, key: &str, value: &str) -> Result<(), StorageError> {
+        self.store(key, value).await
+    }
+
+    async fn delete(&self, key: &str) -> Result<(), StorageError> {
```
martcpp, line 392: This treats `delete` as destructive, but Key Vault soft-delete is [on by default and cannot be disabled](https://learn.microsoft.com/en-us/azure/key-vault/general/soft-delete-overview) on any vault created since 2020. The object stays in a soft-deleted state for 7–90 days (default 90), and per the same doc "a key vault object can't be created in a given vault if that key vault contains an object with the same name and which is in a deleted state."

Suggested: `delete_secret` → poll `get_deleted_secret` until the deletion is durable → `purge_deleted_secret`, with a clear error when purge protection makes purge impossible. That needs `Microsoft.KeyVault/vaults/secrets/purge/action` ("Key Vault Purge Operator") added to the IAM docs, and the purge-protection limitation stated explicitly —on such a vault `Storage::delete` cannot satisfy the trait's implied contract at all.

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,326 @@
...
+            info!("Azure Key Vault cache disabled (TTL=0)");
+        }
+
+        Ok(AzureKeyVaultClient { client, cache })
+    }
+}
+
+#[async_trait]
+impl Storage for AzureKeyVaultClient {
+    async fn store(&self, key: &str, value: &str) -> Result<(), StorageError> {
```
martcpp, line 262: Key names reaching this method come from `CertManager` and contain dots:
- `cert_manager.rs:821` → `signing_secret_id()` = `keys-status.example.com`
- `cert_manager.rs:816` → `acme_account_id()` = `acme_accounts-example.com`

Key Vault names allow only `[0-9a-zA-Z-]`, so both are rejected with `400`. Every production deployment fails on its first secret write. This works today only because AWS allows `/_+=.@-` and Vault uses path segments.

Needs a normalisation step, e.g.:
```rust
/// Azure Key Vault object names admit only [0-9a-zA-Z-], max 127 chars.
fn qualify_key(key: &str) -> String {
    key.chars().map(|c| if c.is_ascii_alphanumeric() { c } else { '-' }).collect()
}

File: src/outbound/azure_kv.rs [hunk 13]
```diff
@@ -0,0 +1,326 @@
...
+            ClientSecretCredential::new(
+                tenant_id.as_str(),
+                client_id.clone(),
+                AzureSecret::new(client_secret.expose_secret().to_string()),
+                None,
+            )
+            .map_err(|e| {
+                StorageError::Backend(eyre!("failed to create Azure ClientSecretCredential: {e}"))
+            })?
+        } else {
```
martcpp, line 225: no environment variables, no managed identity, no workload identity. So AKS + workload identity, the most common production Azure deployment, cannot authenticate at all. docs/secrets-backends.md:212 claims this fallback reads

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,326 @@
...
+                    cache.insert(key.to_string(), val.clone()).await;
+                }
+                Ok(value)
+            }
+            Err(err) => {
+                if err.http_status() == Some(StatusCode::NotFound) {
+                    Ok(None)
+                } else {
+                    Err(StorageError::Backend(eyre!(
+                        "Azure secret load failed for key '{key}': {err:?}"
```
martcpp: {err:?}  Debug on azure_core::Error tends to carry request URLs, headers, and sometimes response bodies, and it isn't a stable contract across dependency bumps. Prefer {err} (Display), matching what the GCP, Vault, and AWS adapters do. Same at lines 274 and 289.

Cargo.toml
```diff
@@ -173,10 +196,10 @@ features = ["alloc", "ecdsa", "pem", "pkcs8"]
 [dev-dependencies]
```
martcpp, line 202: Unrelated to the feature and unexplained  reads as a no-op reformat of testcontainers-modules from an inline table to a section. Fine if intentional, just costs a reviewer a moment to confirm.

docs/secrets-backends.md
```diff
@@ -159,6 +161,74 @@ Vault authentication and token operations emit OpenTelemetry counters for succes
...
+
+### Authentication & RBAC Permissions
+
+If `tenant_id`, `client_id`, and `client_secret` are provided, the adapter authenticates as a Service Principal. Otherwise, it falls back to Azure developer/workload identity credentials (`AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, or Azure CLI credentials).
+
+Required RBAC roles (least-privilege):
+
+- **Key Vault Secrets Officer** (full CRUD access to secrets)
+- Or custom role with data actions:
+  - `Microsoft.KeyVault/vaults/secrets/read`
```
martcpp: An operator following this doc gets a role that grants no data access and a 403 they cannot diagnose. Also missing: the vault must be in RBAC authorization mode (not legacy access policies) for role assignments to apply at all.

tests/gcp_secrets_manager.rs
```diff
@@ -0,0 +1,133 @@
...
+        .expect("Failed to build GcpSecretManagerClient");
+
+    (container, client)
+}
+
+#[tokio::test]
+async fn test_gcp_secret_manager_storage_lifecycle() {
+    let (_container, client) = start_emulator(Duration::ZERO).await;
+
+    let secret_key = "test-cert-key";
```
martcpp: test-cert-key is charset-legal in both clouds, but the names CertManager actually generates are not  keys-status.example.com and acme_accounts-example.com (cert_manager.rs:821, :816) both contain dots, which GCP and Azure reject with 400.

This is why the suite can't catch the charset blocker: it's structurally incapable of it. Any lifecycle test for a secrets backend should use CertManager's real key format.

martcpp (changes requested):
@Hermann-Core  drop some review

src/outbound/gcp_secret.rs
```diff
@@ -0,0 +1,387 @@
...
+            .send()
+            .await
+        {
+            Ok(_) => {
+                if let Some(cache) = &self.cache {
+                    cache.insert(key.to_string(), value.to_string()).await;
+                }
+                Ok(())
+            }
+            Err(e) if e.http_status_code() == Some(404) => {
```
martcpp: http_status_code() is likely the wrong discriminator for NOT_FOUND  if so, the GCP adapter fails comprehensively against real GCP.

Per the [gax error docs](https://docs.rs/google-cloud-gax/latest/google_cloud_gax/error/struct.Error.html), http_status_code() is populated when "the error is generated before reaching Google Cloud services. For example, your proxy or Google load balancers may generate errors without detailed payloads." For genuine service errors the AIP-193 Status is the populated field, and the documented idiom is


can you map to exist error code in the code base

martcpp (changes requested):

Cargo.toml
Hermann-Core: This was intentional for readability.

Hermann-Core:
@martcpp thanks for the reviews, can you check again?

Hermann-Core:
@martcpp can you check back this PR so we merge?

martcpp:
> @martcpp can you check back this PR so we merge?

i am on it

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    async fn delete(&self, key: &str) -> Result<(), StorageError> {
+        let secret_name = Self::qualify_key(key);
+        match self.client.delete_secret(&secret_name, None).await {
+            Ok(_) => {
+                if let Some(cache) = &self.cache {
+                    cache.invalidate(key).await;
+                }
+                // Try purging the soft-deleted secret if purge is allowed on the vault
+                let _ = self.client.purge_deleted_secret(&secret_name, None).await;
```
martcpp: 🟠 The purge is fire-and-forget, so the original `409` still happens just less often.

```rust
let _ = self.client.purge_deleted_secret(&secret_name, None).await;

File: src/outbound/azure_kv.rs [hunk 20]
```diff
@@ -0,0 +1,370 @@
...
+            ClientSecretCredential::new(
+                tenant_id.as_str(),
+                client_id.clone(),
+                AzureSecret::new(client_secret.expose_secret().to_string()),
+                None,
+            )
+            .map_err(|e| {
+                StorageError::Backend(eyre!("failed to create Azure ClientSecretCredential: {e}"))
+            })?
+        } else if let Ok(cred) = WorkloadIdentityCredential::new(None) {
```
martcpp: The chain probes at construction time, not token-acquisition time  this likely makes the DeveloperTools branch unreachable.

*::new() succeeding means "the credential object was constructed", not "this credential can obtain a token." ManagedIdentityCredential::new(None) doesn't probe IMDS at construction, so it should succeed essentially everywhere — including on a developer laptop, where it then fails at the first token request with no fallback to az login. That regresses local development, and the resulting error won't point at the cause.

DefaultAzureCredential exists for exactly this: it chains environment → workload identity → managed identity → developer tools and falls through at token acquisition, which is the only point where "can this credential actually authenticate" is answerable. Recommend using it and deleting the hand-rolled chain.

src/utils/cert_manager/storage.rs
```diff
@@ -65,6 +65,31 @@ impl<T: Storage + ?Sized> Storage for Box<T> {
...
 }
 
+/// Normalize a storage key so it is valid and consistent across all secrets backends
+/// (AWS Secrets Manager, GCP Secret Manager, Azure Key Vault, HashiCorp Vault, Memory).
+///
+/// Azure Key Vault is the most restrictive provider, requiring 1–127 characters
+/// and admitting only ASCII alphanumeric characters and hyphens `[0-9a-zA-Z-]`.
+/// This function replaces any character outside `[0-9a-zA-Z-]` with `-` and truncates
+/// to 127 characters.
+pub fn normalize_key(key: &str) -> String {
```
martcpp, line 83: Collision risk is unaddressed, and the truncation makes it reachable.

normalize_key is many-to-one, so distinct logical keys can share a backend name  keys-a.b and keys-a-b both become keys-a-b. For the current fixed key set that's unlikely to bite on its own.

The 127-char truncation at line 86 is the part I'd worry about. signing_secret_id() is keys-{domains.join("-")} (cert_manager.rs:821), so a multi-domain deployment can exceed 127 characters, and any two domain sets sharing a 127-char prefix silently collide — one deployment's signing key overwrites another's, with no error anywhere.


extracting this into storage.rs and applying it to MemoryStorage as well was the right shape  nice that dev/prod now agree on key semantics.

src/config.rs
martcpp, line 1: "Azure Key Vault _configuration error: azure_keyvault.vault_url is required when azure-kv feature is enabled

src/outbound/gcp_secret.rs
```diff
@@ -0,0 +1,444 @@
...
+    /// Build the [`GcpSecretManagerClient`] adapter instance.
+    pub async fn build(self) -> Result<GcpSecretManagerClient, StorageError> {
+        if self.project_id.trim().is_empty() {
+            return Err(StorageError::Backend(eyre!(
+                "GCP Secret Manager configuration error: project_id must not be empty"
+            )));
+        }
+
+        let mut builder = SecretManagerService::builder();
+        if let Some(endpoint) = &self.endpoint {
```
martcpp, line 232: Setting an endpoint and disabling authentication are unrelated decisions, welded together here for one test's convenience. APP_GCP_SECRET_MANAGER__ENDPOINT is a real operator-settable key with legitimate production uses  regional endpoints (secretmanager.europe-west4.rep.googleapis.com, required for data residency) and Private Service Connect / VPC-SC.

Any operator setting it silently loses all authentication. That mattered before; after the #375 merge unified cert_storage and secrets_storage into a single crypto_storage, this backend now holds all cryptographic material — signing key, ACME account, and cert chains — so the blast radius grew.

It's also still absent from the config table in docs/secrets-backends.md, so the behaviour remains invisible to whoever sets it.

 an explicit allow_anonymous_credentials: bool (default false, doc-commented "testing/emulator only"), or gate the anonymous path on cfg(test) / APP_ENV != production. And document endpoint either way.

docs/secrets-backends.md
```diff
@@ -160,6 +162,85 @@ Vault authentication and token operations emit OpenTelemetry counters for succes
...
+### Authentication & IAM Permissions
+
+By default, authentication uses **Application Default Credentials (ADC)** (discovering environment credentials, GKE workload identity, service account metadata server, etc.).
+
+Required IAM roles (least-privilege):
+
+- `roles/secretmanager.secretAccessor` (read secrets)
+- `roles/secretmanager.secretVersionManager` (create/update secret versions)
+- `roles/secretmanager.viewer` (read secret metadata / readiness health check)
+- `roles/secretmanager.admin` or custom role with `secretmanager.secrets.create` and `secretmanager.secrets.delete` (if automated secret creation/deletion is used)
```
martcpp: The Azure half of this section got the least-privilege treatment in the last round — this GCP half didn't.

roles/secretmanager.admin is the opposite of least-privilege: full control including versions.access on every secret in the project, offered as the primary option with "or custom role" as an unelaborated afterthought.

src/setup.rs
martcpp, line 1: uild_crypto_storage is a real improvement over the inline block — extracted, readable, one place to look. The precedence cascade underneath it still stands though: vault > gcp-secrets > azure-kv > aws-secrets > memory, resolved entirely at compile time.

Enabling two backends still silently discards the lower-precedence one, with no warning — and --all-features builds a binary where three of four are dead code. The issue asked to reuse the runtime DNS-provider selector (DnsProviderKind + DnsConfig::resolve, config.rs:219-267 / 493-549), which would also let one binary serve multiple clouds  the stated "reduce AWS lock-in" goal.

If the full selector is out of scope for this PR, the cheap mitigation is a compile_error! when mutually exclusive features are co-enabled, so the silent drop becomes a build failure. Worth an issue either way so it doesn't get lost.

martcpp (changes requested):
@Hermann-Core left some comment

src/setup.rs
Hermann-Core: Feature flags were chosen specifically to only build the required cloud SDK and avoid runtime selection that forces pulling and linking every cloud provider SDK into every binary. The additive cascade in `build_crypto_storage` is the appropriate choice because enforcing `compile_error!` makes `--all-features` fail by definition, breaking workspace CI, clippy, and cargo check. Full configuration consolidation will be handled in #168.

Hermann-Core:
@martcpp can you check again?

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,384 @@
+//! Azure Key Vault secret storage adapter implementing [`Storage`].
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use azure_core::credentials::{Secret as AzureSecret, TokenCredential};
+use azure_core::http::{HttpClient, StatusCode};
+use azure_identity::ClientSecretCredential;
+pub use azure_identity_helpers::default_azure_credential::DefaultAzureCredential;
```
martcpp: Write the "try each in turn" logic in your own codebase. It's about thirty lines, uses only Microsoft's official components, and removes the outside dependency entirely. You end up with less trust placed in strangers, not more. this crate is not official form Microsoft and it not that strong yet we can trust it @Hermann-Core

src/outbound/azure_kv.rs
```diff
@@ -0,0 +1,384 @@
...
+                })?,
+                None,
+            )
+            .await;
+
+        match set_res {
+            Ok(_) => {}
+            Err(err) if err.http_status() == Some(StatusCode::Conflict) => {
+                // If the secret is in soft-deleted state, recover it and retry setting the secret
+                let _ = self.client.recover_deleted_secret(&secret_name, None).await;
```
martcpp: 🟠 Same async-operation race as the purge `recover` is fire-and-forget, then retried immediately.

Key Vault recovery is asynchronous: when `recover_deleted_secret` returns, the secret isn't usable yet — it takes a few seconds to become available. The `set_secret` retry fires straight after, so it will often still fail, and when it does the operator sees the original error with no hint that a recovery was attempted and lost the race.

`let _ =` also discards the outcome entirely. If recovery fails because the identity lacks the permission, that's invisible.

Fix: poll until the operation actually resolves (bounded, with a timeout) before retrying the set, and log the failure reason instead of discarding it.

Note the two paths now interact — `delete` fire-and-forget purges at line 296, and `store` fire-and-forget recovers here. If the purge silently failed earlier, this recover path is what catches it. Both being racy means the failure mode is "works most of the time," which is the hardest kind to debug. Getting one of the two right would be enough; both would be better.

The `409` detection and recover-then-retry structure is the right idea it's just the timing that needs handling.

src/utils/cert_manager/storage.rs
```diff
@@ -65,6 +65,42 @@ impl<T: Storage + ?Sized> Storage for Box<T> {
...
+///
+/// Azure Key Vault is the most restrictive provider, requiring 1–127 characters
+/// and admitting only ASCII alphanumeric characters and hyphens `[0-9a-zA-Z-]`.
+/// This function replaces any character outside `[0-9a-zA-Z-]` with `-`.
+///
+/// If the sanitized key exceeds 127 characters, it is truncated to 110 characters
+/// and appended with a hyphen and a 16-character hex SHA-256 digest of the full original
+/// key (`110 + 1 + 16 = 127` chars). This prevents collisions between distinct keys
+/// that share a long prefix (e.g. multi-domain signing secret keys).
+pub fn normalize_key(key: &str) -> String {
```
martcpp, line 83: Two residuals  the truncation fix itself is good.

The sha256 suffix solves the case I flagged (distinct long keys sharing a 127-char prefix), and the test proves it.

1. **The doc comment overstates the guarantee.** It says the digest "prevents collisions between distinct keys" — but that only holds on the `> 127` path. Under 127 chars the charset mapping is still many-to-one and unhashed: `keys-a.b` and `keys-a-b` both normalize to `keys-a-b` with no disambiguation. Practically unreachable for the current key set, so I'd leave the behaviour alone — but please narrow the doc to say collisions are prevented *for truncated keys*, so the next reader doesn't rely on a stronger promise than the code makes.

2. **`&sanitized[..110]` at line 97 is an unguarded byte slice.** It's safe *only because* the map above replaced every non-ASCII char, making byte 110 always a char boundary. One comment stating that invariant would stop a future reorder from turning this into a panic.

Cosmetic: if `sanitized[..110]` ends in `-`, the result contains `--`. Both legal characters, no functional issue.

martcpp (changes requested):

src/outbound/azure_kv.rs
martcpp: or best do this  rather than that stress

`APP_AZURE_KEYVAULT__CREDENTIAL = service_principal | workload_identity | managed_identity | azure_cli`

It's what Microsoft did. They dropped DefaultAzureCredential at GA deliberately. Implicit chaining is convenient in dev and a liability in prod  when auth fails you don't know which link failed, and worse, when it succeeds you may not know which identity you're running as. Explicit config removes both.

Hermann-Core:
@martcpp you can recheck

Christiantyemele (approved):

martcpp (approved):

PR #378, opened 2026-08-14
feat: add vault and openbao approle authentication

### LINKED ISSUE #360: Add Vault and OpenBao AppRole authentication
## Objective

Add Vault/OpenBao AppRole authentication in replacement to token-based authentication.

## Context

Vault/OpenBao token authentication is useful for local tests and simple setups, but AppRole is a common production machine-to-machine authentication mechanism. It lets operators provision a `role_id` and protected `secret_id`, issue scoped tokens, and rotate credentials without embedding long-lived root/operator tokens in the server environment.

## Deliverables

- [x] Add AppRole settings: `role_id`, `secret_id` or secret reference, mount path, token TTL/renewal behavior, namespace where applicable.
- [x] Exchange AppRole credentials for a Vault client token at startup.
- [x] Renew the Vault token when renewable, or reacquire it before expiry.
- [x] Fail safely if AppRole login/renewal fails, without logging secrets.
- [x] Add tests for successful login, forbidden login, token renewal, expired token, and redaction.
- [x] Document least-privilege Vault policy examples for AppRole.

## Acceptance Criteria

- [x] Vault/OpenBao backends work with token auth and AppRole auth.
- [x] AppRole credentials are treated as secrets and can themselves be loaded through the runtime credential mechanism where practical.
- [x] Token renewal/reacquisition is observable through safe logs/metrics.
- [x] Failure modes are clear and do not leak `secret_id`, tokens, signing keys, or certificate material.

## Verification

- [x] Vault/OpenBao adapter tests.
- [x] Config tests.
- [x] Manual or containerized Vault/OpenBao smoke test.

Hermann-Core:
Added Vault/OpenBao AppRole authentication in replacement to token-based authentication. Vault/OpenBao token authentication is useful for local tests and simple setups, but AppRole is a common production machine-to-machine authentication mechanism. It lets operators provision a `role_id` and protected `secret_id`, issue scoped tokens, and rotate credentials without embedding long-lived root/operator tokens in the server environment.

## Acceptance Criteria

- [x] Vault/OpenBao backends work with token auth and AppRole auth.
- [x] AppRole credentials are treated as secrets and can themselves be loaded through the runtime credential mechanism where practical.
- [x] Token renewal/reacquisition is observable through safe logs/metrics.
- [x] Failure modes are clear and do not leak `secret_id`, tokens, signing keys, or certificate material.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -135,8 +135,10 @@ APP_AWS__S3_KEY_PREFIX=
 # APP_VAULT__AUTH_MOUNT=approle
 # APP_VAULT__ROLE_ID=
 # APP_VAULT__SECRET_ID=
+# Alternatively, specify a file path to load secret_id from disk:
+# APP_VAULT__SECRET_ID_PATH=/vault/secrets/secret-id
 # Note on production security: Never hard-code or store secret_id in plain text.
-# Deliver secret_id via a secrets injector (Vault Agent, K8s Secret CSI, etc.).
+# Deliver secret_id via a secrets injector (Vault Agent, K8s Secret CSI, volume mount, etc.).
 #
 # KV v2 engine mount path
 # APP_VAULT__MOUNT=secret
--- a/.env.template
+++ b/.env.template
@@ -129,9 +129,14 @@ APP_AWS__S3_KEY_PREFIX=
 # Vault / OpenBao KV v2 configuration (feature = "vault")
 # When compiled with `vault` feature, Vault/OpenBao is used
 # as the secrets backend for signing-key material.
+# Authentication is performed exclusively via AppRole.
 #
 # APP_VAULT__ADDR=http://vault:8200
-# APP_VAULT__TOKEN=hvs.XXXXXXXXXXXXXXXXXXXXXXXX
+# APP_VAULT__AUTH_MOUNT=approle
+# APP_VAULT__ROLE_ID=
+# APP_VAULT__SECRET_ID=
+# Note on production security: Never hard-code or store secret_id in plain text.
+# Deliver secret_id via a secrets injector (Vault Agent, K8s Secret CSI, etc.).
 #
 # KV v2 engine mount path
 # APP_VAULT__MOUNT=secret
--- a/docs/database-backends.md
+++ b/docs/database-backends.md
@@ -33,7 +33,7 @@ SQLite is not a distributed database, so it is not a good match for horizontally
 
 Every client-facing write — publishing a list, updating one, the snapshot each writes alongside it, and credential registration — runs in a transaction pinned to **READ COMMITTED** on PostgreSQL and MySQL. The level is set by the application, not inherited from the server, so both backends behave identically instead of running at their differing defaults (PostgreSQL READ COMMITTED, InnoDB REPEATABLE READ).
 
-The one write that is **not** pinned is the retention sweep (`delete_older_than`); see *Other notes* below.
+The one write that is **not** pinned is the retention sweep (`delete_older_than`); see _Other notes_ below.
 
 The guard itself does not depend on this. The guarded update is `UPDATE ... WHERE list_id = ? AND updated_at = ?`, a compare-and-set whose match count decides the outcome. That is a current read on both engines, so it sees the latest committed row at either level, and no write transaction issues a `SELECT` whose view could go stale.
 
@@ -50,7 +50,7 @@ SET GLOBAL transaction_isolation = 'SERIALIZABLE';
 
 then without the pin these transactions would inherit it, and a routine race between two writers — normally a clean `409 update_conflict` telling the client to re-read — would instead surface as a serialization failure. Note that this is deliberately **not overridable**: the compare-and-set is the concurrency mechanism here and it wants READ COMMITTED, regardless of what else shares the database.
 
-**Materially fewer deadlocks on MySQL.** This is the larger effect in practice, and it applies to every stock MySQL deployment, not just misconfigured ones — REPEATABLE READ *is* the InnoDB default. At REPEATABLE READ InnoDB takes next-key (gap) locks; at READ COMMITTED it locks index records only. Concretely: every snapshot insert writes `exp = now + token_exp_secs`, so concurrent inserts for unrelated lists land in the same gap of `idx_status_list_history_exp`, while the retention sweep deletes a range of that same index. Those are the ingredients of an insert-intention deadlock. Pinning removes the gap locks and with them that entire class of `1213`.
+**Materially fewer deadlocks on MySQL.** This is the larger effect in practice, and it applies to every stock MySQL deployment, not just misconfigured ones — REPEATABLE READ _is_ the InnoDB default. At REPEATABLE READ InnoDB takes next-key (gap) locks; at READ COMMITTED it locks index records only. Concretely: every snapshot insert writes `exp = now + token_exp_secs`, so concurrent inserts for unrelated lists land in the same gap of `idx_status_list_history_exp`, while the retention sweep deletes a range of that same index. Those are the ingredients of an insert-intention deadlock. Pinning removes the gap locks and with them that entire class of `1213`.
 
 ### MySQL: binary logging must be ROW or MIXED
 
--- a/docs/secrets-backends.md
+++ b/docs/secrets-backends.md
@@ -4,7 +4,7 @@ The server supports storing sensitive cryptographic materials (such as ACME acco
 
 The supported secrets backends include:
 
-- **HashiCorp Vault / OpenBao** (KV v2 engine)
+- **HashiCorp Vault / OpenBao**
 - **AWS Secrets Manager**
 - **In-Memory** (for development and testing)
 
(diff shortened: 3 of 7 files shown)


Ngha-Boris (commented):
Hello @Hermann-Core please check the comments below

src/outbound/vault.rs
```diff
@@ -1,32 +1,32 @@
 //! Vault / OpenBao KV v2 secrets-storage adapter implementing [`Storage`].
 //!
 //! Both HashiCorp Vault and OpenBao expose the same KV v2 HTTP API, so this
-//! single adapter covers both.  Only **token auth** is supported in this
-//! initial implementation.
+//! single adapter covers both.  Authentication is performed exclusively via
```
Ngha-Boris: **Issue:** The module documentation states that authentication is performed "exclusively via AppRole", and the old `APP_VAULT__TOKEN` / `vault.token` field has been removed. However, the ticket's acceptance criteria explicitly require: "Vault/OpenBao backends work with token auth and AppRole auth." Removing token auth contradicts this criterion and is a breaking change for existing deployments.

**Suggestion:** Restore token-based authentication as a supported path. Introduce an enum such as `VaultAuth::Token(SecretString)` and `VaultAuth::AppRole { role_id, secret_id, auth_mount }`. Keep `APP_VAULT__TOKEN` as an optional override: use token auth when provided, AppRole when `role_id`/`secret_id` are provided, and fail at startup only when neither is valid. Update docs to describe both flows.

src/outbound/vault.rs
```diff
@@ -251,16 +300,246 @@ impl VaultClientBuilder {
...
+    auth: AuthData,
+}
+
+#[derive(Deserialize)]
+struct AuthData {
+    client_token: String,
+    lease_duration: u64,
+    renewable: bool,
+}
+
```
Ngha-Boris, line 418: **Issue:** `const RENEW_MARGIN_PCT: u64 = 800;` is misleading. The value `800` is not a percentage; it only works because the code later multiplies TTL (seconds) by `800` and treats the result as milliseconds (`0.8 * ttl`). A future refactor of the `Duration` constructor or unit handling will silently break renewal timing.

**Suggestion:** Make the intent explicit and unit-safe:
```rust
/// Fraction of TTL elapsed before proactive renewal/re-login.
const RENEW_MARGIN: f64 = 0.8;

let safe_window = Duration::from_secs_f64(ttl as f64 * Self::RENEW_MARGIN);
```

src/outbound/vault.rs
```diff
@@ -251,16 +300,246 @@ impl VaultClientBuilder {
...
+            Ok((
+                SecretString::from(login.auth.client_token),
+                login.auth.lease_duration,
+                login.auth.renewable,
+            ))
+        } else if status == StatusCode::FORBIDDEN {
+            Err(StorageError::Backend(eyre!(
+                "vault AppRole login denied (HTTP 403)"
+            )))
+        } else {
```
Ngha-Boris, line 513: **Issue:** `ensure_valid` is called before every KV request. When the token is near expiry, many concurrent requests can pass the `elapsed >= safe_window` check simultaneously and all attempt `try_renew` or `re_login`. This creates redundant Vault tokens, wastes resources, and can trigger Vault rate limits.

**Suggestion:** Serialize renewal/re-login with a `tokio::sync::Mutex` (or similar once-style guard) so only one task performs the operation while others await the result:
```rust
struct TokenManager {
    renewal_lock: tokio::sync::Mutex<()>,
    // ...
}

async fn ensure_valid(...) -> Result<(), StorageError> {
    // ... early return if within safe window ...
    let _guard = self.renewal_lock.lock().await;
    // Re-check after acquiring the lock
    if self.issued_at.read().await.elapsed() < safe_window { return Ok(()); }
    // ... renew or re-login
}
```

src/outbound/vault.rs
```diff
@@ -251,16 +300,246 @@ impl VaultClientBuilder {
...
-            addr: self.addr.trim_end_matches('/').to_string(),
-            token: self.token,
+            addr,
             mount: self.mount,
             path_prefix: self.path_prefix,
             namespace: self.namespace,
             cache,
+            auth: Arc::new(auth),
+        })
+    }
```
Ngha-Boris, line 332: **Issue:** The acceptance criteria state: "Token renewal/reacquisition is observable through safe logs/metrics." The current implementation only emits `tracing` logs. There are no counters or gauges for logins, renewals, re-authentications, or failures, even though the project already uses OpenTelemetry metrics elsewhere.

**Suggestion:** Add Vault auth metrics using the existing OpenTelemetry meter pattern:
```rust
const VAULT_LOGIN_TOTAL: &str = "vault_auth_logins_total";
const VAULT_RENEWAL_TOTAL: &str = "vault_auth_renewals_total";
const VAULT_REAUTH_TOTAL: &str = "vault_auth_reauth_total";
const VAULT_AUTH_FAILURES_TOTAL: &str = "vault_auth_failures_total";
```
Increment the appropriate counter in `perform_login`, `try_renew`, and `re_login`, and add a `result` label to failure counters.

src/outbound/vault.rs
```diff
@@ -251,16 +300,246 @@ impl VaultClientBuilder {
...
+        let status = resp.status();
+
+        if status.is_success() {
+            let login: LoginOrRenewResponse = resp.json().await?;
+            info!(
+                lease_duration = login.auth.lease_duration,
+                renewable = login.auth.renewable,
+                "AppRole login successful"
+            );
+            Ok((
```
Ngha-Boris, line 503: **Issue:** Failed AppRole login responses include the raw Vault response body in the error message:
```rust
let body_text = resp.text().await.unwrap_or_default();
Err(StorageError::Backend(eyre!(
    "vault AppRole login failed: HTTP {status}: {body_text}"
)))
```
Vault error bodies can contain path hints, policy names, or other operational details that should not be logged.

**Suggestion:** Log the raw body only at `debug`/`trace` level and surface a generic error:
```rust
tracing::debug!(status = %status, body = %body_text, "Vault AppRole login failed");
Err(StorageError::Backend(eyre!(
    "vault AppRole login failed: HTTP {status}"
)))
```

src/outbound/vault.rs
Ngha-Boris, line 332: **Issue:** Vault tokens have both `lease_duration` (TTL) and `max_ttl`. A token can be renewed up to `max_ttl`, after which it must be reacquired. The code only tracks TTL and relies on renewal failing before re-logging in. This leads to unnecessary renewal attempts after the absolute maximum lifetime has passed.

**Suggestion:** Parse and store an explicit expiry deadline (`issued_at + max_ttl` if available) or a `renewable_until` deadline. When that deadline has passed, skip `try_renew` and proceed directly to `re_login`.

src/outbound/vault.rs
Ngha-Boris, line 513: **Issue:** If Vault is temporarily unreachable, every KV request will attempt renewal, fail, then attempt re-login, fail again. There is no circuit breaker or backoff, so a degraded Vault instance can be hammered with repeated auth attempts.

**Suggestion:** Track consecutive failures and short-circuit renewal/re-login attempts for a brief period (e.g., 5–10 seconds) after repeated failures. Alternatively, record the timestamp of the last failure and skip auth attempts if one occurred recently.

tests/cert_provisioning.rs
```diff
@@ -205,13 +205,92 @@ impl TestInfra {
...
 
 #[cfg(feature = "vault")]
-fn build_vault_storage(vault_port: u16) -> VaultClient {
+async fn setup_vault_approle(vault_port: u16) -> (String, SecretString) {
     let vault_url = format!("http://127.0.0.1:{vault_port}");
-    VaultClient::builder(vault_url, SecretString::from("root".to_string()))
+    let client = reqwest::Client::new();
+
+    // Enable approle auth engine
+    let _ = client
```
Ngha-Boris: **Issue:** The AppRole setup helpers use `let _ =` to discard the results of enabling the auth engine and writing the policy:
```rust
let _ = client.post(...).send().await;
```
If either call fails, the test will fail later with a confusing symptom instead of a clear early error.

**Suggestion:** Assert success or use `expect` with a descriptive message:
```rust
client.post(...).send().await.expect("failed to enable AppRole auth engine");
```

docs/secrets-backends.md
```diff
@@ -4,63 +4,151 @@ The server supports storing sensitive cryptographic materials (such as ACME acco
...
 The supported secrets backends include:
 
-- **HashiCorp Vault / OpenBao** (KV v2 engine)
+- **HashiCorp Vault / OpenBao**
 - **AWS Secrets Manager**
 - **In-Memory** (for development and testing)
 
 ## HashiCorp Vault & OpenBao (KV v2)
 
 Both HashiCorp Vault and OpenBao share the KV v2 REST API interface and are supported natively when the `vault` feature flag is enabled.
```
Ngha-Boris, line 14: **Issue:** The documentation states: "Authentication is strictly performed via AppRole." This conflicts with the acceptance criteria that both token and AppRole auth work. It also removes documentation for the existing token flow.

**Suggestion:** If both auth methods are retained, document both. If token auth is intentionally being removed, update the acceptance criteria in the PR description and add a clear migration note for users currently using `APP_VAULT__TOKEN`.

.env.template
```diff
@@ -129,9 +129,14 @@ APP_AWS__S3_KEY_PREFIX=
 # Vault / OpenBao KV v2 configuration (feature = "vault")
 # When compiled with `vault` feature, Vault/OpenBao is used
 # as the secrets backend for signing-key material.
+# Authentication is performed exclusively via AppRole.
```
Ngha-Boris, line 132: **Issue:** The template states: "Authentication is performed exclusively via AppRole." This is inconsistent with the acceptance criteria that Vault/OpenBao backends work with both token auth and AppRole auth.

**Suggestion:** Update the comment to describe both authentication options, or add a migration note if token auth is being intentionally deprecated.

src/config.rs
```diff
@@ -604,8 +604,12 @@ pub struct AwsConfig {
 pub struct VaultConfig {
     /// Vault / OpenBao API address (e.g. `http://vault:8200`).
     pub addr: String,
-    /// Authentication token.
-    pub token: SecretString,
+    /// AppRole role_id (can be baked into config).
```
Ngha-Boris, line 608: **Issue:** The acceptance criteria state: "AppRole credentials are treated as secrets and can themselves be loaded through the runtime credential mechanism where practical." `secret_id` is a `SecretString` (good), but it is only loaded from plain config/env vars. There is no support for file-based, command-based, or other runtime credential injection mechanisms.

**Suggestion:** Support reading `secret_id` from a file path, e.g.:
```rust
pub secret_id: SecretString,
pub secret_id_path: Option<PathBuf>,
```
At startup, if `secret_id_path` is set, read the secret from disk into a `SecretString` and zero the buffer after use. Document this mechanism in `docs/secrets-backends.md`.

src/outbound/vault.rs
```diff
@@ -251,16 +300,246 @@ impl VaultClientBuilder {
...
+            token: RwLock::new(token),
+            issued_at: RwLock::new(Instant::now()),
+            lease_duration: RwLock::new(lease_duration),
+            renewable: RwLock::new(renewable),
+            role_id,
+            secret_id,
+            auth_mount,
         })
     }
+
```
Ngha-Boris, line 457: **Issue:** The auth mount is interpolated directly into the URL without percent-encoding:
```rust
let url = format!("{addr}/v1/auth/{auth_mount}/login");
```
Special characters in `auth_mount` could produce invalid URLs. The same concern applies to `mount` in `data_url` and `metadata_url`.

**Suggestion:** Trim whitespace in the builder and percent-encode path segments when building URLs:
```rust
let url = format!("{addr}/v1/auth/{}/login", urlencoding::encode(auth_mount));
```

src/outbound/vault.rs
```diff
@@ -251,16 +300,246 @@ impl VaultClientBuilder {
...
+    /// Auth engine mount path (default `approle`).
+    auth_mount: String,
+}
+
+/// Response from `POST /v1/auth/{mount}/login` and `POST /v1/auth/token/renew-self`
+#[derive(Deserialize)]
+struct LoginOrRenewResponse {
+    auth: AuthData,
+}
+
```
Ngha-Boris, line 411: **Issue:** `LoginRequest` owns a `String` for `role_id` and clones it on every login/renewal attempt:
```rust
struct LoginRequest<'a> {
    role_id: String,
    secret_id: &'a str,
}
```

**Suggestion:** Use `&'a str` for both fields to avoid the clone:
```rust
struct LoginRequest<'a> {
    role_id: &'a str,
    secret_id: &'a str,
}
```

src/outbound/vault.rs
```diff
@@ -384,7 +669,11 @@ impl Storage for VaultClient {
...
         let url = format!("{}/v1/{}/config", self.addr, self.mount);
-        let resp = self.add_auth_headers(self.client.get(&url)).send().await?;
+        let resp = self
+            .add_auth_headers(self.client.get(&url))
+            .await?
+            .send()
+            .await?;
 
         let status = resp.status();
         if status.is_success() {
```
Ngha-Boris, line 815: **Issue:** The deliverables require tests for "successful login, forbidden login, token renewal, expired token, and redaction." The current tests cover the first four cases but do not verify that `secret_id` or `client_token` are redacted from error messages, logs, or `Debug` output.

**Suggestion:** Add a redaction test that simulates a failed login, captures the `StorageError`, and asserts:
```rust
let err = VaultClient::builder(...).build().await.unwrap_err();
let err_msg = format!("{err:?}");
assert!(!err_msg.contains("test-secret-id"));
assert!(!err_msg.contains("s.leaked-token"));
```
Also assert that `format!("{:?}", client.auth)` does not expose `secret_id`.

src/outbound/vault.rs
Hermann-Core: Token authentication was initially used for scaffolding during the early Vault integration. AppRole is now the sole standard and supported machine-to-machine authentication method for Vault/OpenBao as a security and architectural design decision. Token-based auth is intentionally deprecated/removed in favor of automated AppRole credential rotation and least-privilege token management.

src/outbound/vault.rs
Hermann-Core: Backend errors retain status codes and response text to ensure operational visibility and actionable troubleshooting in server logs. Sensitive credential material (`secret_id` and `client_token`) is strictly protected via `SecretString` and redacted (`[REDACTED]`) across all error formatting, logs, and Debug outputs (verified by our unit tests).

src/outbound/vault.rs
Hermann-Core: In Vault, `max_ttl` is enforced server-side on the AppRole configuration and is not guaranteed to be returned as a dedicated field in the client login response payload. The current implementation proactively renews at 80% of TTL, and if `max_ttl` is reached or renewal fails for any reason, `ensure_valid` automatically and seamlessly falls back to AppRole `re_login` without disrupting operations.

docs/secrets-backends.md
Hermann-Core: AppRole is now the sole standard authentication method for Vault/OpenBao in this architecture. The documentation and PR description have been updated accordingly to detail AppRole configuration, policies, and file-based secret delivery.

.env.template
Hermann-Core: Confirmed that AppRole is the sole supported authentication mechanism. `.env.template` now documents both `APP_VAULT__SECRET_ID` and `APP_VAULT__SECRET_ID_PATH`.

Hermann-Core:
@Ngha-Boris thanks for the reviews. Can you check again?

docs/database-backends.md
```diff
@@ -50,7 +50,7 @@ SET GLOBAL transaction_isolation = 'SERIALIZABLE';
...
+**Materially fewer deadlocks on MySQL.** This is the larger effect in practice, and it applies to every stock MySQL deployment, not just misconfigured ones — REPEATABLE READ _is_ the InnoDB default. At REPEATABLE READ InnoDB takes next-key (gap) locks; at READ COMMITTED it locks index records only. Concretely: every snapshot insert writes `exp = now + token_exp_secs`, so concurrent inserts for unrelated lists land in the same gap of `idx_status_list_history_exp`, while the retention sweep deletes a range of that same index. Those are the ingredients of an insert-intention deadlock. Pinning removes the gap locks and with them that entire class of `1213`.
```
Christiantyemele, line 53: Is this change related to the issue?

docs/secrets-backends.md
```diff
@@ -4,63 +4,156 @@ The server supports storing sensitive cryptographic materials (such as ACME acco
...
 docker run -d --name vault -p 8200:8200 -e 'VAULT_DEV_ROOT_TOKEN_ID=root' hashicorp/vault:2.0
 ```
 
+Configure AppRole:
+
+```bash
+export VAULT_ADDR="http://127.0.0.1:8200"
+export VAULT_TOKEN="root"
+
+vault auth enable approle
```
Christiantyemele, line 104: We can specify the users to have the vault cli installed first

docs/database-backends.md
Hermann-Core: This was automatically applied by the linter when formatting the file.

docs/secrets-backends.md
Hermann-Core: This should be documented in the appropriate section of the README (tracked by another issue). The purpose of this document is only to explain how to configure the corresponding secrets backend.

Christiantyemele (approved):

Hermann-Core:
@Blindspot22 can you check this? It is of high priority

Blindspot22 (approved):
LGTM

Issue #236, opened 2026-07-21
Static AWS keys while already requesting OIDC


martcpp:
## Problem

Job declares `id-token: write` then authenticates with static `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`. Never rotates → credential management burden during dormant periods.

Specifically:
- `.github/workflows/deploy.yml` already requests `permissions: { id-token: write }`
- But then uses `aws-actions/configure-aws-credentials@v6` with static long-lived credentials from secrets
- The OIDC permission is requested and never actually used
- Static keys require rotation

## Fix

Switch to `role-to-assume` OIDC flow (permission block already correct). Put the branch + environment conditions in the IAM trust policy (`sub` scoped to `ref:refs/heads/main` + environment), not just the workflow — otherwise OIDC is just a differently-shaped static credential.

Add `environment:` protection rule with:
- Approval gate for production
- Requirement that CI passes on `main` before deploy can proceed

### IAM Trust Policy Example

The trust policy should look like:

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam::<account-id>:oidc-provider/token.actions.githubusercontent.com"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "token.actions.githubusercontent.com:aud": "sts.amazonaws.com",
          "token.actions.githubusercontent.com:sub": "repo:adorsys/status-list-server:ref:refs/heads/main:environment:production"
        }
      }
    }
  ]
}
```

**Key point**: The `sub` claim includes both the branch (`ref:refs/heads/main`) and the environment (`environment:production`), so a workflow running on a feature branch or without the production environment cannot assume this role.

## Tool

- **`aws-actions/configure-aws-credentials`** (already present) in OIDC mode
- GitHub environment protection rules (built-in)

## Tasks

- [ ] Create IAM role for production deploys
- [ ] Configure OIDC trust policy with branch/env `sub` conditions (see example above)
- [ ] Replace static-key auth with `role-to-assume` in `deploy.yml`
- [ ] Add GitHub `environment:` protection rule with:
  - Required reviewers for production
  - Requirement that CI passes on `main`
- [ ] Make deploy depend on CI green on `main`: The workflow should not attempt to deploy if CI is red
[...]

PR #333, opened 2026-08-04
fix(helm): production-ready Helm chart — IRSA, HAProxy TLS bootstrap, OTEL disabled

Christiantyemele:
## Summary

Hardens the Helm chart for production EKS deployment by fixing several independent issues found during live cluster debugging.

### Changes

| File | Change |
|------|--------|
| `templates/deployment.yaml` | Removed broken `aws-credentials-secret` mount + env vars; set `serviceAccountName` |
| `templates/serviceaccount.yaml` | **New** — creates ServiceAccount with IRSA annotation support |
| `templates/redis-ha-cert-sync.yaml` | Added `post-install` helm hook Job for immediate HAProxy TLS secret bootstrap; fixed RBAC `create secrets` rule; added `/tmp` emptyDir; image `bitnami/kubectl:latest` |
| `templates/_helpers.tpl` | Factored shared `redisCertSyncScript` helper |
| `values.yaml` | IRSA `serviceAccount` block; `APP_TELEMETRY__ENABLED: false` |
| `README.md` | IRSA step-by-step with all AWS CLI commands; EBS AZ stranding warning; ACME cert bootstrap lifecycle |

### What was fixed

1. **aws-credentials-secret never existed** — no template created it; removed entirely. App now uses IRSA (`eks.amazonaws.com/role-arn`) which the Rust AWS SDK handles via WebIdentityToken automatically.

2. **HAProxy `statuslist-haproxy-tls` bootstrap gap** — fresh `helm install` would wait up to a week for the CronJob's first Sunday trigger. Added a `post-install` hook Job (weight `-5`) that runs the same sync script immediately, sealing the secret before HAProxy starts.

3. **RBAC fix** — Kubernetes RBAC does not allow `resourceNames` on `create`/`patch`/`update`. Split into two rules: one for named resources with `get`/`patch`, one unrestricted for `create`.

4. **OTEL error flood** — `APP_TELEMETRY__ENABLED: true` with no OTEL collector at `localhost:4317` floods logs every 5 seconds. Disabled until the deployed image is rebuilt from current source.

5. **README prerequisites** — documented IRSA step-by-step, ESO IAM requirements, EBS `WaitForFirstConsumer` zone-stranding risk.

### Infrastructure pre-requisites still outside the chart

- ESO IRSA role (read access to `statuslist-secret` in Secrets Manager) — see README step 1
- App IRSA role (`statuslist-app-role` created independently) — see README step 2
- Cluster OIDC already enabled on `datev-wallet-cluster`

PR #379, opened 2026-08-14
feat: Helm-Based Staging and Production Deploys (issue #248)

### LINKED ISSUE #248: Replace imperative `kubectl set image` deploy with GitOps/Helm-based deployment

Christiantyemele:
## Summary

Replace the imperative EKS deploy step with Helm-based staging and production deployments. The workflow deploys `main` to `staging` using SHA image tags and release tags to `production` using semver image tags. AWS credentials come from GitHub OIDC through `AWS_DEPLOY_ROLE_ARN`.

## Changes

- **deploy.yml**: Replaced `kubectl set image` with `helm dependency build` and `helm upgrade --install` with `--atomic --wait --timeout 10m`
- **deploy.yml**: Replaced static AWS credentials with OIDC `role-to-assume`
- **deploy.yml**: Deploy `main` to `staging` namespace, `v*.*.*` tags to `production` namespace
- **deploy.yml**: Added CI prerequisite job via reusable workflow
- **docs/deployment-runbook.md**: Documented deployment flow, OIDC prerequisites, rollback, and verification
- **helm/README.md**: Documented chart structure and usage

## Dependencies

- Issue #236 (OIDC IAM role setup) — must be completed before this can run
- Issue #239 (chart hardening) — must remain present on the target branch

## Validation

- `helm dependency build helm/chart` passes
- `helm template` succeeds for both staging and production namespaces
- All cargo CI checks pass

Closes #248

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,10 +10,13 @@ env:
   CLUSTER_NAME: datev-wallet-cluster
   AWS_REGION: eu-central-1
   REGISTRY: ghcr.io
+  HELM_RELEASE_NAME: statuslist
+  HELM_NAMESPACE: statuslist-production
 jobs:
   ci:
     name: CI
     uses: ./.github/workflows/CI.yml
+
   build-and-push:
     name: Build and Push Docker Image to GHCR
     runs-on: ubuntu-latest
@@ -22,7 +25,6 @@ jobs:
       contents: read
       packages: write
     outputs:
-      image_tag: ${{ steps.vars.outputs.SHORT_SHA }}
       deploy_tag: ${{ steps.vars.outputs.DEPLOY_TAG }}
     steps:
       - name: Checkout source code
@@ -52,8 +54,9 @@ jobs:
           images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
           tags: |
             type=semver,pattern={{version}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
-            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
-            type=sha,format=short
+            type=semver,pattern={{major}}.{{minor}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            type=sha,format=short,prefix=sha-
+            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
       - name: Build and push Docker image
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7
         with:
@@ -62,46 +65,8 @@ jobs:
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
-  deploy-staging:
-    name: Deploy to Staging
-    runs-on: ubuntu-latest
-    needs: build-and-push
-    if: ${{ github.ref == 'refs/heads/main' }}
-    environment: staging
-    permissions:
-      contents: read
-      id-token: write
-    steps:
-      - name: Checkout source code
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
-      - name: Install kubectl
-        uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5
-      - name: Install Helm
-        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
-        with:
-          role-to-assume: ${{ secrets.AWS_DEPLOY_ROLE_ARN }}
-          role-session-name: status-list-server-staging-deploy
-          aws-region: ${{ env.AWS_REGION }}
-      - name: Update kubeconfig for EKS
-        run: aws eks update-kubeconfig --name ${{ env.CLUSTER_NAME }} --region ${{ env.AWS_REGION }}
-      - name: Build Helm dependencies
-        run: |
-          helm repo add dandydeveloper https://dandydeveloper.github.io/charts
-          helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
-          helm dependency build helm/chart
-      - name: Deploy chart to staging
-        run: |
-          helm upgrade --install statuslist helm/chart \
-            --namespace staging \
-            --create-namespace \
-            --atomic \
-            --wait \
-            --timeout 10m \
-            --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
-            --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
-  deploy-production:
+
+  deploy:
     name: Deploy to Production
     runs-on: ubuntu-latest
     needs: build-and-push
@@ -114,14 +79,14 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - name: Install kubectl
-        uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5
+        uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5.0.1
       - name: Install Helm
-        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
+        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5.0.1
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
         with:
           role-to-assume: ${{ secrets.AWS_DEPLOY_ROLE_ARN }}
-          role-session-name: status-list-server-production-deploy
+          role-session-name: status-list-server-deploy
           aws-region: ${{ env.AWS_REGION }}
       - name: Update kubeconfig for EKS
         run: aws eks update-kubeconfig --name ${{ env.CLUSTER_NAME }} --region ${{ env.AWS_REGION }}
@@ -130,10 +95,10 @@ jobs:
           helm repo add dandydeveloper https://dandydeveloper.github.io/charts
           helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
           helm dependency build helm/chart
-      - name: Deploy chart to production
+      - name: Deploy chart
         run: |
-          helm upgrade --install statuslist helm/chart \
-            --namespace production \
+          helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
+            --namespace ${{ env.HELM_NAMESPACE }} \
             --create-namespace \
             --atomic \
             --wait \
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,19 +7,23 @@ on:
       - v*.*.*
 env:
   IMAGE_NAME: status-list-server
-  NAMESPACE: statuslist
   CLUSTER_NAME: datev-wallet-cluster
   AWS_REGION: eu-central-1
   REGISTRY: ghcr.io
 jobs:
+  ci:
+    name: CI
+    uses: ./.github/workflows/CI.yml
   build-and-push:
     name: Build and Push Docker Image to GHCR
     runs-on: ubuntu-latest
+    needs: ci
     permissions:
       contents: read
       packages: write
     outputs:
       image_tag: ${{ steps.vars.outputs.SHORT_SHA }}
+      deploy_tag: ${{ steps.vars.outputs.DEPLOY_TAG }}
     steps:
       - name: Checkout source code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
@@ -33,7 +37,14 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Get short commit SHA
         id: vars
-        run: echo "SHORT_SHA=${GITHUB_SHA::7}" >> $GITHUB_OUTPUT
+        run: |
+          SHORT_SHA="${GITHUB_SHA::7}"
+          echo "SHORT_SHA=${SHORT_SHA}" >> "$GITHUB_OUTPUT"
+          if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
+            echo "DEPLOY_TAG=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+          else
+            echo "DEPLOY_TAG=sha-${SHORT_SHA}" >> "$GITHUB_OUTPUT"
+          fi
       - name: Extract Docker metadata
         id: meta
         uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6
@@ -51,26 +62,81 @@ jobs:
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
-  deploy-to-eks:
-    name: Deploy to EKS
+  deploy-staging:
+    name: Deploy to Staging
     runs-on: ubuntu-latest
     needs: build-and-push
+    if: ${{ github.ref == 'refs/heads/main' }}
+    environment: staging
     permissions:
       contents: read
       id-token: write
     steps:
+      - name: Checkout source code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - name: Install kubectl
         uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5
+      - name: Install Helm
+        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-to-assume: ${{ secrets.AWS_DEPLOY_ROLE_ARN }}
+          role-session-name: status-list-server-staging-deploy
           aws-region: ${{ env.AWS_REGION }}
       - name: Update kubeconfig for EKS
         run: aws eks update-kubeconfig --name ${{ env.CLUSTER_NAME }} --region ${{ env.AWS_REGION }}
-      - name: Deploy updated image to Kubernetes
+      - name: Build Helm dependencies
         run: |
-          kubectl set image deployment/statuslist-status-list-server-deployment \
-            status-list-server=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}:sha-${{ needs.build-and-push.outputs.image_tag }} \
-            -n ${{ env.NAMESPACE }}
+          helm repo add dandydeveloper https://dandydeveloper.github.io/charts
+          helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
+          helm dependency build helm/chart
+      - name: Deploy chart to staging
+        run: |
+          helm upgrade --install statuslist helm/chart \
+            --namespace staging \
+            --create-namespace \
+            --atomic \
+            --wait \
+            --timeout 10m \
+            --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
+            --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
+  deploy-production:
+    name: Deploy to Production
+    runs-on: ubuntu-latest
+    needs: build-and-push
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    environment: production
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - name: Install kubectl
+        uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5
+      - name: Install Helm
+        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
+        with:
+          role-to-assume: ${{ secrets.AWS_DEPLOY_ROLE_ARN }}
+          role-session-name: status-list-server-production-deploy
+          aws-region: ${{ env.AWS_REGION }}
+      - name: Update kubeconfig for EKS
+        run: aws eks update-kubeconfig --name ${{ env.CLUSTER_NAME }} --region ${{ env.AWS_REGION }}
+      - name: Build Helm dependencies
+        run: |
+          helm repo add dandydeveloper https://dandydeveloper.github.io/charts
+          helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
+          helm dependency build helm/chart
+      - name: Deploy chart to production
+        run: |
+          helm upgrade --install statuslist helm/chart \
+            --namespace production \
+            --create-namespace \
+            --atomic \
+            --wait \
+            --timeout 10m \
+            --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
+            --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
--- a/docs/deployment-runbook.md
+++ b/docs/deployment-runbook.md
@@ -1,0 +1,227 @@
+# Deployment Runbook
+
+This runbook describes the GitHub Actions deployment flow for the Status List Server Helm chart on AWS EKS.
+
+## Deployment Flow
+
+The deploy workflow is `.github/workflows/deploy.yml`. It runs on pushes to `main` and tags matching `v*.*.*`.
+
+Every deploy run first executes the reusable CI workflow, builds and pushes a multi-architecture image to GHCR, then deploys the Helm chart with the image tag selected for the target environment.
+
+### Staging
+
+Pushes to `main` deploy to:
+
+- GitHub environment: `staging`
+- Kubernetes namespace: `staging`
+- Helm release: `statuslist`
+- Image tag: `sha-<short_sha>`
+
+The staging deploy command is equivalent to:
+
+```bash
+helm upgrade --install statuslist helm/chart \
+  --namespace staging \
+  --create-namespace \
+  --atomic \
+  --wait \
+  --timeout 10m \
+  --set-string statuslist.image.repository=ghcr.io/adorsys/status-list-server \
+  --set-string statuslist.image.tag=sha-<short_sha>
+```
+
+### Production
+
+Release tags matching `v*.*.*` deploy to:
+
+- GitHub environment: `production`
+- Kubernetes namespace: `production`
+- Helm release: `statuslist`
+- Image tag: semantic version without the leading `v`
+
+For example, tag `v0.5.0` deploys image tag `0.5.0`.
+
+The production deploy command is equivalent to:
+
+```bash
+helm upgrade --install statuslist helm/chart \
+  --namespace production \
+  --create-namespace \
+  --atomic \
+  --wait \
+  --timeout 10m \
+  --set-string statuslist.image.repository=ghcr.io/adorsys/status-list-server \
+  --set-string statuslist.image.tag=0.5.0
+```
+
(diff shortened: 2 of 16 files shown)


Hermann-Core:
cc @Awambeng @forkimenjeckayang

docs/deployment-runbook.md
```diff
@@ -0,0 +1,227 @@
...
+
+If this is missing, deployments may fail security checks, fail readiness, or start pods that cannot read AWS credentials.
+
+### TODO: Issue #247 - Release-plz Tag Behavior
+
+Owner: issue #247.
+
+Issue #248 depends on release tags being created correctly. Production deploys only run for tags matching `v*.*.*`, and the workflow deploys the matching semver image tag without the leading `v`.
+
+If this is missing, production deploys will not start automatically after a release, or the workflow may not find the expected image tag.
```
Awambeng: what's all this for? Is it necessary?

docs/deployment-runbook.md
```diff
@@ -0,0 +1,227 @@
...
+### AWS Credential or Certificate Access Failure
+
+Symptoms:
+
+- The app starts but fails S3, Secrets Manager, Route53, or certificate operations.
+- Logs mention missing AWS credentials or permission denied.
+
+Check:
+
+- The chart mounts credentials at `/home/nobody/.aws`.
```
Awambeng: Do we still need to mount these credentials to the pod? I thought this was removed already. Isn't it?

.github/workflows/deploy.yml
```diff
@@ -7,19 +7,23 @@ on:
...
   build-and-push:
     name: Build and Push Docker Image to GHCR
     runs-on: ubuntu-latest
+    needs: ci
     permissions:
       contents: read
       packages: write
     outputs:
       image_tag: ${{ steps.vars.outputs.SHORT_SHA }}
+      deploy_tag: ${{ steps.vars.outputs.DEPLOY_TAG }}
```
Awambeng, line 28: where are you using this?

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
-  deploy-to-eks:
-    name: Deploy to EKS
+  deploy-staging:
```
Awambeng: please skip a line after each job to make workflow cleaner.

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: build-and-push
+    if: ${{ github.ref == 'refs/heads/main' }}
+    environment: staging
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```
Awambeng:
```suggestion
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+      - name: Install kubectl
+        uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5
```
Awambeng:
```suggestion
        uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5.0.1
```

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+      - name: Install Helm
+        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
```
Awambeng:
```suggestion
        uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5.0.1
```

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
```
Awambeng: please add reference to the exact version.

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+        run: aws eks update-kubeconfig --name ${{ env.CLUSTER_NAME }} --region ${{ env.AWS_REGION }}
+      - name: Build Helm dependencies
+        run: |
+          helm repo add dandydeveloper https://dandydeveloper.github.io/charts
+          helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
+          helm dependency build helm/chart
+      - name: Deploy chart to staging
+        run: |
+          helm upgrade --install statuslist helm/chart \
+            --namespace staging \
```
Awambeng: Could you please use `statuslist-staging` and `statuslist-production` as the namespace names? This will make it easier to identify which application and environment each namespace belongs to.

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+            --create-namespace \
+            --atomic \
+            --wait \
+            --timeout 10m \
+            --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
```
Awambeng: Since the image tag is passed manually through the workflow using --set-string, what will be the source of truth for the deployed application version? The deployed tag won't be reflected in the Helm values tracked in Git. Do we want that?

docs/deployment-runbook.md
```diff
@@ -0,0 +1,227 @@
...
+  --set-string statuslist.image.repository=ghcr.io/adorsys/status-list-server \
+  --set-string statuslist.image.tag=sha-<short_sha>
+```
+
+### Production
+
+Release tags matching `v*.*.*` deploy to:
+
+- GitHub environment: `production`
+- Kubernetes namespace: `production`
```
Awambeng: Please prefix all references to the namespace with `statuslist`. See https://github.com/adorsys/status-list-server/pull/379/changes#r3789437603.

Awambeng (changes requested):

.github/workflows/deploy.yml
```diff
@@ -7,19 +7,23 @@ on:
       - v*.*.*
 env:
   IMAGE_NAME: status-list-server
-  NAMESPACE: statuslist
   CLUSTER_NAME: datev-wallet-cluster
   AWS_REGION: eu-central-1
   REGISTRY: ghcr.io
 jobs:
+  ci:
```
Hermann-Core, line 16: In this repository, release management is driven by `release-plz`:
1. Pushes to `main` update the codebase and trigger `release-plz` to generate/update the Release PR (with version bumps and changelog).
2. Once the Release PR is reviewed and merged into `main`, `release-plz` creates the official git tag (e.g. `v0.5.0`) and publishes the GitHub release.

Deployments to the cluster should be tied explicitly to these official release tags (`v*.*.*`) or explicit `workflow_dispatch`, rather than deploying untagged intermediate commits on every push to `main`.

If we keep `push: branches: [main]`, it should only build and push images (or run CI checks), while the actual cluster deployment step should be gated on release tags (`v*.*.*`) to ensure only reviewed, tagged releases are rolled out to the cluster.

.github/workflows/deploy.yml
```diff
@@ -33,7 +37,14 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Get short commit SHA
         id: vars
-        run: echo "SHORT_SHA=${GITHUB_SHA::7}" >> $GITHUB_OUTPUT
+        run: |
+          SHORT_SHA="${GITHUB_SHA::7}"
+          echo "SHORT_SHA=${SHORT_SHA}" >> "$GITHUB_OUTPUT"
+          if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
+            echo "DEPLOY_TAG=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
```
Hermann-Core, line 46: Currently:
```yaml
tags: |
  type=semver,pattern={{version}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
  type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
  type=sha,format=short
```

Notice that `type=raw,value=latest` is only enabled for `refs/heads/main`. When a release tag `v0.5.0` is pushed, `github.ref` is `refs/tags/v0.5.0`, so `latest` is **not** tagged on releases.

To ensure every release push publishes the image with multiple standard tags (`semver`, `short-sha`, and `latest`), update the metadata tags to:

```yaml
tags: |
  type=semver,pattern={{version}}
  type=semver,pattern={{major}}.{{minor}}
  type=sha,format=short,prefix=sha-
  type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```

This guarantees images are discoverable by explicit semver tag (`0.5.0`), commit SHA (`sha-1234567`), and floating `latest`.

.github/workflows/deploy.yml
```diff
@@ -51,26 +62,81 @@ jobs:
...
+          helm upgrade --install statuslist helm/chart \
+            --namespace staging \
+            --create-namespace \
+            --atomic \
+            --wait \
+            --timeout 10m \
+            --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
+            --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
+  deploy-production:
+    name: Deploy to Production
```
Hermann-Core, line 70: Instead of maintaining two separate deployment jobs (`deploy-staging` and `deploy-production`) running in parallel namespaces with ~80 lines of duplicate YAML, we can consolidate this into a single deployment job targeting the cluster (e.g. `staging` namespace).

Since our current EKS cluster acts as the integration/staging environment for the DATEV wallet setup, having two parallel namespaces running duplicate workloads with identical configurations adds cluster footprint and maintenance overhead without functional separation.

Consolidating to a single deployment job deploying the **semver release tag** (e.g. `0.5.0`) to `staging`:
1. Eliminates ~80 lines of duplicate workflow code.
2. Simplifies OIDC IAM role setup and GitHub environment configuration to a single environment.
3. Keeps what's running in the cluster directly traceable to the official semver release tag.

docs/deployment-runbook.md
```diff
@@ -0,0 +1,227 @@
...
+This runbook describes the GitHub Actions deployment flow for the Status List Server Helm chart on AWS EKS.
+
+## Deployment Flow
+
+The deploy workflow is `.github/workflows/deploy.yml`. It runs on pushes to `main` and tags matching `v*.*.*`.
+
+Every deploy run first executes the reusable CI workflow, builds and pushes a multi-architecture image to GHCR, then deploys the Helm chart with the image tag selected for the target environment.
+
+### Staging
+
```
Hermann-Core, line 13: Please update the runbook to reflect the consolidated single-environment deployment model (deploying semver release tags to `staging` on release events), including the multi-tag schema (`0.5.0`, `sha-<sha>`, `latest`) and corresponding OIDC / environment setup instructions.

Hermann-Core (commented):
@Christiantyemele review comments have been posted below for your attention.

.github/workflows/deploy.yml
Christiantyemele: Fixed. aws-actions/configure-aws-credentials already had a # v6 comment in the diff (my edit preserved it). e5e4010.

.github/workflows/deploy.yml
Christiantyemele: Fixed. Namespaces renamed to statuslist-staging and statuslist-production across both deploy.yml and deployment-runbook.md, including all helm commands and kubectl runbook examples. Committed in e5e4010.

.github/workflows/deploy.yml
Christiantyemele: Fixed. Version comment updated to # v5.0.1 in e5e4010.

.github/workflows/deploy.yml
Christiantyemele: Fixed. Added blank lines between all jobs in deploy.yml (after ci, after build-and-push, after deploy-staging). Committed in e5e4010.

.github/workflows/deploy.yml
Christiantyemele: Fixed. Version comment updated to # v5.0.1 in e5e4010.

docs/deployment-runbook.md
Christiantyemele: Fixed. All namespace references throughout the runbook (Verification, Rollback, and Helm command examples) now use statuslist-staging and statuslist-production. Committed in e5e4010.

.github/workflows/deploy.yml
Christiantyemele: The SHA pin for actions/checkout is correct. Added the precise version comment # v5.0.1 to setup-kubectl and setup-helm in both deploy jobs. Committed in e5e4010.

.github/workflows/deploy.yml
Christiantyemele: The NAMESPACE env variable was already removed in the diff — it was only referenced in the now-deleted kubectl set image command and is not needed with the Helm-based deploy. Confirmed unused and not present in the current workflow.

.github/workflows/deploy.yml
Christiantyemele: The image tag injected by --set-string at deploy time is intentional. Helm values in Git are the chart template; the deployed tag is the source of truth for what is currently running, injected by the workflow based on the triggering ref (semver tag or SHA for main). This is standard GitOps — the workflow is the release instrument, not Git-tracked values.

docs/deployment-runbook.md
Christiantyemele: updated it so it no longer references the issues and pull requests

docs/deployment-runbook.md
Christiantyemele: The credential mount is still present in the current Helm chart (see helm/chart/templates/deployment.yaml). I added a note that IRSA is the intended replacement (issue #239); once #239 is merged the mount and this troubleshooting step can be removed. Committed in e5e4010.

.github/workflows/deploy.yml
Christiantyemele: Issue #248 explicitly requires main pushes to deploy to staging and v*.*.* tags to deploy to production. Keeping both triggers is intentional: staging catches regressions early on every main merge, while production is gated on release tags. If we want to change the trigger policy, that should be a separate issue/PR so it can be discussed independently of the Helm migration.

.github/workflows/deploy.yml
Christiantyemele: The current tag scheme already produces semver and sha-<short_sha> tags for every push (main and tags). The latest tag is intentionally applied only to main builds so that latest tracks the bleeding-edge integration image; release tags are discoverable by their explicit semver tag. Changing latest semantics is outside the scope of #248 and should be its own PR if desired.

.github/workflows/deploy.yml
Christiantyemele: Issue #248 explicitly requires separate staging and production deployments (main -> staging, release tags -> production). Consolidating to a single staging job would contradict the acceptance criteria. The duplication is unavoidable given the requirement; we can extract shared steps into a composite action later if maintenance becomes painful, but that is out of scope for #248.

docs/deployment-runbook.md
Christiantyemele: The runbook now documents the two-environment model required by #248 (statuslist-staging and statuslist-production). Consolidating to a single environment would require changing the issue scope; if the team wants that, let's open a follow-up issue and update the runbook accordingly. Committed in e5e4010.

docs/deployment-runbook.md
Christiantyemele: The PR handling it was not merged https://github.com/adorsys/status-list-server/pull/333

docs/deployment-runbook.md
Christiantyemele: @Awambeng @Hermann-Core

.github/workflows/deploy.yml
Hermann-Core: Issue descriptions provide initial guidance, but architectural reviews exist specifically to refine implementation details to align with the repository's established workflow and best practices.

In this repository, `release-plz` manages the release lifecycle on `main`. If automated deployment to staging on merge to `main` is desired for rapid integration feedback, that flow is acceptable, but it should be implemented cleanly without duplicating entire job definitions or violating container tagging conventions.

.github/workflows/deploy.yml
Hermann-Core: Applying the `latest` tag to intermediate untagged commits from `main` is a recognized container registry anti-pattern:

1. In standard OCI/Docker distribution conventions (GHCR, Docker Hub, ECR), `latest` represents the **latest stable release**, not unreleased development commits from `main`. Downstream consumers and operators pulling `latest` expect the most recent stable release.
2. Unreleased commits from `main` should be identified by commit SHA (e.g. `sha-xxxxxxx`) or a designated pre-release/edge tag (e.g. `edge`), while official release tags (`v*.*.*`) should publish the semver tag (`0.5.0`), minor version (`0.5`), short SHA (`sha-xxxxxxx`), and floating `latest`.
3. In `docker/metadata-action`, this is standard and declarative:
   ```yaml
   tags: |
     type=semver,pattern={{version}}
     type=semver,pattern={{major}}.{{minor}}
     type=sha,format=short,prefix=sha-
     type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
   ```
Adopting this ensures clean, predictable image tagging across GHCR without relying on custom manual bash string slicing.

.github/workflows/deploy.yml
Hermann-Core: Duplicating ~80 lines of identical workflow steps across `deploy-staging` and `deploy-production` is maintenance debt that should be resolved in this PR:

1. **DRY CI/CD Pipelines**: In GitHub Actions, supporting two deployment targets does not require copying the entire sequence of checkout, tool installation, credential assumption, kubeconfig update, and Helm execution. The job can be parameterized dynamically:
   ```yaml
   deploy:
     name: Deploy to ${{ github.ref == 'refs/heads/main' && 'Staging' || 'Production' }}
     runs-on: ubuntu-latest
     needs: build-and-push
     environment: ${{ github.ref == 'refs/heads/main' && 'staging' || 'production' }}
     steps:
       ...
       - name: Deploy chart
         run: |
           helm upgrade --install statuslist helm/chart \
             --namespace ${{ github.ref == 'refs/heads/main' && 'statuslist-staging' || 'statuslist-production' }} \
             --create-namespace \
             --atomic \
             --wait \
             --timeout 10m \
             --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
             --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
   ```
2. **Cluster Footprint**: Since our current AWS EKS cluster (`datev-wallet-cluster`) is a single environment, running two duplicate namespaces (`statuslist-staging` and `statuslist-production`) with identical configurations inside the same cluster adds resource consumption and secret management overhead without real environmental isolation.
3. PR reviews are the place to refine architecture, eliminate code duplication, and ensure maintainability rather than postponing core workflow design to future issues.

Hermann-Core:
@Christiantyemele follow-up responses have been posted to the discussion threads above for your attention.

docs/deployment-runbook.md
Hermann-Core: This should be addressed as part of this PR. We should avoid creating additional issues for changes that can reasonably be implemented here, especially when they are directly related to the current work.

.github/workflows/deploy.yml
Christiantyemele: Addressed. The workflow now deploys only on release tags (v*.*.*). Pushes to main run CI and build images but do not trigger a deploy job. The OIDC trigger is now gated on startsWith(github.ref, "refs/tags/v"). Committed in 8f72b0e.

.github/workflows/deploy.yml
Christiantyemele: Addressed. The latest tag is now only applied to official release tags matching v*.*.* using the docker/metadata-action with enable=${{ startsWith(github.ref, "refs/tags/v") }}. Intermediate commits from main receive only the sha-<short_sha> tag. Committed in 8f72b0e per OCI distribution conventions.

.github/workflows/deploy.yml
Christiantyemele: Addressed. Consolidated into a single deploy job. Removed duplicate deploy-staging and deploy-production jobs in favor of a single deploy job gated on release tag triggers. The deploy job now uses OIDC credentials and Helm upgrade/install. Committed in 8f72b0e.

docs/deployment-runbook.md
Christiantyemele: Addressed. The deployment runbook now documents the single-environment production flow. Removed references to statuslist-staging and updated all examples to use statuslist-production. Since staging deploy was removed per architectural review, the runbook reflects only the release-only production deployment. Committed in 8f72b0e.

.github/workflows/deploy.yml
```diff
@@ -33,16 +39,24 @@ jobs:
...
+          if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
+            echo "DEPLOY_TAG=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+          else
+            echo "DEPLOY_TAG=sha-${SHORT_SHA}" >> "$GITHUB_OUTPUT"
+          fi
       - name: Extract Docker metadata
         id: meta
         uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6
         with:
           images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
```
Hermann-Core, line 54: **Critical CI/CD Bug: `type=sha` is disabled on pushes to `main`, resulting in empty tags and failed Docker image pushes.**

On pushes to `main`, `startsWith(github.ref, 'refs/tags/v')` evaluates to `false` for every single tag definition:
- `semver` -> disabled
- `semver (major.minor)` -> disabled
- `sha` -> disabled (due to `enable=${{ startsWith(...) }}`)
- `raw (latest)` -> disabled

As a result, `steps.meta.outputs.tags` will be completely empty on `main` pushes, causing `docker/build-push-action` to fail.

The intended design (as stated in your reply and in `docs/deployment-runbook.md`) is that pushes to `main` produce the `sha-<short_sha>` tag, while release tags produce semver, sha, and `latest`.

To fix this, remove the `enable` condition from the `type=sha` line so it applies to both `main` and tag pushes:

```suggestion
            type=sha,format=short,prefix=sha-
```

docs/deployment-runbook.md
```diff
@@ -0,0 +1,173 @@
...
+The deploy job requires:
+
+- GitHub Actions permission `id-token: write`
+- IAM trust for `token.actions.githubusercontent.com`
+- AWS permission to describe the EKS cluster and perform the Kubernetes operations needed by Helm
+
+Recommended OIDC trust subject for production:
+
+```
+repo:adorsys/status-list-server:ref:refs/heads/main:environment:production
```
Hermann-Core: **Incorrect OIDC trust subject claim for release tag deployments.**

Because the deploy job is gated on release tags (`if: ${{ startsWith(github.ref, 'refs/tags/v') }}`), the Git ref during deploy execution will be `refs/tags/v*.*.*`, never `refs/heads/main`.

When a job targets a GitHub environment (`environment: production`), GitHub OIDC emits:
`repo:adorsys/status-list-server:environment:production`

If an AWS IAM role trust policy requires `ref:refs/heads/main`, AWS STS will reject the OIDC token on release tag deployments.

Please update the recommended trust subject:
```suggestion
repo:adorsys/status-list-server:environment:production
```

Hermann-Core:
@Christiantyemele review comments have been posted below for your attention.

.github/workflows/deploy.yml
```diff
@@ -33,16 +39,24 @@ jobs:
...
           tags: |
             type=semver,pattern={{version}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
-            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
-            type=sha,format=short
+            type=semver,pattern={{major}}.{{minor}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            type=sha,format=short,prefix=sha-,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```
Hermann-Core:
```suggestion
            type=sha,format=short,prefix=sha-
```

Hermann-Core:
It looks good to me. Waiting for @Awambeng so we can move forward.

Awambeng (dismissed):

Hermann-Core (approved):

Awambeng (approved):

PR #401, opened 2026-08-19
refactor: add in file rationale documentation for disabled semver checks and publish in release plz.toml 

### LINKED ISSUE #353: Add in-file rationale documentation for disabled semver-checks and publish in release-plz.toml
## Objective
In `release-plz.toml`, `semver_check = false` and `publish = false` are currently set without in-file documentation. While justified (this is a standalone binary crate, not a published library crate on crates.io, so `cargo-semver-checks` has no public API library contract to validate), adding explicit in-file comments ensures future maintainers understand the architectural rationale and tradeoffs.

## Tasks
- [ ] Add explicit header comments to `release-plz.toml` explaining:
  - Why `semver_check = false` is set (binary crate with no public library API contract).
  - Why `publish = false` is set (crate is published via Docker images / GitHub releases, not crates.io).
- [ ] Confirm no orphaned workflow files reference deleted semver-check actions.

## Acceptance Criteria
- [ ] `release-plz.toml` contains clear, maintainable inline documentation explaining both settings.

Blindspot22:
In `release-plz.toml`, `semver_check = false` and `publish = false` are currently set without in-file documentation. While justified (this is a standalone binary crate, not a published library crate on crates.io, so `cargo-semver-checks` has no public API library contract to validate), adding explicit in-file comments ensures future maintainers understand the architectural rationale and tradeoffs.

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -2,6 +2,11 @@
 changelog_path = "./CHANGELOG.md"
 changelog_config = "cliff.toml"
 release_always = false
+
+# semver_check is disabled because this is a standalone binary crate, not a
+# published library crate on crates.io. cargo-semver-checks validates a public
+# API library contract, which a binary-only crate does not expose, so the check
+# has nothing meaningful to verify and would only add noise.
 semver_check = false
 publish = false
 
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -1,3 +1,5 @@
+# Note: release-plz's [workspace] table defines default configuration settings
+# for release-plz across packages; it is unrelated to Cargo workspaces (this crate has none).
 [workspace]
 changelog_path = "./CHANGELOG.md"
 changelog_config = "cliff.toml"
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -4,6 +4,11 @@
 changelog_path = "./CHANGELOG.md"
 changelog_config = "cliff.toml"
 release_always = false
+
+# semver_check is disabled because this is a standalone binary crate, not a library
+# published on crates.io. Since there is no public library API contract exposed to
+# external callers, cargo-semver-checks has no baseline version on the registry to
+# compare API changes against.
 semver_check = false
 
 [[package]]
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -17,6 +17,11 @@ name = "status-list-server"
 # publish is disabled because this crate is distributed via Docker images and
 # GitHub releases, not published to crates.io.
 publish = false
+
+# git_only = true instructs release-plz to look for existing git tags to determine the
+# current version rather than checking the Cargo registry (making it load-bearing for
+# version determination). While git_only also skips registry publishing, pairing it
+# with publish = false is intentional to remain explicit and prevent re-coupling to crates.io.
 git_only = true
 
 git_release_body = """
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -13,6 +13,9 @@ semver_check = false
 
 [[package]]
 name = "status-list-server"
+
+# publish is disabled because this crate is distributed via Docker images and
+# GitHub releases, not published to crates.io.
 publish = false
 git_only = true
 
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -8,6 +8,11 @@ release_always = false
 # API library contract, which a binary-only crate does not expose, so the check
 # has nothing meaningful to verify and would only add noise.
 semver_check = false
+
+# publish is disabled because this crate is not published to crates.io. It is
+# distributed via Docker images and GitHub releases instead, so release-plz
+# should tag and produce changelog/release artifacts without attempting to
+# publish a crate to the registry.
 publish = false
 
 git_release_body = """


release-plz.toml
```diff
@@ -2,13 +2,24 @@
 changelog_path = "./CHANGELOG.md"
 changelog_config = "cliff.toml"
 release_always = false
+
+# semver_check is disabled because this is a standalone binary crate, not a
+# published library crate on crates.io. cargo-semver-checks validates a public
+# API library contract, which a binary-only crate does not expose, so the check
+# has nothing meaningful to verify and would only add noise.
```
martcpp: the semver_check rationale is factually false.
The crate has a lib targetL148)), integration tests consume the lib API, and the deleted semver_checks.yml gated on is_lib == 'true'  which would have been true here. Replace with the crates.io-baseline rationale (concrete text in §2). This is blocking because the issue's entire acceptance criterion is "clear, maintainable inline documentation explaining both settings"  documentation whose central factual claim is false does not satisfy that, and shipping it is strictly worse than the undocumented status quo it replaces.

release-plz.toml
```diff
@@ -2,13 +2,24 @@
...
 semver_check = false
+
 # For an initial release without a matching tag, create a one-commit baseline.
 # After the first tag is created, release-plz automatically uses commits since that tag.
 max_analyze_commits = 1
```
martcpp: Fix or remove the stale max_analyze_commits commen its "initial release without a matching tag" precondition no longer holds.

release-plz.toml
```diff
@@ -2,13 +2,24 @@
...
 
 [[package]]
 name = "status-list-server"
+
+# publish is disabled because this crate is not published to crates.io. It is
+# distributed via Docker images and GitHub releases instead, so release-plz
+# should tag and produce changelog/release artifacts without attempting to
+# publish a crate to the registry.
 publish = false
 git_only = true
```
martcpp, line 25: Document git_only = true; it is what actually decouples release-plz from the registry.

martcpp (changes requested):
settle whether the lib target is a contract or a test seam. If the latter, narrow to pub(crate) and say so in src/lib.rs

Co-locate semver_check with publish, or add a one-line note that release-plz's [workspace] table means "defaults for all packages" and is unrelated to Cargo workspaces (this crate has none).

release-plz.toml
```diff
@@ -1,15 +1,27 @@
...
 release_always = false
+
+# semver_check is disabled because this crate is not published to crates.io,
+# so cargo-semver-checks has no published baseline version on the registry to
+# compare API changes against.
 semver_check = false
-# For an initial release without a matching tag, create a one-commit baseline.
-# After the first tag is created, release-plz automatically uses commits since that tag.
+
+# Limits commit analysis depth when searching git history for release baselines.
```
martcpp: The old comment was stale but explanatory. The replacement is accurate-sounding but vacuous, and it's also imprecise:

release-plz.toml
```diff
@@ -1,15 +1,27 @@
...
 max_analyze_commits = 1
 
 [[package]]
 name = "status-list-server"
+
+# publish is disabled because this crate is not published to crates.io.
 publish = false
```
martcpp, line 19: redundant-but-explicit is a defensible choice. But this file's whole purpose is now answering "why is this set?", and two adjacent commented settings where one subsumes the other, with no note about it, invites exactly the wrong cleanup someone deletes git_only as "the redundant one" and silently re-couples version determination to the registry. Add half a sentence stating the pair is intentional and which one is load-bearing.

release-plz.toml
```diff
@@ -1,15 +1,27 @@
...
+
+# git_only = true decouples release-plz from the crates.io registry, instructing it
+# to generate changelogs, tag commits, and publish GitHub releases without attempting
+# to upload package archives to a package registry.
```
martcpp: If true, release-plz will look for existing git tags to determine the current version, rather than checking the cargo registry." That version-source switch is the primary effect and it's the direct reason semver_check has no baseline — the link that makes the two comments cohere. The current text gestures at it with "decouples … from the crates.io registry" and then spends its words on the skip-publish half, which is the part publish = false already cover

martcpp (changes requested):

release-plz.toml
```diff
@@ -1,15 +1,27 @@
+# Note: release-plz's [workspace] table defines default configuration settings
+# for release-plz across packages; it is unrelated to Cargo workspaces (this crate has none).
 [workspace]
 changelog_path = "./CHANGELOG.md"
 changelog_config = "cliff.toml"
 release_always = false
+
+# semver_check is disabled because this crate is not published to crates.io,
```
Ngha-Boris: **Issue:** The comment explains the technical reason (no published baseline) but **misses the key architectural rationale** stated in the ticket: that this is a **standalone binary crate** with no public library API contract to validate. The ticket specifically calls out "binary crate with no public library API contract" as the rationale.

**Suggestion:**
```toml
# semver_check is disabled because this is a standalone binary crate, not a library
# published on crates.io. Since there is no public library API contract exposed to
# external callers, cargo-semver-checks has no baseline version on the registry to
# compare API changes against.
semver_check = false
```

release-plz.toml
```diff
@@ -1,15 +1,27 @@
...
+# so cargo-semver-checks has no published baseline version on the registry to
+# compare API changes against.
 semver_check = false
-# For an initial release without a matching tag, create a one-commit baseline.
-# After the first tag is created, release-plz automatically uses commits since that tag.
+
+# Limits commit analysis depth when searching git history for release baselines.
 max_analyze_commits = 1
```
Ngha-Boris: **Issue:** The comment is **incomplete** per the ticket requirements. It only states the negative (not published to crates.io) without explaining **how** the crate is actually distributed. The ticket specifically requires documenting that "crate is published via Docker images / GitHub releases, not crates.io".

**Suggestion:**
```toml
# publish is disabled because this crate is distributed via Docker images and
# GitHub releases, not published to crates.io.
publish = false
```

src/lib.rs
```diff
@@ -1,3 +1,9 @@
+//! `status-list-server` internal library target.
```
Ngha-Boris: **Issue:** This file modification is **out of scope** for the ticket. The ticket only asks for documentation in `release-plz.toml`. While this documentation is helpful, it should ideally be in a separate PR or explicitly included in the ticket scope.

**Suggestion:** Either:
- Remove this change from the PR and open a separate PR for it, OR
- Update the ticket description to include this scope expansion

Ngha-Boris (changes requested):
Hello @Blindspot22 Please see the comments for specific suggestions.

src/lib.rs
martcpp: i will suggest to let go will you review a 6 line comment change in the futurs

martcpp (changes requested):

Hermann-Core:
@Blindspot22 what is done here doesn't align with the ticket scope. Please respect the ticket requirement.

Issue #410, opened 2026-08-20
Scheduled re-scan of the deployed image digest


martcpp:
Follow-up to #388.

`docs/supply-chain.md` names this as the single largest gap in the container supply-chain pipeline, and it is the common cause of three others.

## Problem

`.github/workflows/deploy.yml` triggers only on pushes to `main`, on release tags matching `v*.*.*`, and on manual dispatch. The image scan therefore runs a handful of times a month, always tied to someone shipping something. Three consequences follow:

1. **Post-merge advisories are not detected.** A newly published RustSec advisory affecting a shipped crate first surfaces whenever someone next cuts a release. This is the capability usually claimed for container image scanning, and the pipeline does not currently deliver it. `cargo-audit` / `cargo-deny` / `cargo-vet` cover the source tree on every PR, but only for commits that exist — an advisory published against an unchanged lockfile triggers nothing.

2. **`expired_at` lapses surface as blocked releases.** Once the gate blocks (#409), an expiring exception in `.trivyignore.yaml` fires at the only moment this pipeline runs — during a release — rather than as advance warning. Today the guidance in `docs/supply-chain.md` is "prefer expiry dates that do not fall close to a planned release", which is a workaround for this gap rather than a practice worth keeping.

3. **Release-tag scan results have no durable sink.** The SARIF upload is branch-gated to `main` on purpose: code scanning files an analysis against a ref, and alerts for `refs/tags/*` are not surfaced in the Security tab's branch view, so uploading there would succeed into a place nobody reads. That leaves release runs with results only in run artifacts, which expire with artifact retention.

## Proposal

One scheduled workflow that:

- resolves the image digest currently deployed to production,
- scans it with the same severity floor and the same `.trivyignore.yaml` as `scan-image`, so the scheduled result and the release-time result mean the same thing,
- uploads SARIF against the default branch, so findings land somewhere durable and readable regardless of which ref produced the image,
- opens or updates an issue on failure rather than only turning a run red.

The last point matters: an unmonitored scheduled job that fails silently is indistinguishable from one with nothing to report, which is the same trap the rest of this pipeline is built to avoid.

## Acceptance criteria

- [ ] A scheduled workflow scans the deployed digest on a cron.
- [ ] Scan configuration is shared with, or provably identical to, `scan-image`.
- [ ] Results are visible outside the run that produced them.
- [ ] A failure produces a notification a maintainer actually receives.
[...]

PR #438, opened 2026-08-24
feat(ci): add container image scanning

### LINKED ISSUE #388: Add container image scanning, SBOM, and provenance to CI
## Objective

Strengthen container supply-chain security by scanning the built image and publishing SBOM/provenance metadata from the image build pipeline.

## Scope

- [ ] Add container image vulnerability scanning after the Docker image is built.
- [ ] Use Trivy, Grype, or the repository-standard scanner in image mode.
- [ ] Decide the failure threshold for critical/high CVEs and document exceptions.
- [ ] Enable SBOM generation in the image build/push flow.
- [ ] Enable provenance/attestation generation where supported by the build action and registry.
- [ ] Make scan results visible in GitHub Actions summaries or artifacts.
- [ ] Document how maintainers triage scanner findings.

## Acceptance Criteria

- [ ] CI scans the actual built container image, not only Helm manifests.
- [ ] CI publishes or stores an SBOM for release images.
- [ ] CI publishes provenance/attestation metadata where supported.
- [ ] Critical/high vulnerability behavior is explicit and reproducible.
- [ ] Maintainers have a documented triage path for findings and false positives.

## Verification

- [ ] Image build workflow emits scan output.
- [ ] SBOM artifact or registry-attached SBOM is visible.
- [ ] Provenance/attestation is visible for pushed images.
- [ ] A test/fake finding demonstrates the configured failure behavior where practical.

## Related Issues

- Source analysis: #380
- Production readiness EPIC: #141
- Local/remote CI parity: #316

est: 2 days

### LINKED ISSUE #403: CI: kube-linter SARIF output written but never uploaded
## Summary

`CI.yml:300-306` runs kube-linter and writes `kube-linter-results.sarif`. There is no
upload step after it  the next line is `yaml-lint:` at `CI.yml:308`. The file is never
uploaded to code scanning and never attached as an artifact, so it is destroyed with the
runner at the end of the job.

## Impact

Findings are unreachable. The only way a maintainer sees a kube-linter result today is by
scrolling the raw job log, which does not happen in practice.

This is worse than the job not existing. A reader of the workflow reasonably concludes that
Kubernetes manifest misconfiguration is covered in CI. It is not covered in any way that
reaches a human. An absent check invites someone to add one; a check that runs and discards
its output invites nobody.

Regardless of the original intent, the job is advisory-only as configured.

## Fix

Upload the SARIF to code scanning after the kube-linter step.

The supply-chain PR () establishes this repo's first SARIF upload
pattern and its `security-events: write` job-permission convention. Follow that pattern
rather than introducing a second style — one upload mechanism, one permission convention,
one place to look when container and manifest findings both need triage.

The repo is public, so code scanning upload requires no GitHub Advanced Security.

## Notes

- Scope is `CI.yml` only. Deliberately not folded into the supply-chain PR, which touches
  `deploy.yml`; mixing them would obscure that diff.
- Blocked on nothing. Can land independently once the upload pattern exists.

## Acceptance criteria

- [ ] kube-linter SARIF is uploaded to code scanning and findings appear in the Security tab
- [ ] `security-events: write` is scoped to that job only, matching the repo's per-job
      least-privilege style

### LINKED ISSUE #406: Enforce expired_at and `statement` on .trivyignore.yaml entries
# Enforce `expired_at` and `statement` on `.trivyignore.yaml` entries

## Summary

The `vulnerabilities:` block added to `.trivyignore.yaml` requires every entry to carry an `id`, a written `statement`, and an `expired_at` date. That requirement exists only in prose  in the file's header comment and in `docs/supply-chain.md`. Nothing checks it.

Trivy itself does not require these fields. An entry with only `id` is valid and will suppress findings indefinitely.

## Why it matters

The `expired_at` requirement is the entire reason this is an exception ledger rather than a
suppression list. It forces each exception to be re-reviewed on a date instead of quietly
becoming permanent.

Unenforced, the decay path is undramatic and near-certain:

1. Someone under release pressure adds an entry without a date. Nothing complains.
2. The next person copies the shape of the entry already in the file.
3. Within a year the file is a list of undated suppressions and the design intent is gone.

There is no point in that sequence where anyone does anything obviously wrong, which is
exactly why a comment does not prevent it. A rule nobody enforces is a hope.

The security consequence is concrete: suppressions outliving their justification means the
gate reports green on findings nobody has looked at since they were first waved through.

## Proposed fix

A lightweight CI check on `.trivyignore.yaml`, failing when any entry under
`vulnerabilities:` is missing `statement` or `expired_at`.

Worth deciding when picked up:

- Whether an **already-expired** `expired_at` should fail CI or warn. Failing is the stronger
  posture but turns an unrelated PR into a triage task, so warn-plus-summary may be the
  practical choice.
- Whether to extend the same requirement to the existing `misconfigurations:` block, which
  predates this convention and currently has no such fields.

Implementation can be a small script or an existing YAML schema validator — no preference,
lightweight is the requirement.

## Why this wasn't in the supply-chain PR

Scope. That PR establishes the convention; enforcing it is separable work with its own
decisions, and the convention has value from day one even unenforced. Recorded here so the
enforcement gets built while the reasoning behind `expired_at` is still legible.

## Acceptance criteria

- [ ] CI fails when a `vulnerabilities:` entry lacks `statement` or `expired_at`
- [ ] Expired-entry behaviour decided and documented
- [ ] `docs/supply-chain.md` states the check exists and what it rejects
- [ ] The check runs on PRs touching `.trivyignore.yaml`

martcpp:
## Objective

Strengthen container supply-chain security by scanning the built image and publishing SBOM/provenance metadata from the image build pipeline.

## Scope

- [ ] Add container image vulnerability scanning after the Docker image is built.
- [ ] Use Trivy, Grype, or the repository-standard scanner in image mode.
- [ ] Decide the failure threshold for critical/high CVEs and document exceptions.
- [ ] Enable SBOM generation in the image build/push flow.
- [ ] Enable provenance/attestation generation where supported by the build action and registry.
- [ ] Make scan results visible in GitHub Actions summaries or artifacts.
- [ ] Document how maintainers triage scanner findings.

## Acceptance Criteria

- [ ] CI scans the actual built container image, not only Helm manifests.
- [ ] CI publishes or stores an SBOM for release images.
- [ ] CI publishes provenance/attestation metadata where supported.
- [ ] Critical/high vulnerability behavior is explicit and reproducible.
- [ ] Maintainers have a documented triage path for findings and false positives.

## Verification

- [ ] Image build workflow emits scan output.
- [ ] SBOM artifact or registry-attached SBOM is visible.
- [ ] Provenance/attestation is visible for pushed images.
- [ ] A test/fake finding demonstrates the configured failure behavior where practical.

**note** there are fellow issue her to fix everything else https://github.com/adorsys/status-list-server/issues/388


cliff.toml
```diff
@@ -133,6 +133,9 @@ protect_breaking_commits = true
 filter_commits = false
 # glob pattern for matching git tags
 tag_pattern = "v[0-9]*"
+# regex for skipping tags — leave empty; skip_tags anchors the comparison range
+# which is wrong for pre-releases (use ignore_tags instead)
+skip_tags = ""
```
Hermann-Core: This will break changelog generation; it should not be set. Please revert back.

Hermann-Core (changes requested):
## Required before merge

The core release path is heading in the right direction: build once, scan by digest, promote only after the gate, and deploy the digest that was scanned. However, this PR is substantially larger than the stated image-scanning scope (1,792 additions across 25 files) and needs to be reduced before it is safe to maintain.

- Scan every published platform image, not just amd64.
- Keep the release scan summary concise; retain full Trivy output only as an artifact.
- Move image-reference rendering verification into its own workflow, instead of growing the Helm config scan job.
- Remove the unrelated hardening/release/configuration changes and submit them separately under the referenced follow-up work.
- Replace the generated-looking implementation diary with short operator documentation, and remove comments that merely narrate obvious code. Document only durable decisions and operator actions.

I also do not think this PR needs a bespoke YAML-policy validator plus 304 lines of tests and an unpinned runtime `pip install` to deliver image scanning. Keep the exception policy simple, use Trivy’s supported file format directly, and add extra policy automation only when it has a separately agreed owner and scope.

.github/workflows/deploy.yml
```diff
@@ -36,47 +60,436 @@ jobs:
...
+    name: Scan Image for Vulnerabilities
+    runs-on: ubuntu-latest
+    needs: build-and-push
+    timeout-minutes: 20
+    permissions:
+      contents: read
+      packages: read
+    env:
+      IMAGE_REF: ${{ needs.build-and-push.outputs.image_ref }}
+      # Vulnerability scanning covers one architecture: both are built from the same
```
Hermann-Core: The release publishes both `linux/amd64` and `linux/arm64`, but the gate scans only amd64. A shared lockfile and a scratch final stage do not prove that target-specific Rust dependencies, compiled binaries, or scanner output are identical. Scan each manifest that we publish (or publish one platform); otherwise a HIGH/CRITICAL finding unique to arm64 can be promoted and deployed without ever crossing the gate.

.github/workflows/deploy.yml
```diff
@@ -36,47 +60,436 @@ jobs:
...
+            echo
+            echo "### Published SBOM"
+            echo
+            cat sbom-summary.md 2>/dev/null || echo "_SBOM collection did not complete._"
+            echo
+            echo "### Findings"
+            echo
+            echo '```text'
+            # Step summaries are capped at 1 MiB.
+            head -c 900000 trivy-image-report.txt 2>/dev/null || echo "No report produced."
```
Hermann-Core: Please keep the Actions summary deliberately small: digest, platforms, totals, blocking HIGH/CRITICAL findings, and an artifact link/name are enough. Writing up to 900 KB of raw Trivy text makes the release UI noisy and can bury the actual gate decision. The complete JSON/table reports are already correctly preserved as artifacts.

.github/workflows/CI.yml
```diff
@@ -261,6 +306,15 @@ jobs:
       - name: Fetch Repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      - name: Validate .trivyignore.yaml
+        # Before the scan that consumes the file, so a malformed entry fails in seconds
+        # as itself rather than later as a scanner error on an unrelated change.
+        run: |
+          set -eu
+          python3 -c 'import yaml' 2>/dev/null || pip install --quiet pyyaml
```
Hermann-Core: This adds an unpinned `pip install` and a bespoke 219-line validator plus 304-line test suite to the generic Helm configuration job, although the PR's objective is release-image scanning. Please remove this from this PR. If policy validation is later needed, define its ownership and dependency pinning in a focused change; do not make every PR fetch an extra runtime dependency for an otherwise empty exception ledger.

.github/workflows/CI.yml
```diff
@@ -277,6 +331,81 @@ jobs:
           mkdir -p /tmp/rendered
           helm template status-list-server helm/chart -f helm/chart/values.yaml --output-dir /tmp/rendered
 
+      - name: Verify image reference resolution
```
Hermann-Core, line 339: Please move this image-reference verification into a dedicated workflow file (for example, `verify-image-reference.yml`) with focused triggers. It is valuable coverage, but it is unrelated to a Trivy Helm-config scan and this 74-line inline script makes the main CI workflow noisy and difficult to own. A separate workflow gives it an explicit name, path filters, and an independent failure signal.

.github/workflows/auto_merge.yml
```diff
@@ -15,7 +15,17 @@ permissions:
 jobs:
   dependabot:
     runs-on: ubuntu-latest
-    if: github.actor == 'dependabot[bot]'
+    # `github.event.pull_request.user.login`, not `github.actor`. `github.actor` is
```
Hermann-Core, line 18: This is a worthwhile security hardening change, but it is unrelated to building/scanning the release image. The PR explicitly points to #388 for the remaining work. Please remove this and the other unrelated hardening/release/configuration changes from this PR, then send them as focused follow-up PRs so their security semantics can be reviewed independently.

docs/supply-chain.md
```diff
@@ -0,0 +1,371 @@
+# Container Supply Chain
+
+This document covers how the container image is scanned, what metadata it carries, and how maintainers triage findings. The pipeline is `.github/workflows/deploy.yml`.
```
Hermann-Core: Please replace this 371-line implementation narrative with a concise operator guide. It should state the gate threshold, where to retrieve the SBOM/provenance/scan artifact, how to triage a finding, and the narrowly defined exception process. Details such as internal job sequencing, retry mechanics, tool-pin ownership, hypothetical failure modes, and known implementation gaps belong in the workflow, an issue, or an ADR only when they are durable architectural decisions. The same reduction is needed for the new explanatory comments: descriptive names and commands should carry the obvious `what`; comments should retain only non-obvious long-term `why`.

Hermann-Core (changes requested):
Inline details for the required changes: platform-complete scanning, concise release output, a dedicated image-reference workflow, removal of unrelated work, and an operator-focused documentation rewrite.

.github/workflows/deploy.yml
martcpp: There is and  issue for that open already

.github/workflows/auto_merge.yml
martcpp: You can fix sub issue that are attach to it that so this pr closes those issue also. There not need open sub pr to fix few line of ci check

.github/workflows/CI.yml
martcpp: > Please move this image-reference verification into a dedicated workflow file (for example, `verify-image-reference.yml`) with focused triggers. It is valuable coverage, but it is unrelated to a Trivy Helm-config scan and this 74-line inline script makes the main CI workflow noisy and difficult to own. A separate workflow gives it an explicit name, path filters, and an independent failure signal.

The noise complaint is fair 74 inline lines is too much for one step. I will rather not make it a separate workflow, though. It asserts on a rendered chart, so a standalone workflow has to repeat setup-helm, helm dependency build and helm template to get back to the state this job is already in. And path filters would make it skippable, which is the wrong property for the check that underwrites "production runs the digest CI scanned" a change to deployment.yaml isn't the only way to break it.

what i will do rather his extract it to scripts/verify-image-reference.sh, the same pattern this PR already uses for scripts/vuln-gate.sh. That removes the inline bulk from the workflow, gives it a name and an owner, and makes it runnable locally  without duplicating the Helm setup or letting it be skipped. If you still want the independent failure signal after that, it's a one-line job split.

docs/supply-chain.md
martcpp: > Please replace this 371-line implementation narrative with a concise operator guide. It should state the gate threshold, where to retrieve the SBOM/provenance/scan artifact, how to triage a finding, and the narrowly defined exception process. Details such as internal job sequencing, retry mechanics, tool-pin ownership, hypothetical failure modes, and known implementation gaps belong in the workflow, an issue, or an ADR only when they are durable architectural decisions. The same reduction is needed for the new explanatory comments: descriptive names and commands should carry the obvious `what`; comments should retain only non-obvious long-term `why`.

The operator guide is a real gap: someone paged at 2am shouldn't read 371 lines to find the threshold and the artifact name. I'll add that as a short section at the top threshold, where to get the SBOM/provenance/scan artifact, how to triage, the exception process.

I will keep the rationale rather than delete it, and move it below that section. By your own standard it's the durable-decision material: why there are two unconnected suppression ledgers (deny.toml and .trivyignore.yaml) and what breaks when they drift; why --all-features doesn't cover the release feature set; and the Known Gaps, which are linked to tracked issues (#410) rather than hypothetical. Deleting that loses the reasoning and the next person re-derives it  an ADR is a fine destination, but a move, not a drop

.github/workflows/auto_merge.yml
martcpp: > This is a worthwhile security hardening change, but it is unrelated to building/scanning the release image. The PR explicitly points to #388 for the remaining work. Please remove this and the other unrelated hardening/release/configuration changes from this PR, then send them as focused follow-up PRs so their security semantics can be reviewed independently.

I will rather keep these in one PR, and I don't think the split gets you what you're after.

The auto_merge.yml change is twelve lines on a single if: condition, with the reasoning written inline above it. Reviewing it doesn't require isolation from the rest of the diff  GitHub's Files tab already gives you that view, and the security semantics are entirely local to that one expression. What a separate PR would add is three CI runs instead of one and six required approvals instead of two, on a repo where this PR has already been waiting on review bandwidth.

What I think you are actually asking for is that these changes be reviewable independently, and I can give you that without the overhead: I'll split them into their own commits on this branch one for the Dependabot guard, one for the release-plz token permissions  so each is a self-contained diff you can review and sign off on separately. If either one doesn't hold up, it comes out on its own without touching the rest.

If after that you still want them as separate PRs, I'll do it  but I will like the scanning work to land rather than fan out into four PRs that each need two approvals.

.github/workflows/deploy.yml
```diff
@@ -1,33 +1,57 @@
...
   IMAGE_NAME: status-list-server
   CLUSTER_NAME: datev-wallet-cluster
   AWS_REGION: eu-central-1
   REGISTRY: ghcr.io
+  IMAGE_BASE: ghcr.io/${{ github.repository_owner }}/status-list-server
   HELM_RELEASE_NAME: statuslist
   HELM_NAMESPACE: statuslist-production
 jobs:
   ci:
     name: CI
```
Ngha-Boris, line 28: **Issue:** The `image_ref` output uses `${{ env.IMAGE_BASE }}`, but `IMAGE_BASE` is defined at the workflow `env:` level. Job outputs cannot reference workflow-level env variables directly in this way - the expression may not resolve correctly.

**Suggestion:** Move the IMAGE_REF construction to the `vars` step and output it there:
```yaml
- name: Resolve deploy tag
  id: vars
  run: |
    if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
      echo "DEPLOY_TAG=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
    else
      echo "DEPLOY_TAG=sha-${GITHUB_SHA::7}" >> "$GITHUB_OUTPUT"
    fi
    echo "IMAGE_REF=${IMAGE_BASE}@${{ steps.build.outputs.digest }}" >> "$GITHUB_OUTPUT"
```

Then reference it here as `${{ steps.vars.outputs.IMAGE_REF }}`

.github/workflows/CI.yml
```diff
@@ -68,6 +92,27 @@ jobs:
       - name: Memory-only Build
         run: cargo check --no-default-features --features memory
 
+      - name: Release image feature set
+        # The feature set the release image is actually built with. Two gaps make it
+        # worth checking on its own. Nothing validates the `ARG FEATURES` string: it is
+        # a bare string handed to `cargo build` inside the image build, so a value
+        # naming no real feature fails only when someone cuts a release. And
```
Ngha-Boris, line 99: **Issue:** The features string is hardcoded here (`postgres,aws-secrets,acme`), but the `cargo-build` job correctly reads it from the Dockerfile. This creates duplication and a maintenance burden.

**Suggestion:** Read the features from the Dockerfile like the `cargo-build` job does:
```yaml
- name: Docker build (production features)
  run: |
    features=$(sed -nE 's/^ARG FEATURES=\"([^\"]+)\".*/\1/p' Dockerfile)
    docker build --build-arg FEATURES="$features" .
```

This ensures a single source of truth for the release feature set.

Ngha-Boris (changes requested):
## Review Summary

The implementation is **comprehensive and well-architected**. It addresses all major ticket requirements with thorough documentation, proper security controls, and good separation of concerns.

### Highlights
- ✅ Container image scanning with Trivy for both architectures (amd64/arm64)
- ✅ SBOM generation with BuildKit and verification
- ✅ SLSA provenance/attestation with `mode=max`
- ✅ Clear failure threshold (HIGH/CRITICAL) with `ignore-unfixed: false`
- ✅ Comprehensive documentation in `docs/supply-chain.md`
- ✅ Self-testing vulnerability gate with fixtures
- ✅ Scan results visible in job summaries and artifacts

### Changes Requested
Two minor issues need to be addressed before merging (see comments below).

martcpp:
@Ngha-Boris @Christiantyemele pls recheck

Hermann-Core (dismissed):
LGTM

Ngha-Boris (approved):
Thank you for considering my comments

Hermann-Core (approved):

Issue #417, opened 2026-08-20
CI: zizmor runs with '|| true' and cannot fail; 84 findings including one High (spoofable bot check in auto_merge.yml)


martcpp:
`CI.yml` runs the repository's workflow security scanner like this:

```yaml
- name: Run Zizmor
  run: zizmor .github/workflows/ || true
```

The `|| true` means it can never fail, so nothing it finds has ever blocked anything. Measured with the pinned version (`zizmor@1.28.0`, `--persona regular`) against `.github/workflows/`:

```
84 findings (35 suppressed, 26 unsafe fixes): 1 informational, 0 low, 47 medium, 1 high
```

## Breakdown

| Rule | Severity | Count |
| --------------------- | ------------- | ----- |
| `artipacked`          | Medium        | 25    |
| `excessive-permissions` | Medium      | 22    |
| `bot-conditions`      | **High**      | 1     |
| `template-injection`  | Informational | 1     |

By file: `CI.yml` 36, `cargo_deny.yml` 5, `deploy.yml` 4, and one each in `crate_type.yml`, `cache.yml`, `auto_merge.yml`, `codeql.yml`.

## The High finding is real and worth looking at first

```
error[bot-conditions]: spoofable bot actor check
  --> .github/workflows/auto_merge.yml:18:9
   |
16 |   dependabot:
17 |     runs-on: ubuntu-latest
18 |     if: github.actor == 'dependabot[bot]'
```

An auto-merge job gated on `github.actor` alone. `github.actor` is not a trustworthy authorization signal — it reflects who triggered the most recent activity, not who authored the change, so the guard is spoofable. Auto-merge is exactly the wrong place to have a bypassable actor check.

This should probably be split out and fixed on its own rather than waiting for the bulk triage below.

## The Medium findings are mostly two mechanical classes

- **`artipacked`** (25): `actions/checkout` without `persist-credentials: false`, which leaves the token in `.git/config` where a later step or a packaged artifact can pick it up. Largely auto-fixable, but each one needs checking for steps that actually rely on the persisted credential.
- **`excessive-permissions`** (22): jobs without an explicit `permissions:` block, inheriting the repository default. Mechanical to fix, and worth doing job by job rather than in bulk.

## Suggested approach

1. Fix `bot-conditions` in `auto_merge.yml` separately and soon.
2. Work through `artipacked` and `excessive-permissions`, ideally per-workflow so each change is reviewable.
[...]

PR #440, opened 2026-08-24
chore: refactor Redis and database configuration to enhance security by using split credentials

### LINKED ISSUE #391: Reduce database and Redis credential exposure in Kubernetes pod metadata
## Objective

Reduce or explicitly document the exposure of assembled database URLs and other sensitive connection strings in Kubernetes pod metadata.

## Context

`APP_DATABASE__URL` can be assembled as an environment variable in the deployment template. Even when the password comes from a Kubernetes Secret, the resulting connection string may be visible to users with pod-inspection permissions.

## Scope

- [ ] Audit Helm templates for assembled sensitive URLs or credentials in environment variables.
- [ ] Decide whether to construct sensitive connection strings inside the container at startup, use separate redacted config fields, or accept the Kubernetes env-var tradeoff with RBAC documentation.
- [ ] Prefer a design that avoids exposing fully assembled database URLs in `kubectl describe pod` when practical.
- [ ] Apply the same reasoning to Redis and future runtime credential references.
- [ ] Ensure logs, errors, and readiness failures never print assembled credentials.
- [ ] Document the chosen operational/security tradeoff.

## Acceptance Criteria

- [ ] SQL database credentials are not exposed as a fully assembled URL in pod metadata, or the accepted exposure is explicitly documented with required RBAC mitigations.
- [ ] Redis credentials receive the same treatment when Redis is enabled.
- [ ] Startup/config errors identify missing fields without printing secret values.
- [ ] Helm examples use the safer pattern.
- [ ] Tests or template checks cover the selected pattern.

## Verification

- [ ] Helm template review for sensitive env values.
- [ ] Config/startup tests for redacted errors.
- [ ] Manual Kubernetes review: `kubectl describe pod` does not reveal full credentials if the safer pattern is implemented.

1 day

Ngha-Boris:
### Summary
This PR implements a security enhancement to prevent assembled database URLs and Redis URIs from being exposed in Kubernetes pod metadata. Instead of constructing connection strings in Helm templates (which would be visible via `kubectl describe pod`), the application now assembles credentials internally from split configuration fields.

---

### Motivation
The `APP_DATABASE__URL` and `APP_REDIS__URI` environment variables were previously assembled in the deployment template using Kubernetes variable expansion (e.g., `postgres://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@...`). Even when the password comes from a Kubernetes Secret, the resulting connection string is visible to any user with pod-inspection permissions.

This change moves the URL assembly into the application container at startup, so `kubectl describe pod` only shows:
- Non-sensitive fields (host, port, scheme, username)
- Secret references (`valueFrom.secretKeyRef`) rather than actual values

---

#### 1. `src/config.rs`
- **Added split credential fields to `DatabaseConfig`**:
  - `host`, `port`, `username`, `password`, `name`
  - Added `resolved_url()` method to assemble URL internally
  
- **Added split credential fields to `RedisConfig`**:
  - `scheme`, `host`, `port`, `username`, `password`, `database`
  - Added `resolved_uri()` method to assemble URI internally
  
- **Security features**:
  - All sensitive values use `SecretString` type
  - Error messages only contain field names, never values
  - URL encoding for username/password components to handle special characters
  - Backward compatibility maintained: if split fields aren't provided, falls back to full URL/URI

#### 2. `src/setup.rs`
- Updated database connection to use `config.database.resolved_url()` instead of directly using `config.database.url`
- Added error wrapping to provide context without exposing sensitive data

#### 3. `helm/chart/templates/deployment.yaml`
- **Replaced assembled database URL**:
  - Removed: `APP_DATABASE__URL` with `$(POSTGRES_PASSWORD)` expansion
  - Added: Split fields `APP_DATABASE__HOST`, `APP_DATABASE__PORT`, `APP_DATABASE__USERNAME`, `APP_DATABASE__PASSWORD` (via secretKeyRef), `APP_DATABASE__NAME`
  
- **Replaced assembled Redis URI**:
  - Removed: `APP_REDIS__URI` with `$(REDIS_PASSWORD)` expansion
  - Added: Split fields `APP_REDIS__SCHEME`, `APP_REDIS__HOST`, `APP_REDIS__PORT`, `APP_REDIS__PASSWORD` (via secretKeyRef)
  
- Made database port configurable via `.Values.postgres.auth.port` with default "5432"

#### 4. `helm/chart/templates/_helpers.tpl`
- Replaced single `redisUri` helper with three separate helpers:
  - `redisScheme` - returns "redis" or "rediss" based on TLS configuration
  - `redisHost` - returns the Redis hostname
  - `redisPort` - returns the Redis port
- These helpers no longer assemble credentials into a URI

#### 5. `helm/README.md`
- Added "Credential Exposure Model" section documenting:
  - The split field approach
  - Security trade-offs
  - RBAC recommendations
  - Backward compatibility note for `APP_DATABASE__URL` and `APP_REDIS__URI`

#### 6. `docs/REDIS_TLS_SETUP.md`
- Updated Redis configuration examples to use split fields instead of assembled URI
- Added explanatory note about the security improvement

#### 7. `tests/helm_sensitive_env.rs` (new file)
- Added template guard test that verifies:
  - No `APP_DATABASE__URL` or `APP_REDIS__URI` in deployment
  - No `$(POSTGRES_PASSWORD)` or `$(REDIS_PASSWORD)` expansion
  - All required split fields are present
  - Secret references are properly configured

---

### Security Improvements
1. **Pod metadata no longer contains assembled credentials** - only non-sensitive fields and Secret references
2. **Error messages are redacted** - missing field errors only mention the config key, never the value
3. **URL encoding prevents injection** - special characters in credentials are properly encoded
4. **RBAC defense in depth** - documentation emphasizes restricting pod inspection, Secret access, and container exec permissions

---

### Backward Compatibility
- The application still supports `APP_DATABASE__URL` and `APP_REDIS__URI` for:
  - Local development workflows
  - Existing custom deployments
  - External Redis instances (not using redis-ha)
- If no split fields are provided, the application uses the full URL/URI as before
- Helm templates only use split fields when `redis-ha.enabled=true`

---

### Testing
- [x] `cargo test` passes all existing tests
- [x] New `helm_sensitive_env.rs` test verifies template security
- [x] Config tests verify URL/URI resolution with split fields
- [x] Error message tests verify no credential leakage


src/config.rs
martcpp, line 1: sea-orm prints the assembled URL with password into kubectl logs on connect failure. Fix this first; it's the one that can make the PR a net exposure regression.

src/config.rs
```diff
@@ -565,13 +565,116 @@ pub struct DatabasePoolConfig {
...
+    #[serde(default)]
+    pub host: Option<String>,
+    #[serde(default)]
+    pub port: Option<u16>,
+    #[serde(default)]
+    pub username: Option<String>,
+    #[serde(default)]
+    pub password: Option<SecretString>,
+    #[serde(default)]
+    pub name: Option<String>,
```
martcpp, line 581: no TLS config path; sslmode=verify-full deployments silently downgrade to Prefer.

helm/chart/templates/deployment.yaml
```diff
@@ -132,10 +127,21 @@ spec:
             - name: AWS_CONFIG_FILE
               value: /home/nobody/.aws/config
             {{- end }}
-            - name: POSTGRES_DB
+            - name: APP_DATABASE__BACKEND
+              value: "postgres"
```
martcpp: hardcoded APP_DATABASE__BACKEND after the $env loop can silently point a MySQL deployment at the wrong database.

src/config.rs
```diff
@@ -565,13 +565,116 @@ pub struct DatabasePoolConfig {
...
+                .is_some_and(|value| !value.trim().is_empty())
+            || self.port.is_some()
+    }
+
+    /// Resolve the database connection string from either the backward-compatible
+    /// full URL or the split connection fields used by the Helm chart.
+    pub fn resolved_url(&self) -> Result<SecretString, ConfigError> {
+        if !self.has_split_connection_fields() {
+            return Ok(self.url.clone());
+        }
```
martcpp, line 733: silent url vs. split-field precedence (root cause of 2 and 3).

tests/helm_sensitive_env.rs
```diff
@@ -0,0 +1,37 @@
+#[test]
+fn deployment_template_uses_split_database_credentials() {
+    let deployment = include_str!("../helm/chart/templates/deployment.yaml");
+    let helpers = include_str!("../helm/chart/templates/_helpers.tpl");
+
+    assert!(
+        !deployment.contains("APP_DATABASE__URL"),
+        "Helm deployment must not render a fully assembled database URL in pod metadata"
+    );
```
martcpp, line 120: replace the substring test with an assertion over helm template output.

martcpp (changes requested):

Ngha-Boris:

martcpp:
> Hello @martcpp please check again
@Ngha-Boris Still open from the original review
Values.postgres.auth.port is still a phantom keyreal key is postgres.service.port.) still hardcodes 5432, so using the knob firewalls the pod off from its database.
 password still silently trim()-ed at config.rs .
database.host still unvalidated.
split fields still absent from .env.template and docs/database-backends.md.
env-var-secret still excluded in .kube-linter.yaml.
dead POSTGRES_USER in values.yaml; four dead dandydeveloper repo-adds.
branch still bundles four unrelated workstreams.

tests/helm_sensitive_env.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+#[test]
+fn rendered_chart_respects_database_backend_override() {
+    let rendered = render_helm(&["--set", "statuslist.env.APP_DATABASE__BACKEND=mysql"]);
+
+    assert!(
+        rendered.contains("name: APP_DATABASE__BACKEND\n              value: \"mysql\""),
+        "rendered Helm output must preserve operator-provided database backend"
```
martcpp, line 161: No test for the APP_DATABASE__PASSWORD fail guard (only APP_DATABASE__URL has one), and no test that APP_DATABASE__QUERY renders.

tests/helm_sensitive_env.rs
```diff
@@ -0,0 +1,92 @@
...
+            "--set",
+            "statuslist.env.APP_DATABASE__URL=postgres://user:pass@db:5432/status-list",
+        ])
+        .output()
+        .expect("failed to execute helm template");
+
+    assert!(
+        !output.status.success(),
+        "helm template should reject APP_DATABASE__URL in statuslist.env"
+    );
```
martcpp: the APP_DATABASE__PASSWORD fail guard is right, but combined with the now-overridable APP_DATABASE__HOST it creates a half-supported external-DB path: you can point at RDS, but the password must live in the chart's postgres-password secret key. Workable, undocumented, surprising

src/config.rs
```diff
@@ -565,13 +565,167 @@ pub struct DatabasePoolConfig {
...
+                .to_string()
+            });
+            return format!(
+                "backend={}, host={}, port={}, database={}",
+                self.backend.as_str(),
+                host,
+                port,
+                name
+            );
+        }
```
martcpp: the <missing-host> / <missing-name> placeholders are unreachable from the only call site (resolved_url() already errored). And the Memory | Sqlite => 0 port arm renders port=0, which will confuse anyone who sees it. Harmless, but it's speculative code in a security-critical helper.

martcpp (changes requested):

src/config.rs
```diff
@@ -565,13 +565,167 @@ pub struct DatabasePoolConfig {
...
+            .as_ref()
+            .and_then(|value| trim_non_empty(Some(value.expose_secret())))
+            .ok_or_else(|| {
+                ConfigError::Message("Missing required config field: database.password".to_string())
+            })?;
+        let name = required_config_field(self.name.as_deref(), "database.name")?;
+        let port = self.port.unwrap_or(default_port);
+        let query = trim_non_empty(self.query.as_deref())
+            .map(|query| format!("?{}", query.trim_start_matches('?')))
+            .unwrap_or_default();
```
martcpp, line 769: validate database.query keys at load time.

tests/helm_sensitive_env.rs
```diff
@@ -0,0 +1,92 @@
...
+        .output()
+        .expect("failed to execute helm template");
+
+    assert!(
+        output.status.success(),
+        "helm template failed: {}",
+        String::from_utf8_lossy(&output.stderr)
+    );
+
+    String::from_utf8(output.stdout).expect("helm template output should be valid UTF-8")
```
martcpp: restore a green build (move assertions to trivy-config, or add azure/setup-helm + helm dependency build to cargo-nextest and make the tests skip gracefully when helm is absent, so contributors without helm aren't blocked).

src/config.rs
```diff
@@ -565,13 +565,167 @@ pub struct DatabasePoolConfig {
...
+    }
+
+    fn url_is_default(&self) -> bool {
+        matches!(
+            self.url.expose_secret(),
+            "postgres://postgres:postgres@localhost:5432/status-list"
+                | "mysql://mysql:mysql@localhost:3306/status-list"
+                | "sqlite::memory:"
+                | "memory:"
+                | "memory"
```
martcpp: url: Option<SecretString>; delete the hardcoded default list.

src/setup.rs
```diff
@@ -191,9 +195,12 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
 
-            let db = sea_orm::Database::connect(opt)
-                .await
-                .wrap_err("Failed to connect to database")?;
+            let db = sea_orm::Database::connect(opt).await.map_err(|_| {
+                color_eyre::eyre::eyre!(
+                    "Failed to connect to database ({})",
+                    config.database.redacted_target()
+                )
+            })?;
```
martcpp: classify the DbErr instead of |_|; add the sentinel-password test.

helm/README.md
martcpp, line 1: decide and document the in-cluster TLS posture.

martcpp (changes requested):

Ngha-Boris:
@martcpp please recheck

.github/workflows/deploy.yml
```diff
@@ -11,6 +11,7 @@ env:
   REGISTRY: ghcr.io
   HELM_RELEASE_NAME: statuslist
   HELM_NAMESPACE: statuslist-production
+  APP_DATABASE_PORT: ${{ vars.APP_DATABASE_PORT }}
```
Christiantyemele, line 26: Important (deployment risk): the deploy job now hard-depends on `vars.APP_DATABASE_PORT` (line 14), consumed at line 107 via `--set-string statuslist.env.APP_DATABASE__PORT="${APP_DATABASE_PORT}"`. If this GitHub Actions repository variable is not created, the value is empty and the new `fail` guard in `deployment.yaml` ("APP_DATABASE__PORT is required") aborts every release-tag deploy. CI never catches this because the trivy/kube-linter jobs hardcode 5432, and `docs/deployment-runbook.md` prerequisites (which list `AWS_DEPLOY_ROLE_ARN`) do not mention this variable. Recommend: (1) create the repo variable, (2) document it in the runbook prerequisites, and (3) optionally fail fast with a clear message when unset (e.g. `[[ -n "${APP_DATABASE_PORT}" ]] || { echo 'APP_DATABASE_PORT repo variable is required' >&2; exit 1; }`).

tests/helm_sensitive_env.rs
```diff
@@ -0,0 +1,235 @@
...
+version: 0.1.0
+appVersion: "1.0.0"
+"#,
+    )
+    .expect("failed to write dependency-free Chart.yaml");
+
+    chart_dir
+}
+
+fn helm_available() -> bool {
```
Christiantyemele, line 53: Suggestion: the security assertions (no `APP_DATABASE__URL`, no `postgres://`, no `$(POSTGRES_PASSWORD)`) silently skip when helm is absent — `helm_available()` returns false, `render_helm`/`render_helm_failure` return `None`, and each test early-returns. The `cargo-nextest` CI job (CI.yml:106-124) installs only nextest, never helm, so these assertions are permanently skipped in CI and run only on a contributor's machine. This makes the regression guards for the core security property effectively dead in CI. Consider adding `azure/setup-helm` (+ `helm dependency build`) to the cargo-nextest job so the assertions are enforced, or folding the assertions into the existing trivy-config/kube-linter render step that already runs helm.

src/config.rs
```diff
@@ -565,13 +565,225 @@ pub struct DatabasePoolConfig {
...
+            return Err(ConfigError::Message(format!(
+                "Invalid database.query: credential-like query parameter key '{key}' is not allowed"
+            )));
+        }
+    }
+
+    Ok(())
+}
+
+fn validate_database_host(host: &str) -> Result<(), ConfigError> {
```
Christiantyemele, line 678: Suggestion: `validate_database_host` rejects any host containing ':' (line 667: `host.contains(':')`), so IPv6 literal endpoints (e.g. `fd00::1`) can never be used with the split-field path. It also rejects a trailing dot via `ends_with('.')` (line 673), which is a valid FQDN form. These are legitimate endpoints. Consider special-casing IPv6 (allow ':' when the value parses as a valid IPv6 address) or at least documenting the restriction in the config docs.

src/config.rs
```diff
@@ -804,7 +1018,7 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("server.port", 8000)?
         .set_default("server.enable_metrics", false)?
         .set_default("server.aggregation_uri", Option::<String>::None)?
-        .set_default("database.url", default_db_url)?
+        .set_default("database.url", Option::<String>::None)?
```
Christiantyemele, line 1037: Suggestion: removing the hardcoded default URL (`set_default("database.url", Option::<String>::None)`) is the intended fail-closed change, but it silently changes startup for non-postgres builds. For example a `--features sqlite` build with no env config previously booted to `sqlite::memory:` and now aborts in `resolved_url()` (lines 710-718) with "Missing required config field: database.url or split database fields". This can break existing local/dev scripts without an obvious cause. Consider a short migration note or a `sqlite::memory:` default guard so the failure is self-explanatory.

helm/chart/templates/deployment.yaml
```diff
@@ -132,10 +148,11 @@ spec:
             - name: AWS_CONFIG_FILE
               value: /home/nobody/.aws/config
             {{- end }}
-            - name: POSTGRES_DB
-              value: {{ .Values.postgres.auth.database | quote }}
-            - name: APP_DATABASE__URL
-              value: "postgres://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@{{ .Release.Name }}-postgres.{{ .Release.Namespace }}.svc.cluster.local:5432/$(POSTGRES_DB)"
+            - name: APP_DATABASE__PASSWORD
```
Christiantyemele, line 163: Nit: `APP_DATABASE__PASSWORD`'s `secretKeyRef` key is hardcoded to `postgres-password` regardless of `APP_DATABASE__BACKEND`. When the chart is pointed at a MySQL/MariaDB external database via split fields, the secret key must still literally be named `postgres-password`, which is misleading. helm/README does say "key named postgres-password", but consider deriving the key from the backend (or documenting it as a hard requirement for non-postgres backends) to avoid operator confusion.

Christiantyemele (changes requested):
..

.github/workflows/deploy.yml
Ngha-Boris: @martcpp what do you think about this

martcpp: Christian's right, and it's the same class as the branch-protection settings we just wired up on #417: a control living in repo config with nothing detecting its absence. Unset `vars.APP_DATABASE_PORT` renders empty rather than erroring, so the empty value reaches Helm, the chart's `fail` guard fires, and every release-tag deploy aborts. CI can't catch it because trivy and kube-linter render with hardcoded values.

Taking (1) and (2). On (3) just change the shape: since CI hardcodes 5432 and the port doesn't vary across our environments, a default is better than a guard  `${{ vars.APP_DATABASE_PORT || '5432' }}` removes the failure mode rather than making it readable. A guard turns a silent break into a legible one; a default means there's nothing to break.

helm/chart/templates/deployment.yaml
```diff
@@ -78,6 +78,15 @@ spec:
...
             {{- end }}
+            {{- if hasKey $env "APP_DATABASE__URL" }}
+              {{- fail "statuslist.env.APP_DATABASE__URL is not supported by this chart because it exposes assembled credentials in pod metadata; use APP_DATABASE__HOST/PORT/USERNAME/NAME/QUERY plus the chart-managed APP_DATABASE__PASSWORD SecretKeyRef" }}
+            {{- end }}
+            {{- if hasKey $env "APP_DATABASE__PASSWORD" }}
+              {{- fail "statuslist.env.APP_DATABASE__PASSWORD must not be set as a plain env value; provide postgres.auth.existingSecret or externalSecret data for the postgres-password key" }}
+            {{- end }}
+            {{- if or (not (hasKey $env "APP_DATABASE__PORT")) (eq (trim (toString (get $env "APP_DATABASE__PORT"))) "") }}
+              {{- fail "statuslist.env.APP_DATABASE__PORT is required; set it to the database service port" }}
+            {{- end }}
```
martcpp, line 101: add APP_DATABASE__PORT: "5432" to values.yaml (or fix both READMEs).

helm/chart/templates/deployment.yaml
martcpp, line 1: add one dbPort helper wired into network-policy.yaml and the two initContainers, plus a test tying egress port to the rendered env

martcpp (changes requested):

martcpp (commented):
Enable TLS on the subchart and set APP_DATABASE__QUERY: "sslmode=require" in values.yaml, or
Add an explicit paragraph to helm/README.md: in-cluster database traffic is unencrypted by default, relies on CNI/mesh encryption, and APP_DATABASE__QUERY is the knob to change that for external databases.

Ngha-Boris:
@martcpp please recheck this

Hermann-Core (dismissed):
LGTM

tests/helm_sensitive_env.rs
```diff
@@ -0,0 +1,236 @@
...
+        !rendered.contains("postgres://"),
+        "rendered Helm output must not contain an assembled Postgres URL"
+    );
+    assert!(
+        !rendered.contains("$(POSTGRES_PASSWORD)"),
+        "rendered Helm output must not assemble credentials with env expansion"
+    );
+    assert!(
+        !rendered.contains("APP_REDIS__URI") && !rendered.contains("rediss://"),
+        "rendered Helm output must not contain Redis credential URI metadata after Redis chart removal"
```
Christiantyemele: Important (stale scope/description): this assertion's rationale ("after Redis chart removal") and the whole PR title/body describe Redis split-credential work (`RedisConfig` split fields, `APP_REDIS__URI`, `docs/REDIS_TLS_SETUP.md`) that is not present in this diff — Redis was already removed upstream before the merge base, and `src/config.rs` has no `RedisConfig` anymore. The net change here is exclusively database-split-credentials + chart hardening. Please update the PR title/description to reflect the actual scope (and drop this stale comment) so reviewers aren't misled about what the PR does.

src/config.rs
```diff
@@ -565,13 +565,246 @@ pub struct DatabasePoolConfig {
...
+                .is_some_and(|value| !value.trim().is_empty())
+            || self
+                .password
+                .as_ref()
+                .is_some_and(|value| !value.expose_secret().is_empty())
+            || self
+                .name
+                .as_deref()
+                .is_some_and(|value| !value.trim().is_empty())
+            || self.port.is_some()
```
Christiantyemele: Suggestion: folding `port` into `has_split_connection_fields()` (this `|| self.port.is_some()`) means any operator who already runs a working `APP_DATABASE__URL` and later adds only `APP_DATABASE__PORT` (e.g. from a shared env file) now hits the hard 'Ambiguous database configuration' failure at startup, where previously `port` did not exist. Consider deciding split-vs-URL on the presence of a *credential-bearing* split field (host/username/password/name) and ignoring `port`/`query` alone for that decision, or at least documenting this sharp edge in the config docs.

helm/chart/templates/_helpers.tpl
```diff
@@ -61,3 +61,11 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Database port helper: returns the configured database port from env or default
+*/}}
+{{- define "status-list-server-chart.dbPort" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- get $env "APP_DATABASE__PORT" | default "5432" }}
```
Christiantyemele: Nit: the `default "5432"` here is unreachable in any valid render, because `deployment.yaml` (line ~87) hard-fails when `APP_DATABASE__PORT` is absent from `statuslist.env`. The two defaults (helper fallback and required-env guard) are a second source of truth that can silently drift for `network-policy.yaml` and the initContainers. Since the app always consumes the same env key, they agree today, but consider dropping the fallback (let it inherit the guard contract) or centralizing the default in one place.

Christiantyemele (changes requested):
Overall this is solid, well-tested security work: moving DB connection-string assembly into the process (split `APP_DATABASE__*` fields), redaction of connection errors, URL-encoding of credentials, fail-closed validation, and the Helm guards that reject assembled URLs/plain passwords are all correct, and `cargo check`/`clippy -D warnings`/the config+helm tests all pass. The earlier review threads (IPv6/trailing-dot host, password trim, `APP_DATABASE_PORT` deploy validation, dead `POSTGRES_USER`, dandydeveloper repo-adds, `.kube-linter.yaml` exclusion) are resolved; the two prior open items (hardcoded `postgres-password` secret key regardless of backend, and the missing `sqlite::memory:` default/migration guard for non-postgres builds) remain open and are not re-filed here. My findings: 1 Important (the PR title and description describe Redis split-credential work that this diff does not contain — Redis was already removed upstream, and the test still carries a stale 'after Redis chart removal' comment), plus 1 Suggestion (adding only `database.port` alongside an existing `database.url` now hard-fails as 'Ambiguous') and 1 Nit (dead `dbPort` default vs the required-env guard). Requesting changes to tighten the PR scope/description and address the Suggestion.

Christiantyemele (changes requested):
LGTM

Christiantyemele (dismissed):

martcpp (dismissed):

Christiantyemele (approved):

martcpp (approved):

PR #444, opened 2026-08-25
fix(deploy): remove --create-namespace from helm deploy step

### LINKED ISSUE #443: Deploy fails: `--create-namespace` causes Forbidden on server-side apply of Namespace
## Summary

The production deploy workflow `.github/workflows/deploy.yml` fails with a `Forbidden` error on the Namespace object:

```
Error: server-side apply failed for object /statuslist-production /v1, Kind=Namespace: namespaces "statuslist-production" is forbidden: User "arn:aws:sts::982081049921:assumed-role/github-actions-production-deploy/status-list-server-deploy" cannot patch resource "namespaces" in API group "" in the namespace "statuslist-production"
```

## Root Cause

The deploy workflow runs `helm upgrade --install` with **`--create-namespace`**. Even though the `statuslist-production` namespace already exists (created by Terraform, `Active`), Helm's `--create-namespace` together with `--atomic` issues a **server-side apply / patch on the Namespace object**. A Namespace is a **cluster-scoped** resource, but the deploy role uses the namespace-scoped `AmazonEKSAdminPolicy`, which does **not** grant `patch` on `namespaces`. Hence `Forbidden`.

This confirms the coordination note from review: once Terraform owns the namespace and it already exists, `--create-namespace` is no longer needed (and is actively harmful under the namespace-scoped OIDC role).

## Fix

In `.github/workflows/deploy.yml`, remove the `--create-namespace \` line from the deploy step:

```yaml
- name: Deploy chart
  run: |
    helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
      --namespace ${{ env.HELM_NAMESPACE }} \
      --atomic \
      --wait \
      --timeout 10m \
      --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
      --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
```

The namespace already exists (`Active`), so Helm will use the existing namespace without trying to create/patch it. The namespace-scoped admin role plus the `statuslist-deploy` RoleBinding (which covers `external-secrets.io` CRDs) is then sufficient for the deployment.

## Caveat / Follow-up Checks

- Confirm the `statuslist-production` namespace, Role, and RoleBinding are actually present (the access entry and namespace were confirmed `Active` earlier). If the RoleBinding was not applied, the external-secrets CRDs would fail next.
- The `--atomic` flag also uses server-side apply; if further `Forbidden` errors appear on other resources after this fix, review whether the deploy role covers those resources.

## Acceptance Criteria

- [ ] `--create-namespace` removed from the deploy step in `.github/workflows/deploy.yml`
- [ ] Deploy to `statuslist-production` succeeds via OIDC role without `Forbidden` on the Namespace

Christiantyemele:
Closes #443

## Summary

Fix the deploy workflow so the OIDC deploy to `statuslist-production` succeeds, addressing two issues in the `helm upgrade --install` command in `.github/workflows/deploy.yml`.

## 1. Remove `--create-namespace`

The deploy fails with `Forbidden` when Helm tries to server-side apply/patch the `statuslist-production` Namespace:

```
Error: server-side apply failed for object /statuslist-production /v1, Kind=Namespace: namespaces "statuslist-production" is forbidden: User "...github-actions-production-deploy..." cannot patch resource "namespaces" ...
```

A Namespace is a cluster-scoped resource, but the deploy role uses the namespace-scoped `AmazonEKSAdminPolicy`, which does not grant `patch` on `namespaces`. The namespace is created and owned by Terraform and already exists (`Active`), so `--create-namespace` is unnecessary.

## 2. Use `--rollback-on-failure` and pin Helm 4

Helm 4 deprecates `--atomic` in favor of `--rollback-on-failure`. Because `--rollback-on-failure` is a Helm 4-only flag, the `azure/setup-helm` step now pins `version: v4.2.4` explicitly (previously it resolved `latest` and could fall back to Helm 3.18.4, where `--rollback-on-failure` is not a valid flag). This makes the Helm version deterministic and lets us use the non-deprecated flag. Rollback-on-failure behavior is preserved via `--rollback-on-failure --wait`.

## Change

Deploy step now:

```yaml
- name: Install Helm
  uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5.0.1
  with:
    version: v4.2.4
...
helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
  --namespace ${{ env.HELM_NAMESPACE }} \
  --wait \
  --rollback-on-failure \
  --timeout 10m \
  --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
  --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
```

## Validation

- [ ] Deploy to `statuslist-production` succeeds via OIDC role without `Forbidden` on the Namespace.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -98,8 +98,8 @@ jobs:
         run: |
           helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
             --namespace ${{ env.HELM_NAMESPACE }} \
-            --atomic \
             --wait \
+            --rollback-on-failure \
             --timeout 10m \
             --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
             --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -99,7 +99,7 @@ jobs:
           helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
             --namespace ${{ env.HELM_NAMESPACE }} \
             --wait \
-            --rollback-on-failure \
+            --atomic \
             --timeout 10m \
             --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
             --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -98,7 +98,6 @@ jobs:
         run: |
           helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
             --namespace ${{ env.HELM_NAMESPACE }} \
-            --create-namespace \
             --atomic \
             --wait \
             --timeout 10m \
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -81,6 +81,8 @@ jobs:
         uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5.0.1
       - name: Install Helm
         uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5.0.1
+        with:
+          version: v4.2.4
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
         with:
@@ -99,7 +101,7 @@ jobs:
           helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
             --namespace ${{ env.HELM_NAMESPACE }} \
             --wait \
-            --atomic \
+            --rollback-on-failure \
             --timeout 10m \
             --set-string statuslist.image.repository=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }} \
             --set-string statuslist.image.tag=${{ needs.build-and-push.outputs.deploy_tag }}


Awambeng (commented):
Could you share reference to the terraform PR?

Christiantyemele:
> Could you share reference to the terraform PR?

https://github.com/ADORSYS-GIS/wallet-eks-env/pull/32

martcpp (dismissed):

.github/workflows/deploy.yml
```diff
@@ -98,9 +98,8 @@ jobs:
         run: |
           helm upgrade --install ${{ env.HELM_RELEASE_NAME }} helm/chart \
             --namespace ${{ env.HELM_NAMESPACE }} \
-            --create-namespace \
-            --atomic \
             --wait \
+            --rollback-on-failure \
```
forkimenjeckayang: `--rollback-on-failure` is a Helm 4 flag: the [Helm 4 documentation](https://docs.helm.sh/docs/overview/#cli-flags-renamed) records it as the new name for `--atomic`, while the [Helm v3.18.4 upgrade source](https://github.com/helm/helm/blob/v3.18.4/cmd/helm/upgrade.go#L253-L256) registers only `--atomic`. This workflow does not pin Helm. The [exact setup action source used here](https://github.com/Azure/setup-helm/blob/9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310/src/run.ts#L12-L56) normally resolves `latest` but falls back to Helm `v3.18.4` if that lookup fails. The normal path currently installs Helm 4, but the fallback would make this deployment fail with an unknown-flag error. Please either retain `--atomic`, which remains compatible with Helm 3 and Helm 4 despite the Helm 4 deprecation warning, or explicitly pin a supported Helm 4 version in the setup step.

forkimenjeckayang (changes requested):
please check @Christiantyemele

.github/workflows/deploy.yml
Christiantyemele: Fixed. Reverted to `--atomic` and retained it in deploy.yml (committed in `5ea5925`). Since this workflow does not pin a Helm version and `azure/setup-helm` can resolve Helm 3 (where `--rollback-on-failure` is not a registered flag), `--atomic` is kept as it remains compatible with both Helm 3 and Helm 4 — it only emits the deprecation warning on Helm 4. Rollback-on-failure behavior is preserved via `--atomic --wait`. I also updated the PR description to reflect this.

Christiantyemele: Update on the resolution — we decided to go with your second suggested option: pin Helm 4 explicitly and keep the non-deprecated flag. The `azure/setup-helm` step now sets `version: v4.2.4` (committed in `ff8b1aa`), making the Helm version deterministic so it never falls back to Helm 3.18.4, and `--rollback-on-failure` remains in place. Confirmed current latest Helm is v4.2.4 (Helm 4). Thanks for catching the non-determinism.

forkimenjeckayang (approved):
Re-reviewed at `ff8b1aa`, the setup step now pins Helm `v4.2.4`, matching `--rollback-on-failure` and removing the Helm 3 fallback path.
LGTM

Hermann-Core (approved):

.github/workflows/deploy.yml
Awambeng: Could you please share a reference confirming that Helm `v4.2.4` supports `--rollback-on-failure`?

Christiantyemele: > Could you please share a reference confirming that Helm `v4.2.4` supports `--rollback-on-failure`?

https://docs.helm.sh/docs/overview/#cli-flags-renamed

Awambeng (approved):

PR #451, opened 2026-08-27
feat: add webhook based alert notifications for observability alerts

### LINKED ISSUE #446: Add Webhook-Based Alert Notifications for Observability Alerts
## Objective

Provide a mechanism to forward actionable observability alerts to external notification systems through webhooks, such as Discord, Slack, or other compatible services.

## Context

The observability work defines Prometheus alert rules for request latency, error rates, cache hit ratio, DB latency, certificate renewal failures, and token-generation failures. However, defining alerts alone is not sufficient for operational readiness if engineers are not notified when an alert fires.

We should support webhook-based alert notifications so that alert events can be delivered to third-party systems such as Discord, Slack, or other services that expose webhook endpoints.

This should build on the existing Prometheus/Grafana alerting setup rather than introducing a separate alerting mechanism.

## Scope

* [ ] Define a generic webhook notification mechanism for alert events.
* [ ] Configure webhook endpoints through deployment configuration/Helm values.
* [ ] Support sending alerts to external systems such as Discord through webhooks.
* [ ] Ensure sensitive webhook URLs/secrets are not hardcoded or committed to the repository.
* [ ] Define the alert message/payload format, including:

  * Alert name
  * Severity
  * Status (`firing` / `resolved`)
  * Description
  * Affected service/component
  * Timestamp
  * Relevant labels/metadata
  * Link to the relevant Grafana dashboard or runbook when available
* [ ] Document how to configure a webhook notification target.
* [ ] Add an example configuration for a Discord webhook.
* [ ] Test that alerts are delivered when an alert fires and that recovery notifications are delivered when the alert is resolved.
* [ ] Document failure/retry behavior when the webhook endpoint is unavailable.

## Acceptance Criteria

* Alert notifications can be forwarded to an external service through a configurable webhook.
* A Discord webhook can be configured without modifying application code.
* Webhook credentials/URLs are provided through secure configuration mechanisms.
* Notifications are sent for both `firing` and `resolved` alert states.
* Notification payloads contain enough context for an engineer to understand and act on the alert.
* Webhook delivery has been tested with a real or mocked external endpoint.
* Configuration and setup instructions are documented.
* Existing Prometheus/Grafana alerting functionality remains unchanged.

## Non-Goals

* Implementing a custom notification service.
* Implementing Discord-specific business logic.
* Replacing Prometheus or Grafana alerting.
* Supporting every third-party notification provider individually.

## Dependencies

* Existing Prometheus alert rules and Grafana dashboards from the observability/SLO ticket.
* Related: #160 (OpenTelemetry collector)

## Example Use Case

When the certificate renewal failure rate exceeds the configured threshold:

`Prometheus Alert → Alertmanager → Webhook → Discord`

The Discord notification should indicate that certificate renewal is failing, the severity, the affected component, when the alert started, and provide links to the relevant dashboard/runbook.

## Suggested Implementation

Use **Alertmanager webhook notifications** as the integration point rather than implementing webhook calls inside the application. This keeps notification concerns outside the application and makes it possible to add Discord, Slack, or other webhook-compatible destinations without changing application code.

ndefokou:
This PR adds a configurable webhook mechanism to forward Prometheus observability alerts to external notification systems (Discord, Slack, Teams, Mattermost, email, or a generic webhook endpoint). It builds on the existing Prometheus/Grafana alerting setup from the SLO observability work rather than introducing a separate alerting mechanism.

Alert events flow through **Prometheus → Alertmanager → Webhook → Discord** (or any configured channel).

## Changes

- **`observability/prometheus/generate-alertmanager-config.sh`** — generates the Alertmanager receiver configuration at container start from environment variables, so the same stack can notify any platform without platform-specific code in the repo.
- **`docker-compose.yml`** — adds an `alertmanager` service that renders the generated config and forwards fired alerts to the configured channel.
- **`observability/prometheus/prometheus.yml`** — routes fired alerts to the in-stack Alertmanager.
- **`.env.template` / `.env`** — documents and supplies the notification platform + credentials via environment variables (platform-agnostic).
- **Docs** — configuration instructions in `observability/README.md` and live-testing guidance in `observability/LIVE_TESTING.md`.

## Supported Platforms
Set `ALERTMANAGER_PLATFORM` and the matching credential:
- `slack` → `SLACK_WEBHOOK_URL`
- `discord` → `DISCORD_WEBHOOK_URL`
- `teams` → `MS_TEAMS_WEBHOOK_URL`
- `mattermost` → `MATTERMOST_WEBHOOK_URL`
- `email` → `ALERTMANAGER_EMAIL_TO` (+ SMTP settings)
- `webhook` → `ALERTMANAGER_WEBHOOK_URL` (generic JSON)

## Security
- Webhook URLs/credentials are **not** hardcoded or committed. They are supplied through environment variables / a secrets manager.
- Real tokens stay out of version control.

## What's covered
- Notifications are sent for both **firing** and **resolved** (`send_resolved: true`) alert states.
- Payloads inherit the rich Alertmanager context (alert name, severity via `severity` label, status, description/annotations, affected `sli`, timestamps, labels, and Grafana/runbook links via `runbook_url`).
- Delivery to real or mocked endpoints can be tested end-to-end.

## Notes
- Prometheus/Grafana alert rules themselves remain unchanged.
- This does **not** implement a custom notification service, Discord-specific business logic, or replace Prometheus/Grafana alerting.


Hermann-Core (changes requested):
..

observability/prometheus/generate-alertmanager-config.sh
```diff
@@ -0,0 +1,161 @@
...
+  fi
+  cat <<'EOF'
+# Generated from ALERTMANAGER_PLATFORM + credentials at container start.
+# Routing is platform-independent; receivers are platform-specific.
+
+route:
+  group_by: ['alertname', 'sli']
+  group_wait: 30s
+  group_interval: 5m
+  receiver: 'default'
```
Hermann-Core: The always-firing `Watchdog` alert in `alerting.rules.yml` (severity=`none`) and the reference `alertmanager.example.yml` both rely on a dead-man's-switch + `noop` sink: unknown/`none` severities must be absorbed, not routed to a human channel. This static route has only `page`/`warn` routes and a top-level `receiver: 'default'` — so the Watchdog (and any rule without `severity`) falls into the configured human channel (Discord/Slack/webhook/email) and re-notifies every `group_interval`/`repeat_interval`. That both spams the channel and silently drops the absence-based dead-man's-switch detection this SLO setup was designed for. Please add a `severity="none"` → dead-man's-switch route (or a `noop` sink default) here, matching `alertmanager.example.yml`.

observability/prometheus/generate-alertmanager-config.sh
```diff
@@ -0,0 +1,161 @@
...
+        - severity="warn"
+      receiver: 'warn'
+      repeat_interval: 24h
+
+receivers:
+EOF
+} > "$OUT"
+
+{
+  emit_receiver default
```
Hermann-Core: For every platform these three receivers (`default`, `page`, `warn`) emit the identical credential (`DISCORD_WEBHOOK_URL`, `ALERTMANAGER_EMAIL_TO`, `ALERTMANAGER_WEBHOOK_URL`, `SLACK_WEBHOOK_URL`), which I verified by running the generator. The route tree implies three distinct severity-segregated targets, but `page` and `warn` both land in the same channel/inbox, so the two-severity model (page = act now, warn = act this week) provides no operational separation. The header comment even references `ALERTMANAGER_*_TO` per-channel overrides for email that don't exist — email uses the single `ALERTMANAGER_EMAIL_TO` for all three. Either implement real per-severity targets or collapse to one documented channel and drop the misleading three-receiver routing.

observability/prometheus/generate-alertmanager-config.sh
```diff
@@ -0,0 +1,161 @@
...
+  smtp_from: '$_from'
+  smtp_require_tls: true
+
+EOF
+  fi
+  cat <<'EOF'
+# Generated from ALERTMANAGER_PLATFORM + credentials at container start.
+# Routing is platform-independent; receivers are platform-specific.
+
+route:
```
Hermann-Core, line 155: Recommended: the generated config omits the `inhibit_rules` block that `alertmanager.example.yml` has (don't page a `warn` while a `page` for the same `sli`+`service` is firing). The alert rules deliberately create correlated fast-burn `page` + slow-burn `warn` pairs for the same SLI, so without inhibition both will fire into the same channel at once. Consider emitting `inhibit_rules` from the generator so the deployed config carries the same guarantee as the reference.

observability/alertmanager/tests/test-alertmanager-config.sh
```diff
@@ -0,0 +1,162 @@
...
+MOCK_PORT=18080
+AM_PORT=19093
+AM_CONFIG="$TMP/am-delivery.yml"
+AM_API="http://127.0.0.1:$AM_PORT/api/v2/alerts"
+
+# For a fast, focused delivery test we use a minimal webhook config with 1s
+# grouping/interval so the firing and resolved notifications are dispatched
+# immediately (the production generator intentionally batches with
+# group_wait=30s / group_interval=5m, which would make this test minutes long).
+cat > "$AM_CONFIG" <<EOF
```
Hermann-Core: The delivery test writes its own minimal webhook config rather than exercising the output of `generate-alertmanager-config.sh`, so the generated production routing (group_by `sli`, page/warn/default, and especially the Watchdog path) is never covered — which is part of why the Watchdog dead-man's-switch / `noop`-sink gap in the generator went uncaught. Config validity is checked separately via amtool (good), but consider a delivery test that boots Alertmanager with the **actual generated config** so the real routing and severities are asserted. That would have caught the routing gaps flagged on `generate-alertmanager-config.sh`.

observability/alertmanager/tests/test-alertmanager-config.sh
```diff
@@ -0,0 +1,162 @@
...
+}
+
+resolved_payload() {
+  # Both timestamps in the past (startsAt < endsAt) so Alertmanager treats the
+  # alert as resolved and dispatches a "resolved" notification.
+  cat <<JSON
+[
+  {"labels":{"alertname":"$ALERT_NAME","severity":"$SEVERITY","service":"status-list-server","sli":"cert_renewal"},
+   "annotations":{"summary":"certificate renewal is failing","runbook_url":"https://github.com/adorsys/status-list-server/blob/develop/observability/runbooks/cert-renewal.md"},
+   "startsAt":"$(date -u -d '5 minutes ago' +%Y-%m-%dT%H:%M:%SZ)","endsAt":"$(date -u -d '1 minute ago' +%Y-%m-%dT%H:%M:%SZ)"}
```
Hermann-Core: Minor portability: `date -d` is GNU-only and will fail on macOS/BSD (this repo is actively developed on macOS). It's fine on the ubuntu CI runner, but the runbook tells operators to run this test locally. Prefer a `date -d` / `date -v` / python fallback, e.g. `python3 -c 'import datetime,sys; print((datetime.datetime.utcnow()-datetime.timedelta(minutes=5)).strftime("%Y-%m-%dT%H:%M:%SZ"))'`. Also the fixed `sleep 3` waits could flake on a loaded CI runner; polling for the expected notification would be more robust.

Hermann-Core (commented):
@ndefokou please check the comments I left on below.

ndefokou:
@Hermann-Core and @Ngha-Boris please review this for me

Ngha-Boris (changes requested):
.

observability/runbooks/webhook-notifications.md
```diff
@@ -0,0 +1,214 @@
+# Webhook / Discord Alert Testing Guide (ticket #446)
```
Ngha-Boris: **Issue: Incomplete Alert Payload Format Documentation**

The ticket requires documenting the alert payload format including: Alert name, Severity, Status (firing/resolved), Description, Affected service/component, Timestamp, Relevant labels/metadata, and Link to Grafana dashboard/runbook.

Currently, the documentation mentions that payloads "inherit the rich Alertmanager context" but doesn't explicitly document the exact JSON structure.

**Suggestion:** Add a dedicated section showing the exact JSON payload structure that will be POSTed to webhooks, including all required fields.

observability/runbooks/webhook-notifications.md
```diff
@@ -0,0 +1,214 @@
...
+Because the dev stack evaluates rules frequently, within ~1–2 windows (seconds
+to a few minutes) Prometheus flips the alert to **firing** and forwards it to
+Alertmanager.
+
+## 5. See the alert in Prometheus, Grafana, and Discord
+
+### Prometheus (<http://localhost:9092>)
+- **Alerts** tab: the alert turns **red / "firing"** with its labels, severity,
+  summary, and a runbook link.
+- **Graph** tab: query the pushed metric (e.g.
```
Ngha-Boris, line 136: **Issue: Incomplete Failure/Retry Behavior Documentation**

The ticket requires documenting "failure/retry behavior when the webhook endpoint is unavailable."

Currently, there's only a brief mention of "exponential back-off (default: 10 retries, back-off up to ~5 minutes per attempt)" but lacks detail on:
- Exact retry configuration parameters
- How to customize retry behavior
- What happens after max retries are exhausted
- How to monitor failed deliveries

**Suggestion:** Expand this section to include a table of retry parameters, configuration examples, and monitoring guidance.

observability/prometheus/generate-alertmanager-config.sh
```diff
@@ -0,0 +1,212 @@
...
+fi
+
+# Emit a single receiver entry for the given platform. Receives: name, then
+# platform-specific vars already expanded by callers where needed.
+emit_receiver() {
+  _name="$1"
+  case "$PLATFORM" in
+    slack)
+      _url="${SLACK_WEBHOOK_URL:?ALERTMANAGER: SLACK_WEBHOOK_URL required for platform=slack}"
+      cat <<EOF
```
Ngha-Boris, line 58: **Issue: Missing SMTP_FROM Validation**

For the email platform, `ALERTMANAGER_SMTP_FROM` defaults to `alertmanager@localhost` which may cause email delivery issues in production. The script should either require this to be explicitly set or at minimum emit a warning.

**Suggestion:** Consider making SMTP_FROM required:
```bash
_from="${ALERTMANAGER_SMTP_FROM:?ALERTMANAGER: ALERTMANAGER_SMTP_FROM required for platform=email}"
```

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,178 @@
...
+              name: {{ $secretName | quote }}
+              key: dms-webhook-url
+            {{- end }}
+          sendResolved: false
+      {{- end }}
+
+    # Human notification channel. Platform-specific receiver config is rendered
+    # below based on alerting.platform; credentials are always read from the Secret
+    # so no webhook URL is committed to source control.
+    - name: human
```
Ngha-Boris, line 106: **Issue: Potential Runtime Error with DMS Secret Reference**

When `existingSecret` is set but the secret doesn't contain a `dms-webhook-url` key, this will fail at runtime. The condition `or .Values.alerting.dmsWebhookUrl .Values.alerting.existingSecret` doesn't guarantee the key exists.

**Suggestion:** Either:
1. Add a separate `dmsEnabled` flag to explicitly enable DMS
2. Or document that `dms-webhook-url` key is optional in the secret and the behavior when missing

observability/alertmanager/tests/test-alertmanager-config.sh
```diff
@@ -0,0 +1,222 @@
...
+HUMAN_PORT=18080
+DMS_PORT=18081
+AM_PORT=19093
+AM_CONFIG="$TMP/am-delivery.yml"
+AM_API="http://127.0.0.1:$AM_PORT/api/v2/alerts"
+HUMAN_LOG="$TMP/human.jsonl"
+DMS_LOG="$TMP/dms.jsonl"
+
+env ALERTMANAGER_PLATFORM=webhook \
+  ALERTMANAGER_WEBHOOK_URL="http://127.0.0.1:$HUMAN_PORT/status-list" \
```
Ngha-Boris, line 146: **Issue: Non-Portable Test (Linux-Only)**

The use of `--network host` is Linux-specific and won't work on macOS or Windows, making this test non-portable for developers on other platforms.

**Suggestion:** Either:
1. Use a Docker network instead for cross-platform compatibility
2. Or add a comment/note that this test is intended for Linux/CI environments only

observability/README.md
```diff
@@ -91,3 +91,31 @@ environments running `kube-prometheus-stack`, the Helm chart includes optional
 `templates/servicemonitor.yaml` (`serviceMonitor.enabled: true`) and
```
Ngha-Boris, line 91: **Issue: Missing Grafana Dashboard Link in Alert Payload**

The ticket requires "Link to the relevant Grafana dashboard or runbook when available." The implementation includes `runbook_url` but doesn't mention Grafana dashboard links.

**Suggestion:** Ensure alerting rules include a `dashboard_url` annotation pointing to the provisioned Grafana dashboard, and document this in the payload format section.

Ngha-Boris (changes requested):
Hello @ndefokou please check the comments below.

Ngha-Boris (dismissed):
.

observability/prometheus/generate-alertmanager-config.sh
Hermann-Core: While `generate-alertmanager-config.sh` now enforces `ALERTMANAGER_SMTP_FROM` as required, this validation is currently bypassed in the deployment manifests:

1. **`docker-compose.yml` (line 191)**: Supplies `ALERTMANAGER_SMTP_FROM: ${ALERTMANAGER_SMTP_FROM:-alertmanager@localhost}`, which defaults back to `alertmanager@localhost` whenever the environment variable is not explicitly set by the operator.
2. **`helm/chart/values.yaml` (line 353)**: Still defaults `smtpFrom: "alertmanager@localhost"`.
3. **`helm/chart/templates/alerting-secret.yaml` (line 41)**: Still defaults `smtp-from: {{ .Values.alerting.email.smtpFrom | default "alertmanager@localhost" | quote }}`, without validating `smtpFrom` as required (unlike `email.to` and `email.smtpHost` in lines 11–18).

To prevent email deliverability failures in production, the `alertmanager@localhost` fallback should be removed from `docker-compose.yml`, `values.yaml`, and `alerting-secret.yaml`, and `alerting.email.smtpFrom` should be validated as required when `alerting.platform=email`.

observability/alertmanager/tests/test-alertmanager-config.sh
Hermann-Core: There are two outstanding issues here:

1. **Process leak in `cleanup()`**: Lines 28–35 define `cleanup()` which checks `MOCK_PID` (which is initialized to empty and never assigned). The script actually spawns `HUMAN_PID` (line 153) and `DMS_PID` (line 155). If Alertmanager fails to start (line 167) or an alert assertion fails, `cleanup()` does not terminate `HUMAN_PID` or `DMS_PID`. Both Python mock servers remain running in the background listening on ports 18080 and 18081, blocking subsequent runs with `Address already in use`. `cleanup()` should terminate and wait both `HUMAN_PID` and `DMS_PID`:
   ```bash
   [ -n "${HUMAN_PID:-}" ] && kill "$HUMAN_PID" 2>/dev/null || true
   [ -n "${DMS_PID:-}" ] && kill "$DMS_PID" 2>/dev/null || true
   [ -n "${HUMAN_PID:-}" ] && wait "$HUMAN_PID" 2>/dev/null || true
   [ -n "${DMS_PID:-}" ] && wait "$DMS_PID" 2>/dev/null || true
   ```
2. **Documentation discrepancy**: While a portability note was added to the script header, `observability/runbooks/webhook-notifications.md` (lines 173–181) still directs operators and developers to run `observability/alertmanager/tests/test-alertmanager-config.sh` locally without mentioning that `--network host` prevents step 2 from running on macOS or Windows. The runbook should explicitly state this Linux/CI requirement, or the test harness should use a Docker bridge network / `host.docker.internal` so it can run cross-platform.

observability/README.md
Hermann-Core: Hardcoding `dashboard_url: "http://localhost:3000/d/status-list-slo"` across all alert rules in `helm/chart/templates/prometheusrule.yaml` causes production Kubernetes alerts to link to `localhost:3000`, which is unreachable for on-call engineers clicking alert notifications in Slack, Discord, or email.

While `http://localhost:3000` is suitable as a default for local Docker Compose testing, in the Helm chart it must be configurable via `values.yaml` (e.g. `.Values.alerting.dashboardUrl` or `.Values.slo.dashboardUrl` with a default), so production Kubernetes deployments can inject their actual Grafana domain or ingress host.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,172 @@
...
+              key: webhook-url
+          sendResolved: true
+      {{- else if eq $platform "email" }}
+      emailConfigs:
+        - to:
+            secretKeyRef:
+              name: {{ $secretName | quote }}
+              key: email-to
+          sendResolved: true
+          smarthost:
```
Hermann-Core: In Prometheus Operator's `AlertmanagerConfig` CRD, `emailConfigs[].smarthost` expects `host:port` (e.g. `smtp.example.com:587`).

Currently:
1. In `alerting-secret.yaml` (lines 39–40), `smtp-host` is written as only `.Values.alerting.email.smtpHost` without the port, while `smtp-port` is written into the Secret but never referenced anywhere by `alertmanagerconfig.yaml`.
2. When `smarthost` receives a host without a port, Alertmanager defaults to port 25 (which is blocked by default across major cloud providers like AWS, GCP, and Azure).
3. `helm/README.md` documents `smtp-host` and `smtp-port` as separate keys for `existingSecret`, but `alertmanagerconfig.yaml` only reads `smtp-host`.

In `alerting-secret.yaml`, please combine host and port when generating `smtp-host` (e.g. `printf "%s:%s" .Values.alerting.email.smtpHost (.Values.alerting.email.smtpPort | default "587")`), and clarify in `helm/README.md` that external secrets must supply `host:port` under `smtp-host`.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,172 @@
...
+            {{ `{{ template "discord.default.title" . }}` }}
+          message: |-
+            {{ `{{ template "discord.default.message" . }}` }}
+      {{- else if eq $platform "slack" }}
+      slackConfigs:
+        - apiURL:
+            secretKeyRef:
+              name: {{ $secretName | quote }}
+              key: webhook-url
+          channel: "#alerts"
```
Hermann-Core: Slack Incoming Webhook URLs generated from Slack Apps are bound to a specific channel upon creation and disallow channel overrides unless granted explicit `chat:write.public` permissions.

Hardcoding `channel: "#alerts"` here (and in `generate-alertmanager-config.sh:64`) will cause deliveries to fail or reject overrides on standard Slack webhooks configured for other channels. `channel` is optional in Alertmanager's `slack_configs` and should either be omitted by default or made configurable via `.Values.alerting.slack.channel` / `${SLACK_CHANNEL:-}`.

observability/prometheus/generate-alertmanager-config.sh
```diff
@@ -0,0 +1,212 @@
...
+# (Slack, Discord, Teams, Mattermost, email, or a generic webhook) without any
+# repo changes or platform-specific code in version control.
+#
+# Routing (severity -> human / dead-man's-switch / noop, repeat intervals) is
+# static and platform-independent; only the human-channel `receivers:` block is
+# generated per platform. Alerts are channelled into a SINGLE human receiver:
+# the two severities (page = act now, warn = act this week) still carry the
+# `severity` label in the payload so an operator can filter in the channel, but
+# they share one configured endpoint. This matches the platform-agnostic model
+# where exactly one credential per platform is available; previously the page
```
Hermann-Core: Nit (comment hygiene): Lines 13–15 describe transient commit history of earlier PR drafts ("previously the page and warn routes pointed at two receivers that emitted the SAME credential...").

Per code guidelines, comments should focus on enduring invariants and architectural rationale rather than past commit iterations. Consider replacing with a concise description of the invariant:
```bash
# All human alerts route to a single receiver carrying the severity label in the payload,
# matching the platform credential model where one destination endpoint is configured.
```

.github/workflows/CI.yml
```diff
@@ -851,6 +851,14 @@ jobs:
             -v "$PWD/observability/prometheus:/etc/prometheus:ro" \
             prom/prometheus:v3.11.3 test rules /etc/prometheus/tests/alerting.test.yml
 
+      - name: Validate Alertmanager config generation + webhook delivery
```
Hermann-Core, line 892: Consider adding fail-closed template validation for `alerting.enabled=true` in `.github/workflows/render-helm-templates/action.yml`:

The chart repository enforces strict fail-closed template testing in `render-helm-templates` for all other features (SecretStore providers, ESO, HPA, PDB, NetworkPolicy). Adding quick checks there for `alerting` will ensure regressions are caught in CI:
1. `alerting.enabled=true` with unsupported `platform` fails.
2. `alerting.enabled=true` with missing `webhookUrl` fails.
3. `alerting.enabled=true` with `platform=email` and missing `email.to` or `smtpHost` fails.
4. `alerting.enabled=true` renders the expected `AlertmanagerConfig` and Secret.

Hermann-Core (changes requested):
@ndefokou Please check the inline comments and follow-ups posted for your attention. I have re-opened some conversations, please check them too.

ndefokou:
> @ndefokou Please check the inline comments and follow-ups posted for your attention. I have re-opened some conversations, please check them too.

@Hermann-Core Please review it again

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,174 @@
...
+# configured notification channel. The route tree mirrors the Docker Compose
+# generate-alertmanager-config.sh logic:
+#   severity=none  -> deadmansswitch (Watchdog; pages on ABSENCE, not on firing)
+#   severity=page  -> human channel, repeat every 4h
+#   severity=warn  -> human channel, repeat every 24h
+#   everything else -> noop sink (never spams a human channel)
+#
+# Credentials are read from the Secret rendered by alerting-secret.yaml (or from
+# alerting.existingSecret when operators manage the Secret externally, e.g. via ESO).
+apiVersion: monitoring.coreos.com/v1alpha1
```
Ngha-Boris: **Issue: Deprecated API Version**

The `v1alpha1` API version for AlertmanagerConfig is deprecated in newer versions of the Prometheus Operator. This may cause compatibility issues with updated clusters.

**Suggestion:** Update to `v1beta1` which is the stable API version:
```yaml
apiVersion: monitoring.coreos.com/v1beta1
```

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,174 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "status-list-server-chart.labels" . | nindent 4 }}
+    app.kubernetes.io/component: alerting
+    {{- with .Values.alerting.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  route:
+    groupBy:
```
Ngha-Boris, line 32: **Issue: GroupBy Missing Service Label**

The grouping only includes `alertname` and `sli`, but the ticket mentions alerts should include "Affected service/component". Without `service` in the groupBy, alerts from different services could be grouped together incorrectly if they share the same alert name.

**Suggestion:** Add `service` to the groupBy labels:
```yaml
groupBy:
  - alertname
  - sli
  - service
```

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,174 @@
...
+      # Mattermost incoming webhook accepts a generic JSON POST.
+      webhookConfigs:
+        - url:
+            secretKeyRef:
+              name: {{ $secretName | quote }}
+              key: webhook-url
+          sendResolved: true
+      {{- else if eq $platform "webhook" }}
+      # Generic endpoint accepting Alertmanager's standard JSON notification payload.
+      webhookConfigs:
```
Ngha-Boris, line 155: **Issue: No Webhook Timeout Configuration**

No timeout is specified for webhook receivers. Alertmanager defaults may not be appropriate for all environments, potentially causing hanging requests during webhook endpoint issues.

**Suggestion:** Add configurable timeout values:
```yaml
webhookConfigs:
  - url:
      secretKeyRef:
        name: {{ $secretName | quote }}
        key: webhook-url
    sendResolved: true
    httpConfig:
      timeout: {{ .Values.alerting.webhookTimeout | default "10s" | quote }}
```

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,174 @@
...
+              key: webhook-url
+          {{- if and .Values.alerting.slack .Values.alerting.slack.channel }}
+          channel: {{ .Values.alerting.slack.channel | quote }}
+          {{- end }}
+          sendResolved: true
+          title: |-
+            {{ `{{ template "slack.default.title" . }}` }}
+          text: |-
+            {{ `{{ template "slack.default.text" . }}` }}
+      {{- else if eq $platform "teams" }}
```
Ngha-Boris, line 130: **Suggestion: Teams Webhook Format Documentation**

Microsoft Teams typically expects an Adaptive Card format, but the implementation uses Alertmanager's generic JSON payload. While some Teams connectors accept JSON, this may not render nicely in Teams channels compared to Discord/Slack native formatting.

**Suggestion:** Document this limitation clearly in the README or consider adding a note that Teams users may need an intermediary service to transform Alertmanager JSON to Adaptive Cards.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,174 @@
...
+        - service
+
+  receivers:
+    # Absorbs anything not explicitly routed.
+    - name: noop
+
+    # Dead-man's-switch for the always-firing Watchdog. When dmsWebhookUrl / dmsEnabled is not
+    # configured, this is an empty receiver and the Watchdog is silently absorbed.
+    # When using existingSecret, set dmsEnabled: true if the secret contains key "dms-webhook-url".
+    - name: deadmansswitch
```
Ngha-Boris, line 90: **Suggestion: Dead Man's Switch Validation Gap**

When using `existingSecret`, users must set `dmsEnabled: true` to enable the DMS receiver, but there's no validation that the `dms-webhook-url` key actually exists in the Secret. If the key is missing, the deployment will fail at runtime.

**Suggestion:** Add validation that checks if `dms-webhook-url` exists when `dmsEnabled: true`, or document this requirement clearly in the values.yaml comments.

helm/chart/templates/alerting-secret.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if and .Values.alerting.enabled (empty .Values.alerting.existingSecret) }}
```
Ngha-Boris, line 2: **Suggestion: Missing Secret Existence Validation**

When using `existingSecret`, there's no validation that the Secret actually exists before the AlertmanagerConfig is created. This could lead to runtime errors.

**Suggestion:** Add a Helm pre-install hook or document that operators must create the Secret before deploying. Alternatively, add a validation step in the NOTES.txt.

observability/prometheus/rules/alerting.rules.yml
```diff
@@ -24,6 +24,7 @@ groups:
         annotations:
           summary: "Prometheus alerting pipeline is healthy (Watchdog)"
           runbook_url: "https://github.com/adorsys/status-list-server/blob/develop/observability/README.md"
+          dashboard_url: "http://localhost:3000/d/status-list-slo"
```
Ngha-Boris, line 27: **Issue: Hardcoded Dashboard URLs**

The dashboard URL is hardcoded to `http://localhost:3000/d/status-list-slo` in the alert rules, but the Helm values allow configuring `alerting.dashboardUrl`. These two can become out of sync.

**Suggestion:** The PrometheusRule template (`helm/chart/templates/prometheusrule.yaml`) should use the configured `alerting.dashboardUrl` value to populate the `dashboard_url` annotation dynamically.

observability/runbooks/webhook-notifications.md
```diff
@@ -0,0 +1,338 @@
...
+| `alerts[].annotations.dashboard_url` | string            | Direct link to the Grafana dashboard (`http://localhost:3000/d/status-list-slo`) for real-time telemetry.                                                |
+| `alerts[].startsAt`                  | string (ISO-8601) | Timestamp when the alert rule first breached its threshold.                                                                                              |
+| `alerts[].endsAt`                    | string (ISO-8601) | Timestamp when the alert resolved (`0001-01-01T00:00:00Z` while firing).                                                                                 |
+| `alerts[].generatorURL`              | string            | Permalink back to the Prometheus graph interface for rule evaluation details.                                                                            |
+| `alerts[].fingerprint`               | string            | Unique cryptographic hash identifying the specific alert instance.                                                                                       |
+
+---
+
```
Ngha-Boris: **Suggestion: Missing Secret Rotation Documentation**

There's no guidance on how to rotate webhook URLs without downtime. When using `existingSecret`, updating the webhook URL requires updating the Secret, but it's unclear if Alertmanager automatically picks up changes.

**Suggestion:** Add a section documenting:
1. How to update webhook URLs (requires Alertmanager reload/restart)
2. Whether the chart supports hot-reloading of Secrets
3. Best practices for credential rotation

observability/alertmanager/tests/test-alertmanager-config.sh
```diff
@@ -0,0 +1,236 @@
...
+#      amtool-valid alertmanager.yml for EVERY supported platform
+#      (webhook|discord|slack|teams|mattermost|email).
+#   2. DELIVERY — a real Alertmanager, configured for the generic `webhook`
+#      platform with `send_resolved: true`, actually POSTs an alert event to a
+#      mock receiver for BOTH the firing and the resolved state.
+#
+# Requires: docker (for the prom/alertmanager image), python3.
+#
+# Portability note:
+#   Step [2/2] uses `--network host` so the containerized Alertmanager can POST
```
Ngha-Boris, line 15: **Suggestion: Test Script Portability**

The script uses `--network host` which only works on Linux. While documented in the runbook, developers on macOS/Windows cannot run the test locally.

**Suggestion:** Consider adding a fallback mechanism or a separate cross-platform test mode that doesn't require host networking. Alternatively, document this as a known limitation in the CONTRIBUTING.md.

helm/chart/values.yaml
```diff
@@ -294,6 +294,71 @@ prometheusRule:
...
+  webhookUrl: ""
+  # Optional dead-man's-switch URL. The always-firing Watchdog (severity=none) pings this
+  # URL so a collapsed Prometheus/Alertmanager pipeline is detected by its ABSENCE. When
+  # empty the Watchdog is silently absorbed and never reaches a human channel.
+  dmsWebhookUrl: ""
+  # Explicitly enable Dead Man's Switch when using existingSecret (reads key "dms-webhook-url" from existingSecret).
+  # When dmsWebhookUrl is set above, DMS is enabled implicitly.
+  dmsEnabled: false
+  # Email settings (required only when platform=email).
+  email:
```
Ngha-Boris, line 402: **Suggestion: No Retry Configuration Exposed**

The Alertmanager webhook retry behavior is documented in `webhook-notifications.md` but cannot be configured through Helm values. Users cannot customize retry backoff parameters without manually editing the AlertmanagerConfig.

**Suggestion:** Expose retry configuration:
```yaml
alerting:
  # ... existing config ...
  retry:
    minBackoff: "10s"
    maxBackoff: "5m"
    maxRetries: 10
```

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,174 @@
...
+          sendResolved: true
+          smarthost:
+            secretKeyRef:
+              name: {{ $secretName | quote }}
+              key: smtp-host
+          from:
+            secretKeyRef:
+              name: {{ $secretName | quote }}
+              key: smtp-from
+          requireTLS: {{ .Values.alerting.email.requireTls | default true }}
```
Ngha-Boris, line 169: **Suggestion: Email TLS Configuration Limited**

The email configuration only allows setting `requireTLS` but doesn't expose other important SMTP security settings like `tlsConfig` for certificate verification or STARTTLS configuration.

**Suggestion:** Add additional TLS configuration options:
```yaml
requireTLS: {{ .Values.alerting.email.requireTls | default true }}
{{- if .Values.alerting.email.tlsConfig }}
tlsConfig:
  {{- toYaml .Values.alerting.email.tlsConfig | nindent 6 }}
{{- end }}
```

Ngha-Boris (changes requested):
Hello @ndefokou please check the comments

CONTRIBUTING.md
```diff
@@ -224,3 +224,15 @@ Automatic on merge:
...
+### Observability / Alertmanager test suite
+
+The webhook-delivery integration test at
+`observability/alertmanager/tests/test-alertmanager-config.sh` requires Docker.
+Its delivery step (`[2/2]`) uses Docker `--network host` so the containerized
+Alertmanager can reach the mock Python receivers on `127.0.0.1`. Host networking
+is only supported on **Linux** (standard for Linux dev machines and CI runners).
+On macOS or Windows (Docker Desktop / Lima), the delivery step cannot run as-is;
+run the suite inside a Linux container or on a Linux CI pipeline instead. The
+config-generation step (`[1/2]`) is portable across platforms.
```
Hermann-Core, line 237: Remove the extra trailing blank line at the end of the file. The file currently ends with two blank lines (lines 238 and 239), which triggers MD012 (no-multiple-blanks) and fails the CI Markdown Lint check. Keep only a single trailing newline.

observability/runbooks/webhook-notifications.md
```diff
@@ -0,0 +1,369 @@
...
+   does not auto-reload on Secret changes, restart the Alertmanager
+   StatefulSet: `kubectl rollout restart statefulset/alertmanager -n <namespace>`.
+3. Verify delivery with `alerthistory` / Alertmanager logs or a test firing; the
+   webhook target should reflect the new URL.
+
+Best practice: treat webhook URLs strictly as secrets, rotate them only via the
+Secret (never in chart values), and keep an audit/allowlist on the receiving
+side so revoked tokens are rejected promptly during rotation.
+
+
```
Hermann-Core: Remove the extra blank line before the horizontal rule. Lines 236 and 237 contain two consecutive blank lines before '---', which triggers MD012 (no-multiple-blanks) and fails the CI Markdown Lint check. Keep one blank line between the preceding paragraph and '---'.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,239 @@
...
+    - name: noop
+
+    # Dead-man's-switch for the always-firing Watchdog. When dmsWebhookUrl / dmsEnabled is not
+    # configured, this is an empty receiver and the Watchdog is silently absorbed.
+    # When using existingSecret, set dmsEnabled: true if the secret contains key "dms-webhook-url".
+    - name: deadmansswitch
+      {{- if or .Values.alerting.dmsWebhookUrl .Values.alerting.dmsEnabled }}
+      webhookConfigs:
+        - url:
+            secretKeyRef:
```
Hermann-Core: In Prometheus Operator AlertmanagerConfig (monitoring.coreos.com/v1beta1 and v1alpha1), WebhookConfig.url is a string (type: string). When referencing a Kubernetes Secret, the CRD defines urlSecret as a SecretKeySelector expecting name and key directly without a secretKeyRef wrapper:

```yaml
urlSecret:
  name: {{ $secretName | quote }}
  key: dms-webhook-url
```

Nesting secretKeyRef under url fails Kubernetes OpenAPI schema validation on apply because url expects a string value, not an object. Update the deadmansswitch receiver to use urlSecret with name and key directly.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,239 @@
...
+              name: {{ $secretName | quote }}
+              key: dms-webhook-url
+          {{- if .Values.alerting.webhookTimeout }}
+          timeout: {{ .Values.alerting.webhookTimeout | quote }}
+          {{- end }}
+          {{- if or .Values.alerting.retry.minBackoff .Values.alerting.retry.maxBackoff .Values.alerting.retry.maxRetries }}
+          httpConfig:
```
Hermann-Core: Prometheus Operator HTTPConfig (pkg/apis/monitoring/v1beta1/alertmanager_config_types.go) supports only authorization, basicAuth, bearerTokenSecret, oauth2, tlsConfig, proxyUrl, followRedirects, and enableHttp2. Neither the CRD nor upstream Alertmanager supports minBackoff, maxBackoff, or maxRetries under httpConfig (Alertmanager controls retry cadence globally via group_interval and repeat_interval).

Supplying minBackoff, maxBackoff, or maxRetries causes Kubernetes API server validation to reject the AlertmanagerConfig resource with unknown field errors. Remove this httpConfig block.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,239 @@
...
+      {{- end }}
+
+    # Human notification channel. Platform-specific receiver config is rendered
+    # below based on alerting.platform; credentials are always read from the Secret
+    # so no webhook URL is committed to source control.
+    - name: human
+      {{- if eq $platform "discord" }}
+      discordConfigs:
+        - apiURL:
+            secretKeyRef:
```
Hermann-Core: In Prometheus Operator DiscordConfig, apiURL is typed as SecretKeySelector with properties name and key directly. It does not accept a nested secretKeyRef wrapper:

```yaml
apiURL:
  name: {{ $secretName | quote }}
  key: webhook-url
```

Using secretKeyRef under apiURL fails Kubernetes OpenAPI schema validation with an unknown field error on cluster apply. Remove the secretKeyRef wrapper so name and key are direct children of apiURL.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,239 @@
...
+              key: webhook-url
+          sendResolved: true
+          title: |-
+            {{ `{{ template "discord.default.title" . }}` }}
+          message: |-
+            {{ `{{ template "discord.default.message" . }}` }}
+      {{- else if eq $platform "slack" }}
+      slackConfigs:
+        - apiURL:
+            secretKeyRef:
```
Hermann-Core: In Prometheus Operator SlackConfig, apiURL is typed as SecretKeySelector with properties name and key directly. It does not accept a nested secretKeyRef wrapper:

```yaml
apiURL:
  name: {{ $secretName | quote }}
  key: webhook-url
```

Using secretKeyRef under apiURL fails Kubernetes OpenAPI schema validation with an unknown field error on cluster apply. Remove the secretKeyRef wrapper so name and key are direct children of apiURL.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,239 @@
...
+          text: |-
+            {{ `{{ template "slack.default.text" . }}` }}
+      {{- else if eq $platform "teams" }}
+      # MS Teams classic inbound webhook accepts a generic JSON POST (no native
+      # Alertmanager receiver). Connectors that require Adaptive Cards can be
+      # bridged by an intermediary; the standard Alertmanager JSON carries full
+      # alert context.
+      webhookConfigs:
+        - url:
+            secretKeyRef:
```
Hermann-Core: Similar to the deadmansswitch receiver, WebhookConfig.url is a string (type: string). When referencing a Kubernetes Secret, use urlSecret with direct name and key fields:

```yaml
urlSecret:
  name: {{ $secretName | quote }}
  key: webhook-url
```

Apply this fix to the teams, mattermost, and webhook receivers, and remove the invalid httpConfig block (minBackoff/maxBackoff/maxRetries) from all of them.

helm/chart/templates/alertmanagerconfig.yaml
```diff
@@ -0,0 +1,239 @@
...
+            maxBackoff: {{ .Values.alerting.retry.maxBackoff | quote }}
+            {{- end }}
+            {{- if .Values.alerting.retry.maxRetries }}
+            maxRetries: {{ .Values.alerting.retry.maxRetries }}
+            {{- end }}
+          {{- end }}
+          sendResolved: true
+      {{- else if eq $platform "email" }}
+      emailConfigs:
+        - to:
```
Hermann-Core: In the Prometheus Operator EmailConfig CRD, to, smarthost, and from are defined as plain strings (type: string), not SecretKeySelector objects. Passing maps containing secretKeyRef fails Kubernetes OpenAPI schema validation on apply.

Render these fields as string literals from values:

```yaml
to: {{ .Values.alerting.email.to | quote }}
smarthost: {{ printf "%s:%s" .Values.alerting.email.smtpHost (.Values.alerting.email.smtpPort | default "587" | toString) | quote }}
from: {{ .Values.alerting.email.smtpFrom | quote }}
```

For password-protected SMTP relays, EmailConfig provides authPassword (name and key) to reference credentials from a Secret.

helm/chart/values.yaml
```diff
@@ -315,6 +315,109 @@ prometheusRule:
...
+  webhookTimeout: ""
+  # Retry/backoff behavior for webhook-based receivers (global.http_config in the
+  # AlertmanagerConfig). Optional; when empty, Alertmanager built-in defaults are used
+  # (min_backoff 10s, max_backoff 5m, backoff_factor 2, unlimited attempts within group_interval).
+  # Example:
+  #   retry:
+  #     minBackoff: "10s"
+  #     maxBackoff: "5m"
+  #     maxRetries: 10
+  retry:
```
Hermann-Core: Remove alerting.retry (minBackoff, maxBackoff, maxRetries). Alertmanager does not have per-receiver retry backoff parameters; retry timing is determined by group_wait, group_interval, and repeat_interval. Defining these settings in values.yaml and rendering them into the CRD produces invalid manifests rejected by the Kubernetes API server on apply.

helm/chart/values.schema.json
```diff
@@ -406,6 +406,31 @@
         "minAvailable": { "type": ["integer", "string", "null"] },
         "maxUnavailable": { "type": ["integer", "string", "null"] }
       }
+    },
+    "alerting": {
+      "type": "object",
+      "properties": {
+        "enabled": { "type": "boolean" },
+        "platform": {
```
Hermann-Core, line 414: Add definitions for dashboardUrl, slack (object with property channel), webhookTimeout, and email.tlsConfig (object) to helm/chart/values.schema.json under alerting.properties. Keeping values.schema.json synchronized reflects all available chart options and prevents schema validation drifts.

observability/prometheus/generate-alertmanager-config.sh
```diff
@@ -0,0 +1,216 @@
...
+  smtp_require_tls: true
+
+EOF
+  fi
+  cat <<'EOF'
+# Generated from ALERTMANAGER_PLATFORM + credentials at container start.
+# Routing is platform-independent; receivers are platform-specific.
+
+route:
+  group_by: ['alertname', 'sli']
```
Hermann-Core: Add 'service' to group_by in generate-alertmanager-config.sh:

```yaml
route:
  group_by: ['alertname', 'sli', 'service']
```

This matches the groupBy configured in alertmanagerconfig.yaml and prevents alerts with the same alert name from different services from collapsing into a single notification.

observability/alertmanager/alertmanager.example.yml
```diff
@@ -1,96 +1,131 @@
...
+# With send_resolved: true, Discord/Slack also receive the resolved embed.
 
 global:
   resolve_timeout: 5m
-  # PagerDuty integration global key fallback
-  pagerduty_url: "https://events.pagerduty.com/v2/enqueue"
 
 route:
-  group_by: ["alertname", "service"]
+  group_by: ["alertname", "sli"]
```
Hermann-Core: Add "service" to group_by in alertmanager.example.yml:

```yaml
group_by: ["alertname", "sli", "service"]
```

This keeps the static example consistent with alertmanagerconfig.yaml and generate-alertmanager-config.sh.

helm/README.md
```diff
@@ -182,6 +182,93 @@ curl -s https://<your-host>/health/ready
 
 For local development with a local cluster, use [`chart/values-local.yaml`](chart/values-local.yaml) and follow the same `helm upgrade --install` flow.
 
+## Alerting
+
+The chart ships an `AlertmanagerConfig` CRD (`monitoring.coreos.com/v1alpha1`) that wires
```
Hermann-Core: Update the API version reference from monitoring.coreos.com/v1alpha1 to monitoring.coreos.com/v1beta1 to match the updated apiVersion in alertmanagerconfig.yaml.

Hermann-Core (changes requested):
@ndefokou Thank you for addressing the earlier feedback.

I completed a technical audit of the latest commit (`4e2ea123`) and identified several issues that need correction before this can be merged:

1. **CI Failure (Markdown Lint MD012)**: Multiple consecutive blank lines were introduced at `CONTRIBUTING.md:238-239` and `observability/runbooks/webhook-notifications.md:236-237`.
2. **Prometheus Operator CRD Schema Incompatibilities**:
   - `WebhookConfig.url`: In Prometheus Operator (`v1beta1` and `v1alpha1`), `url` is a plain string. When referencing a Secret, the field is `urlSecret` taking `name` and `key` directly (not nested under `url.secretKeyRef`).
   - `DiscordConfig.apiURL` and `SlackConfig.apiURL`: `apiURL` takes `name` and `key` directly without a `secretKeyRef` wrapper.
   - `EmailConfig`: `to`, `smarthost`, and `from` are string fields, not Secret selectors.
   - `httpConfig`: `minBackoff`, `maxBackoff`, and `maxRetries` do not exist in the Prometheus Operator CRD or in Alertmanager receivers. Supplying them causes Kubernetes API server validation rejection.
3. **Values Schema Drift**: Newly added properties (`alerting.dashboardUrl`, `alerting.slack`, `alerting.webhookTimeout`, `alerting.email.tlsConfig`) are missing from `helm/chart/values.schema.json`.
4. **Route Grouping Consistency**: `service` was added to `groupBy` in `alertmanagerconfig.yaml`, but omitted from `generate-alertmanager-config.sh` and `alertmanager.example.yml`.
5. **Documentation**: `helm/README.md` still cites `v1alpha1` instead of `v1beta1`.

Detailed inline comments and remediation snippets are provided below.

ndefokou:
@Ngha-Boris and @Hermann-Core please check again

.github/workflows/render-helm-templates/action.yml
```diff
@@ -518,3 +518,66 @@ runs:
...
+        # Email platform renders combined smtp-host (host:port).
+        helm template status-list-server helm/chart \
+          --set alerting.enabled=true \
+          --set alerting.platform=email \
+          --set alerting.email.to=ops@example.com \
+          --set alerting.email.smtpHost=smtp.example.com \
+          --set-string alerting.email.smtpPort=587 \
+          --set alerting.email.smtpFrom=alerts@example.com \
+          --output-dir /tmp/rendered/alerting-email
+        if ! grep -q 'smtp-host: "smtp.example.com:587"' /tmp/rendered/alerting-email/status-list-server/templates/alerting-secret.yaml; then
```
Hermann-Core: This test assertion is the cause of the CI failure in `KubeLinter Scan` and `Trivy Config Scan`.

Commit `0b804a99` correctly moved `smarthost` into `alertmanagerconfig.yaml` as a string per the CRD schema. Because `smtp-host` was removed from `alerting-secret.yaml`, checking `alerting-secret.yaml` fails:

```bash
# Update the assertion to verify smarthost in alertmanagerconfig.yaml instead:
if ! grep -q 'smarthost: "smtp.example.com:587"' /tmp/rendered/alerting-email/status-list-server/templates/alertmanagerconfig.yaml; then
  echo "ERROR: alerting.platform=email AlertmanagerConfig must format smarthost as host:port"; exit 1
fi
```

If you also want to test Secret rendering for email credentials, add a test case with `--set alerting.email.smtpPassword=secret` and verify `smtp-password: "secret"` in `alerting-secret.yaml`.

helm/chart/templates/alerting-secret.yaml
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  name: {{ include "status-list-server-chart.fullname" . }}-alerting
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "status-list-server-chart.labels" . | nindent 4 }}
+    app.kubernetes.io/component: alerting
+  annotations:
+    helm.sh/resource-policy: keep
+type: Opaque
+stringData:
```
Hermann-Core, line 38: When `platform=email` and neither `smtpPassword` nor `dmsWebhookUrl` is set, this template renders a Secret where `stringData` contains only comments and no actual keys.

To avoid creating an unreferenced Secret with empty `stringData`, consider guarding the Secret template so it only renders when credentials exist:

```yaml
{{- $hasSecretData := or (ne $platform "email") .Values.alerting.dmsWebhookUrl (and (eq $platform "email") .Values.alerting.email.smtpPassword) }}
{{- if and .Values.alerting.enabled (empty .Values.alerting.existingSecret) $hasSecretData }}
```

Hermann-Core (changes requested):
@ndefokou Excellent progress aligning `AlertmanagerConfig` with the Prometheus Operator `v1beta1` CRD schema and resolving the earlier feedback.

There is one remaining test regression that is currently blocking CI:

In commit `0b804a99`, `smarthost` was moved to `alertmanagerconfig.yaml` (as a plain string value per the CRD spec) and removed from `alerting-secret.yaml`. However, the test assertion in `.github/workflows/render-helm-templates/action.yml` still checks for `smtp-host` in `alerting-secret.yaml`, which causes `KubeLinter Scan` and `Trivy Config Scan` to fail in CI.

Additionally, when `platform=email` without `smtpPassword` or `dmsWebhookUrl`, `alerting-secret.yaml` renders an empty Secret with no keys under `stringData`.

Inline comments and suggested fixes are provided below.

helm/README.md
```diff
@@ -182,6 +182,97 @@ curl -s https://<your-host>/health/ready
...
+| `mattermost`        | `webhook-url`                                        | Generic `webhookConfigs`                                                             |
+| `webhook`           | `webhook-url`                                        | Generic `webhookConfigs` (standard Alertmanager JSON payload)                        |
+| `email`             | `smtp-password` (optional, SMTP AUTH)                | Native `emailConfigs` receiver; `to`/`smtpHost`/`smtpFrom` are plain values          |
+
+### Prerequisites
+
+The `AlertmanagerConfig` CRD must be installed in the cluster. It is provided by
+`kube-prometheus-stack` or the standalone Prometheus Operator. The `alerting.labels`
+must match the `alertmanagerConfigSelector` configured on your Alertmanager instance
+(typically `release: kube-prometheus-stack`).
```
Hermann-Core: Could you move this prerequisite under the upstream prerequisites section with hypertext link for tools and mark it as optional, since it is only required for alert configuration?

observability/alertmanager/alertmanager.example.yml
```diff
@@ -1,96 +1,131 @@
-# Example Alertmanager Configuration for status-list-server
+# Example Alertmanager Configuration for status-list-server (webhook + Discord)
+#
+# This is the hand-written example for the webhook notification mechanism
+# (ticket #446). The deployed stack instead GENERATES a similar config from
```
Hermann-Core:
```suggestion
# This is the hand-written example for the webhook notification mechanism. The deployed stack instead generates a similar config from
```

observability/alertmanager/alertmanager.example.yml
```diff
@@ -1,96 +1,131 @@
...
         send_resolved: true
-        title: '[{{ .Status | toUpper }}] {{ .CommonLabels.alertname }}'
-        text: >-
-          *Summary:* {{ .CommonAnnotations.summary }} *Severity:* {{ .CommonLabels.severity }} *Runbook:* {{ .CommonAnnotations.runbook_url }}
+        title: '{{ template "discord.default.title" . }}'
+        message: '{{ template "discord.default.message" . }}'
+
+# ---------------------------------------------------------------------------
+# ALTERNATIVES — swap a receiver above for any of these. Only credentials differ.
+# ---------------------------------------------------------------------------
```
Hermann-Core, line 101:
```suggestion
# ALTERNATIVES: swap a receiver above for any of these. Only credentials differ.
```

observability/prometheus/prometheus.production.yml
```diff
@@ -14,6 +14,17 @@ rule_files:
   - /etc/prometheus/rules/recording.rules.yml
   - /etc/prometheus/rules/alerting.rules.yml
 
+# Forward fired/resolved alerts to Alertmanager for webhook-based
+# notifications (ticket #446). Under kube-prometheus-stack / managed
```
Hermann-Core:
```suggestion
# Forward fired/resolved alerts to Alertmanager for webhook-based
# notifications. Under kube-prometheus-stack / managed
```

observability/prometheus/prometheus.yml
```diff
@@ -11,6 +11,16 @@ rule_files:
   - /etc/prometheus/rules/recording.rules.yml
   - /etc/prometheus/rules/alerting.rules.yml
 
+# Forward fired/resolved alerts to Alertmanager for webhook-based
+# notifications (ticket #446). In the docker-compose dev stack the compose
```
Hermann-Core:
```suggestion
# Forward fired/resolved alerts to Alertmanager for webhook-based
# notifications. In the docker-compose dev stack the compose
```

observability/runbooks/webhook-notifications.md
```diff
@@ -0,0 +1,368 @@
+# Webhook / Discord Alert Testing Guide (ticket #446)
```
Hermann-Core:
```suggestion
# Webhook / Discord Alert Testing Guide
```

observability/runbooks/webhook-notifications.md
```diff
@@ -0,0 +1,368 @@
...
+   ▼
+Discord channel
+```
+
+and confirms you can also see the alert/metrics in **Prometheus** and **Grafana**.
+
+The same setup also works for `slack`, `teams`, `mattermost`, `email`, or a
+generic `webhook` — the only thing that changes is the credential you set.
+
+---
```
Hermann-Core: Remove all these extra separators across the file.

observability/README.md
```diff
@@ -91,3 +91,33 @@ environments running `kube-prometheus-stack`, the Helm chart includes optional
...
 `templates/prometheusrule.yaml` (`prometheusRule.enabled: true`). Alert delivery
 routing is documented in `observability/alertmanager/alertmanager.example.yml`.
+
+## Webhook alert notifications (ticket #446)
```
Hermann-Core:
```suggestion
## Webhook alert notifications
```

.env.template
```diff
@@ -214,3 +214,84 @@ POSTGRES_DB=status-list
...
+
+#====================================================
+# REQUIRED BEFORE `docker compose up`:
+#   1. ALERTMANAGER_PLATFORM + DISCORD_WEBHOOK_URL  (to deliver alerts)
+#   2. GRAFANA_ADMIN_PASSWORD                       (Grafana login)
+# See the Discord test guide: observability/runbooks/webhook-notifications.md
+#====================================================
+
+#====================================================
+# Alertmanager / notifications (REQUIRED for alerts) — platform-agnostic
```
Hermann-Core:
```suggestion
# Alertmanager / notifications (REQUIRED for alerts)
```

Hermann-Core (changes requested):

ndefokou:
@Hermann-Core please review it again

ndefokou:
> @ndefokou Excellent progress aligning `AlertmanagerConfig` with the Prometheus Operator `v1beta1` CRD schema and resolving the earlier feedback.
> 
> There is one remaining test regression that is currently blocking CI:
> 
> In commit `0b804a99`, `smarthost` was moved to `alertmanagerconfig.yaml` (as a plain string value per the CRD spec) and removed from `alerting-secret.yaml`. However, the test assertion in `.github/workflows/render-helm-templates/action.yml` still checks for `smtp-host` in `alerting-secret.yaml`, which causes `KubeLinter Scan` and `Trivy Config Scan` to fail in CI.
> 
> Additionally, when `platform=email` without `smtpPassword` or `dmsWebhookUrl`, `alerting-secret.yaml` renders an empty Secret with no keys under `stringData`.
> 
> Inline comments and suggested fixes are provided below.

@Hermann-Core You can review again

ndefokou:
> @ndefokou Excellent progress aligning `AlertmanagerConfig` with the Prometheus Operator `v1beta1` CRD schema and resolving the earlier feedback.
> 
> There is one remaining test regression that is currently blocking CI:
> 
> In commit `0b804a99`, `smarthost` was moved to `alertmanagerconfig.yaml` (as a plain string value per the CRD spec) and removed from `alerting-secret.yaml`. However, the test assertion in `.github/workflows/render-helm-templates/action.yml` still checks for `smtp-host` in `alerting-secret.yaml`, which causes `KubeLinter Scan` and `Trivy Config Scan` to fail in CI.
> 
> Additionally, when `platform=email` without `smtpPassword` or `dmsWebhookUrl`, `alerting-secret.yaml` renders an empty Secret with no keys under `stringData`.
> 
> Inline comments and suggested fixes are provided below.

@Hermann-Core You can review again

Hermann-Core (approved):
LGTM

Ngha-Boris (approved):

Issue #316, opened 2026-07-31
Make local-ci.sh install tools and mirror GitHub CI commands


Hermann-Core:
## Objective

Make `local-ci.sh` faithfully replicate the GitHub CI quality gates and install missing local tools when needed.

## Current State

`local-ci.sh` currently runs fmt, build, clippy, nextest, and cargo-machete. The remote CI also runs or validates:

- zizmor
- memory-only build
- domain-layer purity grep
- cargo doc with `RUSTDOCFLAGS=-D warnings`
- doc tests
- cargo vet
- cargo deny/audit
- typos
- tombi
- markdownlint
- Helm rendering plus Trivy/KubeLinter scans
- yamlfmt
- OpenTelemetry collector config validation
- cargo llvm-cov

## Deliverables

- [ ] Inventory all commands from `.github/workflows/CI.yml`, `cargo_deny.yml`, and `crate_type.yml`.
- [ ] Add local tool bootstrap for missing CLIs where practical, without requiring root unless unavoidable.
- [ ] Provide fast/default and full/parity modes if the full remote CI is too heavy for every local run.
- [ ] Use the same flags and configs as CI.
- [ ] Document any checks that cannot be replicated locally and why.

## Acceptance Criteria

- [ ] `./local-ci.sh --full` runs the same command set as remote CI where local execution is practical.
- [ ] `./local-ci.sh` remains approachable for day-to-day use.
- [ ] Missing tools produce actionable install/bootstrap messages.
- [ ] The script exits non-zero on the same failures CI would catch.

## Verification

- [ ] Run default local CI mode.
- [ ] Run full local CI mode on a clean environment or documented dev image.

martcpp:
Surfaced in review of #388, and it belongs to this issue rather than to that PR because the fix is a CI-cost decision, not a bug.

## The parity gap

`#388` adds machinery to the `Dockerfile` builder stage that is **only ever exercised by `deploy.yml`**:

- pinned `cargo-auditable` / `rust-audit-info` build args,
- `cargo auditable build` in place of `cargo build`,
- an assertion that the shipped binary carries a decodable, non-empty `.dep-v0` section.

`deploy.yml` runs on pushes to `main`, on release tags, and on manual dispatch. It does not run on pull requests. So a change that breaks the auditable build is green on the PR and fails later, on the release path.

[...]

PR #452, opened 2026-08-27
ci: make the zizmor gate fail closed and gate merges on a single required check

### LINKED ISSUE #417: CI: zizmor runs with '|| true' and cannot fail; 84 findings including one High (spoofable bot check in auto_merge.yml)

### LINKED ISSUE #439: ci related fix
this is to fic all the ci issue and make  it work well

martcpp:
Closes #417.

`CI.yml` ran the repository's own workflow scanner as `zizmor .github/workflows/ || true`.
It could never fail. Measured at the pinned version: **84 findings — 1 High, 47 Medium.**

Fixing that surfaced a larger problem: **no job in `CI.yml` blocked a merge either.**
The only required check was `Conventional Commits`. Removing `|| true` on its own
would have made zizmor able to fail a workflow that still gated nothing.

## What this does

- **Clears the backlog to zero.** `persist-credentials: false` on all 26 checkouts
  (`artipacked`, 25); explicit `permissions:` on every workflow and job
  (`excessive-permissions`, 22).
- **Fixes the High finding.** `auto_merge.yml` gated auto-merge on `github.actor` —
  spoofable, and reflects who last touched the PR rather than who opened it. Now
  `pull_request.user.login` plus a same-repository check.
- **Removes `|| true`.** Pinned scanner version, fail-closed at every severity.
  Accept a finding with `# zizmor: ignore[rule]` on the line it anchors to.
- **Adds `ci-success`** — one aggregating job, `name: CI Success`, needing all 20
  others. A CI step asserts every job appears in `ci-success.needs`, so a new job
  cannot silently escape the gate the way `docker-build` already had.
- **Adds a self-test.** Every run audits a deliberately-vulnerable fixture and
  asserts the gate exits 13 **and** names `artipacked`. Exit status alone is not
  enough — a failed image pull would satisfy a bare non-zero check just as well.
- **Closes two silent-skip paths.** `crate_type.yml` keyed `is_lib` on `targets[0]`
  (cargo does not order them) and wrote it through a command substitution `set -e`
  cannot see, either of which skipped `cargo-test-doc` silently — and `ci-success`
  counts skipped as a pass. `deploy.yml` gains `validate-tag`: `v*.*.*` is a refspec
  glob, not semver, so `v2024.01.release` matches and the deploy would fall through
  to the mutable `latest`.

## Verified

zizmor 1.28.0, offline and online: 0 findings. `ci-success.needs` 20/20. Self-test
confirmed in a real run:

```
warning[artipacked]: credential persistence through GitHub Actions artifacts
self-test passed: the gate fails, and names the finding, on a known-bad workflow.
every CI.yml job is gated by ci-success.
```

## Branch protection — already applied

`CI Success` is now a required status check on `Rules` (develop, alongside
`Conventional Commits`) and on `main branch guards`. Confirmed blocking on this PR.

## Deliberately not done

*Allow auto-merge* stays off. `auto_merge.yml` warns and passes for that specific
case rather than failing — a job that is red on every dependency bump only teaches
people to ignore red. Every other failure to arm still fails the job.

`auto_merge.yml` also used to run `gh pr review --approve`. That failed on every run
since it was added (GitHub refuses `addPullRequestReview` from Actions), which
silently skipped the auto-merge step behind it. Removed rather than unblocked: a bot
approval cannot satisfy a two-review requirement, and CI approving its own dependency
bumps is what the review requirement exists to prevent.

## Known gaps

- `ci-success` is itself untested; nothing proves it fails when a need fails
- the fixture covers `artipacked` only, not `excessive-permissions` or `bot-conditions`
- `local-ci.sh` cannot reproduce the gate — it now needs Docker, see #316
- `ZIZMOR_VERSION` is invisible to Dependabot and has no bump owner
- `online-audits: true` means an API outage or a new advisory reddens PRs with no
  change to this repository
- `develop` keeps `strict` required checks, so every open PR must now rebase and
  re-run all 20 jobs after any merge; the merge queue is the alternative, not decided
- `deploy.yml`'s permission chain is unexercised until a real tag push
- Dependabot `cooldown: 7` is undocumented

## Scope

#417 suggested landing this per-workflow. It is one PR instead: the two Medium
classes are mechanical and repository-wide, and removing `|| true` is only meaningful
once `ci-success` exists. Review by file rather than by commit — a later commit
reverses a decision made in an earlier one.

## Related

- #388 — added ~300 lines of workflow while the scanner could not fail
- #141 — production readiness EPIC
- #316 — `local-ci.sh` parity


.github/workflows/CI.yml
```diff
@@ -19,49 +19,105 @@ jobs:
...
-        # linter that cannot fail: findings on the repository's highest-privilege
-        # workflows were computed and discarded. Tighten the threshold as the
-        # lower-severity backlog is triaged.
+        uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054 # v0.6.2
+        with:
+          version: ${{ env.ZIZMOR_VERSION }}
+          # `.` also picks up .github/dependabot.yml.
+          inputs: .
+          persona: regular
+          online-audits: true
```
ndefokou, line 72: @martcpp This is the highest-consequence operational risk in the PR. The gate is now fail-closed at *every severity* AND `online-audits: true`. That means the merge path depends on (a) zizmor reaching its online advisory endpoints from the runner and (b) no *newly published advisory* matching any existing workflow pattern. An API outage or a new advisory reddens every PR with zero change to this repository — the exact failure mode this PR exists to prevent (`|| true` made the gate never block; this makes it block for reasons unrelated to the diff).

Recommendation: gate merges on an offline/local run at persona `regular` (deterministic, no external dependency), and reserve `online-audits: true` for a scheduled/full scan. If keeping it on the gate, at minimum add retry/`continue-on-error` semantics and accept the tradeoff explicitly in the PR body. See also the self-test note: the fixture runs with `--no-online-audits`, so the online path itself is never exercised by the self-test either.

ndefokou (commented):
## Inline review — CI hardening for the zizmor gate

@martcpp — I've reviewed the full PR (#452). The core work is sound and well-engineered; these are inline notes on the risks and improvements I found, anchored at their source lines. Summary of the two highest-priority items (both on the release path that is, by your own note, currently unexercised):

1. **`online-audits: true` on the merge gate** (`CI.yml:42`) — availability/DoS risk; see inline.
2. **`scan-image` artifact upload under narrowed permissions** (`deploy.yml:325`) — the `actions/upload-artifact` steps need verifying with `actions: write` (or equivalent) or the release scan can go red with a 403 and no `continue-on-error`.

Plus smaller maintainability items (pin `yq`, add negative/unit tests for the aggregator and `validate-tag`, give the hard tool-version pins a bump owner, and consider splitting the comment-rewrite churn into a separate PR).

.github/workflows/CI.yml
```diff
@@ -19,49 +19,105 @@ jobs:
...
+          echo "self-test passed: the gate fails, and names the finding, on a known-bad workflow."
+
+      # `ci-success.needs` is hand-maintained, so a job missing from it fails without
+      # blocking the merge. `docker-build` already did.
+      - name: Assert every job is gated by ci-success
+        if: ${{ !cancelled() }}
+        run: |
+          set -euo pipefail
+
+          yq -r '.jobs | keys | .[] | select(. != "ci-success")' \
```
ndefokou, line 155: @martcpp The gating assertion shells out to `yq`, but `yq` (mikefarah v4) is not guaranteed to be preinstalled on the `ubuntu-latest` runner image. If it's ever absent, this step fails on the first run of every PR. Direction of failure is safe (it fails closed, blocking the merge) — but a false-red that hits every PR is exactly the kind of noise that teaches people to ignore the gate, which is the thing this PR exists to prevent.

Recommendation: install `yq` explicitly in the step (e.g. `mikefarah/yq` via download or use the standard preinstalled `jq` instead), so the step is deterministic and reviewable rather than dependent on unstated runner image contents. The `docker-build`-escaped-the-gate story is precisely why this assertion matters; make sure its tooling can never be the reason it trips idly.

.github/workflows/CI.yml
```diff
@@ -19,49 +19,105 @@ jobs:
   zizmor-security:
     name: Zizmor Security Check
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      security-events: write
+    env:
+      # The action's `version` defaults to `latest`, so without this an upstream zizmor
+      # release turns every PR red. The action's SHA pins the action, not the scanner.
+      ZIZMOR_VERSION: "1.28.0"
```
ndefokou, line 25: @martcpp You list `ZIZMOR_VERSION` being invisible to Dependabot as a known gap — but the warning generalizes: `CI.yml` now hard-pins `nextest@0.9.101`, `cargo-vet@0.10.2`, `typos-cli@1.49.0`, `cargo-audit@0.22.2`, `cargo-machete@0.9.2`, plus every action SHA/version in this file and `cargo_deny.yml`. None of those are visible to Dependabot either.

The intent (reproducibility) is excellent, but a stale pin on a security-critical scanner becomes the next silent blind spot — which is ironic for a PR whose whole purpose is closing silent blind spots. Recommend a documented bump-owner + a scheduled advisory check, or a renovate-style watcher for these `env:`/`tool:` pins, so they don't drift untouched.

.github/workflows/CI.yml
```diff
@@ -553,23 +631,69 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions: {}
+    steps:
+      # Printed so a failure shows which job caused it without expanding the UI.
+      - name: Summarise job results
+        env:
+          NEEDS: ${{ toJSON(needs) }}
+        run: printf '%s\n' "$NEEDS"
+      - name: Fail if any job failed or was cancelled
+        if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
```
ndefokou, line 972: @martcpp The aggregator logic is correct (`if: always()` + `contains(needs.*.result,'failure')||contains(...,'cancelled')`, skipped-as-pass is documented). But `ci-success` has no negative test — nothing in the suite proves it actually *fails* when a needed job fails. You accepted this as a gap; it's cheap to close. A small script that feeds a synthetic `needs`-results object into the same `contains(...)` predicate (positive: all success → pass; negative: one `failure` → fail; `cancelled` → fail; mix with `skipped` → pass) would lock in the semantics.

Worth doing because the whole upstream chain (branch protection requiring only `CI Success`) trusts this one predicate. One line-fragile hand-maintained block is fine, but a one-line flip in a future 'cleanup' of the `||`/`&&` logic would silently turn the gate off with no test catching it.

.github/workflows/CI.yml
```diff
@@ -553,23 +631,69 @@ jobs:
...
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: coverage-report
           path: target/llvm-cov/html
           if-no-files-found: error
+
+  # The single required status check. The jobs above form several independent chains
+  # rather than one, so no individual job represents the suite. Rollout order:
+  # CONTRIBUTING.md, "Branch Protection".
+  ci-success:
```
ndefokou, line 907: @martcpp `ci-success` includes both `check-crate-type` *and* `cargo-test-doc` in `needs`. That's correct today, but note the skip-as-pass interaction: if `check-crate-type` were ever itself skipped (e.g. someone adds an `if:` upstream), its `is_lib` output is empty, `cargo-test-doc` then sees `'' == 'true'` → skips, and **both** report pass — an empty-output skip silently clears the gate. The in-job non-empty assert mitigates the case where the job *runs*; it does not cover the whole-reusable-workflow-being-skipped case. Low likelihood (a contributor would have to modify CI.yml), but worth a one-line hardening comment or an explicit guard on `check-crate-type` that it never silently skips.

.github/workflows/deploy.yml
```diff
@@ -285,11 +319,10 @@ jobs:
...
 
       - name: Upload scan reports
-        # `warn`, not `error`: the artifacts-always-exist guarantee belongs to the
-        # `[ -s ... ]` fallbacks above. Failing here would only add a second red step
-        # when an earlier one already died.
+        # `warn`, not `error`: the `[ -s ... ]` fallbacks above guarantee the files.
+        # Failing here would only add a second red step.
         if: always()
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```
ndefokou, line 412: @martcpp Key concern: `scan-image` declares only `permissions: contents: read, packages: read`, yet this step runs `actions/upload-artifact` (and again at line 339 for `container-sboms`). Artifact storage is scoped under the **`actions`** permission; a token without `actions: write` can throw `Resource not accessible by integration`. Both steps are `if: always()` with *no* `continue-on-error`, so a 403 here makes the release `scan-image` job red and blocks the deploy.

This is exactly the path your own note calls out as unexercised — CI.yml's equivalent (`cargo-coverage` with `if-no-files-found: error`) carries the same risk on the pull_request path. Please verify with a real artifact upload under these narrowed permissions (dry-run a tag push to a scratch repo, or temporarily grant `actions: write`) before this gate becomes the release authority. This is the highest-priority *unverified* risk I found.

.github/workflows/deploy.yml
```diff
@@ -25,43 +26,81 @@ env:
...
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          IS_TAG: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          if [ "$IS_TAG" != "true" ]; then
+            echo "Not a tag push -- nothing to validate."
+            exit 0
+          fi
+
+          semver='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[0-9A-Za-z.-]+)?$'
```
ndefokou: @martcpp `validate-tag` is a genuinely good catch (a refspec glob, not semver — `v2024.01.release` would have fallen through to the mutable `latest`). The regex is right for the intended contract: semver with optional prerelease, and build metadata deliberately rejected because `+` is illegal in Docker tags.

Suggest a quick unit test in the same script-test style you already use (the repo has `scripts/tests/test_*.py`): happy paths (`v1.2.3`, `v1.2.3-rc.1`), and the rejection cases (`v2024.01.release`, `v1.2.3+meta`, `1.2.3`/non-tag `IS_TAG` guard). Trivial and locks in the release-integrity property. Also worth one doc line anywhere release-cutting humans are instructed, so someone hand-pushing `v1.2.3+meta` isn't surprised.

.github/workflows/auto_merge.yml
```diff
@@ -2,55 +2,87 @@ name: Auto Merge Dependabot
...
-    # arbitrary code merges without review. `user.login` is the account that opened
-    # the pull request and cannot be reassigned by later activity.
-    #
-    # Behaviour is unchanged for `workflow_dispatch`: there is no pull request in the
-    # event payload, so this evaluates false exactly as the previous condition did.
-    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    # `user.login`, not `github.actor`: the actor is whoever last touched the pull
+    # request, so it is both spoofable and wrong after a maintainer reopens one.
+    if: >-
+      github.event.pull_request.user.login == 'dependabot[bot]' && github.event.pull_request.head.repo.full_name == github.repository
```
ndefokou, line 19: @martcpp Correct and important fix. `user.login` is the account that *opened* the PR and cannot be reassigned by later activity (unlike the spoofable `github.actor`), and the added `head.repo.full_name == github.repository` same-repo check cleanly kills fork-PR execution where a token-harvesting attack would otherwise be possible. Together these correctly remediate the High `bot-conditions` finding.

One confirmation to double-check: the comment claims `contents: write` is needed because `--auto` merges immediately when already mergeable rather than just arming. That nuance is real, but it means this job runs on the `pull_request` event with write scopes whenever `user.login == dependabot[bot]` — please confirm there's no path where a failed/restricted dependabot branch could exercise those scopes beyond the metadata/merge steps shown. The eligibility gate (`patch`/`minor`, never `github-actions`, fail-closed on empty update-type) is well-designed.

.github/workflows/CI.yml
```diff
@@ -19,49 +19,105 @@ jobs:
...
-
-      - name: Install Zizmor
-        uses: taiki-e/install-action@c44f6b046f1c29ae5918b1e0bfdbb2f1813836fd # v2
         with:
-          tool: zizmor@1.28.0
+          persist-credentials: false
 
+      # Fail-closed at every severity. To accept a finding use `# zizmor: ignore[rule]`
+      # on the line it anchors to; `ignore-start`/`ignore-end` suppress nothing.
       - name: Run Zizmor
```
ndefokou, line 62: @martcpp The self-test is the strongest part of this PR — asserting exit code **13** (zizmor's 'findings reported') rather than bare non-zero, plus grepping for the `artipacked` finding name, correctly closes both the 'image pull failed but non-zero' and 'evaluated but found nothing' gaps. Nice.

Two coverage notes (you flag both): it runs `--no-online-audits` while the real gate uses `online-audits: true`, so the online path is never exercised (ties into the `CI.yml:42` comment); and it covers only `artipacked`, not `excessive-permissions` or `bot-conditions`. A second fixture keyed on one of those rules would harden the gate against a regression where `persist-credentials` hardening is silently dropped *by mistake* rather than by the fixture the gate trusts.

.github/workflows/CI.yml
```diff
@@ -19,49 +19,105 @@ jobs:
...
+          out=$(docker run --rm \
+            --volume "${GITHUB_WORKSPACE}:/workspace:ro" \
+            --workdir /workspace \
+            "$ZIZMOR_IMAGE" \
+            --persona=regular --no-online-audits --color=never --format=plain \
+            -- "$FIXTURE" 2>&1) || rc=$?
+          printf '%s\n' "$out"
+
+          # 13 is zizmor's "findings were reported". 3 is "no inputs collected" and 1
+          # is any other error -- both would pass a bare non-zero check.
```
Christiantyemele: The comment says "13 is zizmor's 'findings were reported'", but per zizmor's documented exit codes 13 specifically means "one or more findings, highest finding is **medium**" (11=info, 12=low, 13=medium, 14=high). The assertion holds only because the fixture's `artipacked` finding is medium — confirmed in the run log (`3 findings ... 1 medium, 0 high`). This is fragile: if a zizmor bump ever reclassifies `artipacked`'s severity, the self-test will spuriously fail (or silently no longer prove what it claims). Suggest rewording the comment to the accurate contract and, if feasible, asserting on the 11–14 range or making the exit-code expectation a named constant next to `ZIZMOR_VERSION` so it is revisited with every pin bump.

.github/workflows/CI.yml
Christiantyemele, line 72: `online-audits: true` combined with `ci-success` now being the single required merge check means an external advisory or API outage can redden **every** PR with no repository change, blocking all merges until someone intervenes. This is a deliberate, documented trade-off (also called out in the PR body's Known Gaps), so not a defect — but the workflow itself is silent about it. Consider a short comment here or in CONTRIBUTING.md noting that a flaky/blocked external advisory DB can gate merges, and that the recovery is a temporary `zizmor: ignore[...]`/config pin. (Same theme as the PR author's own Known Gaps list.)

.github/workflows/auto_merge.yml
```diff
@@ -2,55 +2,87 @@ name: Auto Merge Dependabot
...
+            printf '%s\n' "$out"
+            exit 0
+          fi
+          printf '%s\n' "$out"
+
+          # `Allow auto-merge` being off is a repository setting, not a fault in this
+          # pull request, and it fails every eligible one until it is turned on. A job
+          # that is red on every dependency bump teaches people to ignore red, so this
+          # case warns. Every other failure still fails.
+          if printf '%s' "$out" | grep -qiE 'auto[ -]?merge is not allowed'; then
```
Christiantyemele: The "auto-merge is not allowed" detection greps gh's human-readable error string: `grep -qiE 'auto[ -]?merge is not allowed'`. gh's error wording is not a stable API and can differ across gh versions/locales, so an upstream gh change could silently convert a "settings off" case into either a red job (line 83-84) or, worse, into the warning path when something else failed. Consider a more robust signal — e.g. pre-checking via `gh api repos/{owner}/{repo} --jq '.auto_merge_allowed'`/the mergeability settings. Minor; the current heuristic is a reasonable best-effort.

.github/dependabot.yml
```diff
@@ -27,6 +27,8 @@ updates:
           - "*"
         update-types:
           - "major"
+    cooldown:
+      default-days: 7
```
Christiantyemele, line 31: Nit: the three `cooldown: default-days: 7` blocks are added, but the PR body's own Known Gaps list notes "Dependabot `cooldown: 7` is undocumented" — and CONTRIBUTING.md's new "Dependabot auto-merge" section never mentions the cooldown. Since cooldown now also configures whether the auto-merge path sees fresh versions, a one-line note there (e.g. "Dependabot applies a 7-day release cooldown to version updates; security updates are exempt") would close the loop and prevent future maintainers from wondering why a just-published version isn't proposed yet.

Christiantyemele (commented):
Overall: an exceptionally well-engineered CI/security PR. I verified the crate_type.yml jq logic empirically (`any(IN($lib[]))` correctly performs set membership across all targets), confirmed `cooldown` is a valid, officially-supported Dependabot option, and confirmed from the run logs that the zizmor gate's `collect: default` over `.` picks up only `.github/` files (the deliberately-vulnerable fixture under `scripts/testdata/` is correctly excluded), that the self-test fires `artipacked` on the fixture, and that all 21 jobs are gated through `ci-success`. No Bug or Important findings. Findings below are Suggestions/Nits; not merge-blocking: 4 Suggestions, 0 Important, 0 Bugs.

.github/workflows/auto_merge.yml
martcpp: I dont see any nedd for auto merge for kow that why I left it

.github/workflows/CI.yml
martcpp: Already documented this no nedd to add it everywhere ci is set once and most no body touch it those too many commen just eat up space

.github/workflows/CI.yml
martcpp: After the pr is merge i will turn the check on to merge that

.github/workflows/deploy.yml
martcpp: This is alrey Verifier's and working thank

.github/workflows/auto_merge.yml
martcpp: Already comfirm

.github/workflows/CI.yml
martcpp: The self-test proves the gate evaluates and reports; making it online would put the self-test itself behind the network flakiness comment 1 is about, and the fixture pins its action by SHA so online audits add nothing deterministic. Left as-is deliberately.

.github/workflows/CI.yml
martcpp: The self-test proves the gate evaluates and reports; making it online would put the self-test itself behind the network flakiness comment 1 is about, and the fixture pins its action by SHA so online audits add nothing deterministic. Left as-is deliberately.

martcpp:
@Christiantyemele and @ndefokou pls rechack

.github/workflows/deploy.yml
```diff
@@ -662,14 +676,16 @@ jobs:
       IMAGE_VARIANT: ${{ vars.IMAGE_VARIANT || 'aws' }}
     steps:
       - name: Checkout source code
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           # Only the Helm chart is needed here; the deploy authenticates to AWS via
           # OIDC, not the checkout token.
           persist-credentials: false
```
ndefokou, line 684: @martcpp Still the top unresolved item from my prior review, now matrix-multiplied. `scan-image` declares `permissions: contents: read, packages: read, actions: read`, yet this job uploads artifacts (this step and the `container-sboms` one below at position ~414), both `if: always()` with `if-no-files-found: warn` and NO `continue-on-error`.

GitHub's artifact storage is gated by the `actions` scope, and `actions/upload-artifact` requires **`actions: write`** — `actions: read` is the download/list scope. A 403 on any of the 8 upload steps (4 variants × 2) fails the whole matrix job and blocks the release. This is exactly the path your own known-gap note says is unexercised until a real tag push.

Recommendation: either grant `actions: write` on these three jobs (build-and-push, scan-image, promote-tags) or verify empirically with a real dry-run tag push to a scratch repo that `upload-artifact` succeeds under the declared permissions. Since `if-no-files-found: warn` swallows the *file missing* case but not a permission 403, a 403 is a hard step failure.

.github/workflows/deploy.yml
```diff
@@ -372,11 +406,10 @@ jobs:
...
-        # `[ -s ... ]` fallbacks above. Failing here would only add a second red step
-        # when an earlier one already died.
+        # `warn`, not `error`: the `[ -s ... ]` fallbacks above guarantee the files.
+        # Failing here would only add a second red step.
         if: always()
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: container-scan-reports-${{ matrix.variant.suffix }}
           path: |
```
ndefokou, line 415: @martcpp Corroborating the scan-image comment: this `build-and-push` job runs `actions/upload-artifact` ("Upload built digest") with `permissions: contents: read, packages: write` — no `actions` scope at all, so it is *even less* equipped for artifact upload than `scan-image`. Same failure mode: a 403 here fails `build-and-push`, cascading into `scan-image`/`promote-tags`/`deploy`, which all `needs` it.

This class (artifact upload under a token that can't write `actions`) is the one release-path risk that is still entirely unproven — nothing in the PR exercises a real `actions/upload-artifact@v7` under these narrowed permissions. Please confirm (grant `actions: write` or dry-run) before this gate owns releases.

.github/workflows/CI.yml
```diff
@@ -19,49 +19,166 @@ jobs:
...
+          advanced-security: false
+          annotations: true
+
+      # Only a transient failure is rescued: a real finding is deterministic and fails
+      # this run too, so the retry cannot turn a red gate green. Not `continue-on-error`,
+      # so this step is the one that fails the job.
+      - name: Run Zizmor (retry)
+        if: ${{ steps.zizmor.outcome == 'failure' }}
+        uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054 # v0.6.2
+        with:
```
ndefokou, line 84: @martcpp Good fix for the transient-outage concern — `continue-on-error` + a gating retry is the right shape, and the deterministic-finding argument (it fails both runs) is correct, so the retry cannot whiten a red gate.

Two residual notes (non-blocking):
1. A **sustained** GitHub API outage (longer than the retry window) still reddens every PR — the retry narrows but does not eliminate the availability coupling. That's an acceptable, clearly-documented tradeoff; just make sure it's a conscious one. If it ever becomes noise, the escape hatch is gating merges offline and moving `online-audits: true` to a scheduled scan.
2. On a *real* finding, the retry re-runs the full scan (cost), because `outcome == 'failure'` can't distinguish a finding from a transient error. Minor; could gate the retry on something that excludes deterministic findings, but it's not worth the complexity now.

ndefokou (commented):
## Re-review — follow-up on the review changes (PR #452)

@martcpp — I've re-reviewed the updated branch (head `9287a04`). The substantive changes that address my earlier review are all present and well-executed:

- **`online-audits` transient-outage risk**: mitigated with `continue-on-error` on the primary scan + a gating retry step, and the deliberate-rationale comment is honest about the tradeoff. The retry cannot turn a real finding green (a finding is deterministic and fails both runs) — that's the correct property.
- **Self-test**: now accepts the whole 11–14 findings band (not a fragile single code) and asserts **both** `artipacked` and `excessive-permissions` — exactly closing my "single-rule coverage" point.
- **`ci-success` skip fail-open**: the new "Fail if a required job was skipped" step closes the skip-as-pass gap, including my flagged transitive `check-crate-type` case. Correct.
- **`validate-tag`**: extracted to `scripts/validate-release-tag.sh` with a thorough unit suite (`test_validate_release_tag.py`) covering accept/reject/`+`-metadata/leading-zero/non-tag — exactly the coverage I asked for. Excellent.
- **`auto_merge.yml`**: reads `allow_auto_merge` from the repository object via `gh api` instead of fragile `gh` error-text matching. Strong improvement.
- **`yq`**: documented as part of the runner-image contract with a link; `ci-success.needs` confirmed 23/23 against jobs; all 21 CI checkouts carry `persist-credentials: false`.

The one substantive item from my last review that remains **unverified and now larger** is the artifact-upload permission on the release path — see the two inline comments. Net position: very close to ready; resolve the artifact-permission verification (or add `actions: write` and test a real tag push) before this becomes the release authority.

.github/workflows/CI.yml
martcpp: No `actions:` scope, despite this job running `actions/upload-artifact`.
   `permissions:` scopes GITHUB_TOKEN, but the artifact client does not use  GITHUB_TOKEN: it authenticates to the artifact backend with the separate  ACTIONS_RUNTIME_TOKEN the runner injects into every job (see  actions/toolkit, packages/artifact -- `getRuntimeToken()`). So upload is unaffected by these scopes, and `actions: write` would grant real API reach over workflow runs for nothing. `actions: read` is the download/list scopefor the REST API, which nothing here calls.
    
 Proven on every pull request, not just asserted: CI.yml's `kube-linter` job  uploads an artifact under `contents: read` + `security-events: write`, with  no `actions` scope at all, and that step passes.

martcpp:
@ndefokou recheck thanks

.github/workflows/deploy.yml
martcpp: The claim is "actions/upload-artifact requires actions: write; actions: read is the download/list scope; a 403 fails the job." That's incorrect, on two independent grounds.

Mechanism. permissions: scopes GITHUB_TOKEN. The artifact client doesn't use GITHUB_TOKEN —it authenticates to the artifact backend with ACTIONS_RUNTIME_TOKEN, a separate token the runner injects into every job. In actions/toolkit, packages/artifact calls getRuntimeToken() and never references GITHUB_TOKEN. So no permissions: value can produce the 403 being described. upload-artifact's README documents no required scope, consistent with that.

Empirically, in this repo, on this branch. CI.yml's kube-linter job uploads an artifact under contents: read + security-events: write — no actions scope at all. Run 33508683127: job KubeLinter Scan (Helm Chart) → success, step Upload KubeLinter SARIF artifact → success. cargo-coverage does the same inheriting only contents: read, with if-no-files-found: error.

So the specific ask  "grant actions: write on build-and-push, scan-image, promote-tags"  would broaden three release-path tokens to real API reach over workflow runs, in a PR whose purpose is narrowing tokens, to fix a 403 that cannot occur. That's the one recommendation here I'd refuse outright.

Worth noting the reviewer says this path is "entirely unproven, nothing exercises upload-artifact under narrowed permissions." It's exercised on every single PR by kube-linter — that's the proof, and it predates this PR.

I added a comment on build-and-push's permissions block recording the mechanism, the toolkit reference, and the kube-linter evidence, since this is the third round on it.

One thing their comment surfaced that is real, in the opposite direction: promote-tags carries actions: read and has no upload steps and no Actions API calls  and scan-image's actions: read is equally unused. Those look like residue from an earlier round chasing this same misconception. They're harmless but they're two scopes this PR could drop. I left them alone rather than widen scope unprompted  say the word and I'll remove both.

Fair residual: the deploy jobs themselves only run on a tag push, so the release path end-to-end is genuinely unexercised  that's already in your Known Gaps. But the permissions question specifically is settled; it isn't part of that gap.

ndefokou (approved):
All my changes have been address LGTM

martcpp:
@Christiantyemele pls check this

Christiantyemele (approved):

Issue #456, opened 2026-08-27
File-watch based zero-downtime rotation for database credentials and token signing keys


Hermann-Core:
## Objective

Implement continuous in-place file-watching so that when an External Secrets Operator or Kubernetes Secret volume updates a mounted file (such as a rotated database password, or updated token signing keys / issuer certificates for JWT/CWT Status List generation), the server reloads the active connection pool or token signing material without pod restarts, dropped requests, or service downtime.

## Context

- **Database**: Connection pool is created once at startup from `resolved_url()`. Changing the password in a mounted file currently requires a pod restart.
- **Token Signing Material**: When using `provisioning_strategy=store` with filesystem paths (`server.cert.store.certificate_path` and `server.cert.store.signing_key_path`), files are loaded once at startup. Rotating keys requires restarting the server.

### 1. Shared FileWatcher Core
- [ ] Implement an asynchronous `FileWatcher` component (using `notify` crate with fallback polling):
  - Debounce file change events (500ms) to handle Kubernetes atomic Secret symlink swaps (`..data`)
  - Always active when a watched file path is configured (no enable/disable flag)
  - Configurable fallback polling interval: `APP_WATCHER__POLL_INTERVAL_SECS` (default: `30`)
  - Dispatches reload events in background Tokio tasks

### 2. Database Credential Rotation
- [ ] Support `APP_DATABASE__PASSWORD_FILE`: path to mounted password file
- [ ] On file change, execute zero-downtime connection pool replacement:
  1. Read new password and reconstruct connection URL via `resolved_url()`
  2. Instantiate new `DatabaseConnection` connection pool
  3. Validate connectivity via test query
  4. Atomically swap active pool pointer (`ArcSwap` or equivalent)
  5. Gracefully drain and close previous pool after active queries finish
- [ ] On validation failure: retain existing healthy pool, emit error log and metric, avoid crash

### 3. Status List Token Signing Key & Issuer Certificate Rotation (for `-fscert`)
- [ ] When `server.cert.store.certificate_path` and `server.cert.store.signing_key_path` are configured, watch both files
- [ ] If either file changes, reload both files together as an atomic pair
- [ ] Validate new key and certificate pair (ensure private key matches certificate public key and parses as valid JWT/CWT signing material)
- [ ] Atomically swap the active `CertificateProvider` / signing key reference
- [ ] On validation failure: retain existing active key pair, log failure reason, emit metric

### 4. Telemetry & Observability
- [ ] Metrics:
[...]

PR #462, opened 2026-08-28
feat(external-secrets): implement ClusterSecretStore support and enhance secret management

### LINKED ISSUE #457: Extend Helm chart for External Secrets Operator environments and file-based rotation
## Objective

Make the Helm chart vendor-neutral and production-ready for constrained environments where secrets are managed by External Secrets Operator (supporting `ClusterSecretStore` and custom target templates), and wire file-based credential/key rotation into the deployment templates.

### 1. External Secrets Operator Extensions
- [x] Support `ClusterSecretStore` via `secretStoreRef.kind` in `helm/chart/templates/external-secrets.yaml`
- [x] Support custom secret `data` mappings and `target.template` for composite secret generation
- [x] Add `values-external-secrets.yaml` showcasing the Vault `ClusterSecretStore` pattern

### 2. File-Based Secret & Key Mounts
- [x] Add `secretMounts` configuration to `helm/chart/values.yaml`:
  ```yaml
  statuslist:
    secretMounts:
      - name: db-credentials
        secretName: postgres-access
        mountPath: /etc/secrets/database
        items:
          - key: password
            path: password
      - name: token-signing-keys
        secretName: statuslist-token-keys
        mountPath: /etc/secrets/token-keys
        items:
          - key: statuslist.crt
            path: statuslist.crt
          - key: statuslist.key
            path: statuslist.key
  ```
- [x] Update `helm/chart/templates/deployment.yaml` to:
  - Mount configured `secretMounts` volumes
  - Inject `APP_DATABASE__PASSWORD_FILE` when database secret mount is present
  - Inject `APP_SERVER__CERT__STORE__CERTIFICATE_PATH` and `APP_SERVER__CERT__STORE__SIGNING_KEY_PATH` when token key mounts are present
  - Inject `APP_WATCHER__POLL_INTERVAL_SECS` when set
- [x] Add checksum annotations for operators preferring rolling restarts over in-place reload:
  ```yaml
  checksum/secret: {{ include (print $.Template.BasePath "/external-secrets.yaml") . | sha256sum }}
  ```
- [x] Provide `values-rotation-example.yaml` documenting the complete zero-downtime deployment pattern

## Acceptance Criteria

- [x] `helm template` cleanly renders manifests for AWS, GCP, Azure, and Vault `ClusterSecretStore`.
- [x] `secretMounts` automatically mounts volumes and sets appropriate file path environment variables.
- [x] Default values remain backward-compatible.
- [x] Passes `helm lint` and `kube-linter` validation.

## Dependencies

- Blocked by #386 (Helm: enable Workload Identity and provider-neutral SecretStore)
- Blocked by #393 (Helm: remove provider-specific defaults and add values schema validation)
- Depends on Ticket: File-watch based zero-downtime rotation for database credentials and token signing keys

Ngha-Boris:
## Summary
Extends the Helm chart to support External Secrets Operator (ESO) environments with ClusterSecretStore and file-based credential/key rotation. This makes the chart vendor-neutral and production-ready for constrained environments where secrets are managed externally.

### External Secrets Operator Extensions
- Added `secretStoreRef.kind` support in `external-secrets.yaml` enabling `ClusterSecretStore` (cluster-scoped) alongside namespace-scoped `SecretStore`
- Added support for custom `data` mappings and `target.template` for composite secret generation
- Added validation to prevent SecretStore rendering when ClusterSecretStore is configured
- Created `values-external-secrets.yaml` with complete examples for:
  - HashiCorp Vault with ClusterSecretStore
  - AWS Secrets Manager with namespace-scoped SecretStore
  - GCP Secret Manager with ClusterSecretStore
  - Azure Key Vault with ClusterSecretStore and WorkloadIdentity

### File-Based Secret & Key Mounts
- Added `secretMounts` configuration to `values.yaml` for mounting secrets as files
- Updated `deployment.yaml` to:
  - Mount configured `secretMounts` volumes with read-only access
  - Auto-inject `APP_DATABASE__PASSWORD_FILE` when `db-credentials` mount is present
  - Auto-inject `APP_SERVER__CERT__STORE__CERTIFICATE_PATH` and `APP_SERVER__CERT__STORE__SIGNING_KEY_PATH` when `token-signing-keys` mount is present
  - Support `items` subPath mapping for selective key mounting
- Added `checksum/secret` annotation for rolling restarts when ExternalSecret templates change

### Documentation & Examples
- Created `values-rotation-example.yaml` documenting:
  - Complete zero-downtime credential rotation pattern
  - Database credential rotation procedure
  - Signing key rotation with grace periods
  - Emergency rotation procedures
  - Validation commands

## Configuration Example

```yaml
# External Secrets with Vault ClusterSecretStore
externalSecret:
  enabled: true
  spec:
    refreshInterval: 15m
    secretStoreRef:
      name: vault-cluster-store
      kind: ClusterSecretStore
    target:
      name: statuslist-secret
      template:
        type: Opaque
        data:
          postgres-password: "{{ .db_password }}"
          connection-string: "postgresql://{{ .db_user }}:{{ .db_password }}@{{ .db_host }}:5432/statuslist"
    data:
      - secretKey: db_password
        remoteRef:
          key: secret/data/statuslist/database
          property: password

secretStore:
  enabled: false  # ClusterSecretStore managed by cluster admins

# File-based secret mounts for rotation
statuslist:
  secretMounts:
    - name: db-credentials
      secretName: statuslist-db-credentials
      mountPath: /etc/secrets/database
      items:
        - key: password
          path: password
    - name: token-signing-keys
      secretName: statuslist-token-keys
      mountPath: /etc/secrets/token-keys
      items:
        - key: statuslist.crt
          path: statuslist.crt
        - key: statuslist.key
          path: statuslist.key
  env:
    APP_WATCHER__POLL_INTERVAL_SECS: "60"
```

## Testing

- [x] `helm lint` passes
- [x] `helm template` renders correctly for AWS Secrets Manager
- [x] `helm template` renders correctly for GCP Secret Manager
- [x] `helm template` renders correctly for Azure Key Vault
- [x] `helm template` renders correctly for HashiCorp Vault ClusterSecretStore
- [x] `secretMounts` volumes mount correctly with `items` subPath
- [x] File path environment variables auto-inject correctly
- [x] Checksum annotations trigger rolling restarts
- [x] Backward compatibility maintained (default values unchanged)

## Checklist

- [x] Chart version bumped (if applicable)
- [x] Documentation updated (`values-rotation-example.yaml`, `values-external-secrets.yaml`)
- [x] Examples provided for all supported secret store providers
- [x] Validation rules prevent misconfiguration
- [x] Security best practices followed (read-only mounts, non-root user)

## Breaking Changes

None. All changes are backward-compatible:
- New `secretMounts` defaults to empty list `[]`
- `secretStoreRef.kind` defaults to `"SecretStore"` (existing behavior)
- Template changes only activate when `externalSecret.enabled=true`

### DIFF
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -106,14 +106,31 @@ spec:
             {{- end }}
             {{- $secretFileEnv := dict }}
             {{- range $mount := .Values.statuslist.secretMounts }}
+              {{- $items := $mount.items | default list }}
+              {{- $itemPaths := dict }}
+              {{- range $item := $items }}
+                {{- $itemPath := trim (toString $item.path) }}
+                {{- $normalizedItemPath := clean $itemPath }}
+                {{- $_ := set $itemPaths $normalizedItemPath true }}
+              {{- end }}
               {{- with $mount.fileEnv }}
                 {{- range $name, $path := . }}
                   {{- $pathString := trim (toString $path) }}
+                  {{- $normalizedPath := clean $pathString }}
                   {{- if eq $pathString "" }}
                     {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s must be a non-empty relative file path" $mount.name $name) }}
                   {{- end }}
                   {{- if hasPrefix "/" $pathString }}
                     {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s must be relative to mountPath, got absolute path %q" $mount.name $name $pathString) }}
+                  {{- end }}
+                  {{- if or (eq $normalizedPath ".") (eq $normalizedPath "..") (hasPrefix "../" $normalizedPath) }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s must resolve inside mountPath, got %q" $mount.name $name $pathString) }}
+                  {{- end }}
+                  {{- if ne $normalizedPath $pathString }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s must be a normalized relative file path, got %q" $mount.name $name $pathString) }}
+                  {{- end }}
+                  {{- if and (gt (len $items) 0) (not (hasKey $itemPaths $normalizedPath)) }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s references %q, but items mounts only: %s" $mount.name $name $pathString (keys $itemPaths | sortAlpha | join ", ")) }}
                   {{- end }}
                   {{- if hasKey $env $name }}
                     {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s duplicates statuslist.env.%s; define each environment variable only once" $mount.name $name $name) }}
@@ -121,7 +138,7 @@ spec:
                   {{- if hasKey $secretFileEnv $name }}
                     {{- fail (printf "statuslist.secretMounts fileEnv defines %s more than once" $name) }}
                   {{- end }}
-                  {{- $_ := set $secretFileEnv $name (printf "%s/%s" $mount.mountPath $pathString) }}
+                  {{- $_ := set $secretFileEnv $name (printf "%s/%s" $mount.mountPath $normalizedPath) }}
                 {{- end }}
               {{- end }}
             {{- end }}
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -142,6 +142,14 @@ spec:
                 {{- end }}
               {{- end }}
             {{- end }}
+            {{- range $name, $_ := $secretFileEnv }}
+              {{- if eq $name "APP_DATABASE__URL" }}
+                {{- fail "statuslist.secretMounts fileEnv must not define APP_DATABASE__URL because it would expose assembled credentials in pod metadata; use APP_DATABASE__HOST/PORT/USERNAME/NAME/QUERY plus APP_DATABASE__PASSWORD_FILE" }}
+              {{- end }}
+              {{- if eq $name "APP_DATABASE__PASSWORD" }}
+                {{- fail "statuslist.secretMounts fileEnv must not define APP_DATABASE__PASSWORD; use APP_DATABASE__PASSWORD_FILE for mounted database password files" }}
+              {{- end }}
+            {{- end }}
             {{- if hasKey $env "APP_DATABASE__URL" }}
               {{- fail "statuslist.env.APP_DATABASE__URL is not supported by this chart because it exposes assembled credentials in pod metadata; use APP_DATABASE__HOST/PORT/USERNAME/NAME/QUERY plus the chart-managed APP_DATABASE__PASSWORD SecretKeyRef" }}
             {{- end }}
@@ -200,6 +208,35 @@ spec:
             {{- if not (hasKey $env "APP_DATABASE__NAME") }}
               {{- $_ := set $env "APP_DATABASE__NAME" .Values.postgres.auth.database }}
             {{- end }}
+            {{- range $name, $_ := $secretFileEnv }}
+              {{- if hasKey $env $name }}
+                {{- fail (printf "statuslist.secretMounts fileEnv defines %s, which duplicates a rendered statuslist environment variable; define each environment variable only once" $name) }}
+              {{- end }}
+            {{- end }}
+            {{- $chartManagedEnv := dict }}
+            {{- $appRegionForValidation := include "status-list-server-chart.appRegion" . }}
+            {{- $watcherPollInterval := "" }}
+            {{- if .Values.statuslist.watcher }}
+              {{- $watcherPollInterval = trim (toString (.Values.statuslist.watcher.pollIntervalSecs | default "")) }}
+            {{- end }}
+            {{- if and (eq .Values.secretStore.provider "aws") $appRegionForValidation }}
+              {{- $_ := set $chartManagedEnv "APP_AWS__REGION" true }}
+            {{- end }}
+            {{- if .Values.statuslist.aws.mountCredentials }}
+              {{- $_ := set $chartManagedEnv "AWS_SHARED_CREDENTIALS_FILE" true }}
+              {{- $_ := set $chartManagedEnv "AWS_CONFIG_FILE" true }}
+            {{- end }}
+            {{- if ne $watcherPollInterval "" }}
+              {{- $_ := set $chartManagedEnv "APP_WATCHER__POLL_INTERVAL_SECS" true }}
+            {{- end }}
+            {{- if not (or (hasKey $env "APP_DATABASE__PASSWORD_FILE") (hasKey $secretFileEnv "APP_DATABASE__PASSWORD_FILE")) }}
+              {{- $_ := set $chartManagedEnv "APP_DATABASE__PASSWORD" true }}
+            {{- end }}
+            {{- range $name, $_ := $secretFileEnv }}
+              {{- if hasKey $chartManagedEnv $name }}
+                {{- fail (printf "statuslist.secretMounts fileEnv defines %s, which duplicates a chart-managed environment variable; define each environment variable only once" $name) }}
+              {{- end }}
+            {{- end }}
             {{- range $name, $value := $env }}
             - name: {{ $name }}
               value: {{ $value | quote }}
@@ -220,11 +257,9 @@ spec:
               value: {{ $path | quote }}
             {{- end }}
             {{- /* Watcher poll interval for credential rotation detection */}}
-            {{- if and .Values.statuslist.watcher .Values.statuslist.watcher.pollIntervalSecs }}
-            {{- if ne .Values.statuslist.watcher.pollIntervalSecs "" }}
+            {{- if ne $watcherPollInterval "" }}
             - name: APP_WATCHER__POLL_INTERVAL_SECS
-              value: {{ .Values.statuslist.watcher.pollIntervalSecs | quote }}
-            {{- end }}
+              value: {{ $watcherPollInterval | quote }}
             {{- end }}
             {{- if not (or (hasKey $env "APP_DATABASE__PASSWORD_FILE") (hasKey $secretFileEnv "APP_DATABASE__PASSWORD_FILE")) }}
             - name: APP_DATABASE__PASSWORD
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -200,6 +200,13 @@ spec:
             {{- end }}
             {{- end }}
             {{- end }}
+            {{- /* Watcher poll interval for credential rotation detection */}}
+            {{- if and .Values.statuslist.watcher .Values.statuslist.watcher.pollIntervalSecs }}
+            {{- if ne .Values.statuslist.watcher.pollIntervalSecs "" }}
+            - name: APP_WATCHER__POLL_INTERVAL_SECS
+              value: {{ .Values.statuslist.watcher.pollIntervalSecs | quote }}
+            {{- end }}
+            {{- end }}
             - name: APP_DATABASE__PASSWORD
               valueFrom:
                 secretKeyRef:
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -22,6 +22,9 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: {{ .Values.statuslist.service.targetPort | quote }}
         prometheus.io/path: "/metrics"
+        {{- if .Values.externalSecret.enabled }}
+        checksum/secret: {{ include (print $.Template.BasePath "/external-secrets.yaml") . | sha256sum }}
+        {{- end }}
         {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
@@ -174,6 +177,29 @@ spec:
             - name: AWS_CONFIG_FILE
               value: /home/nobody/.aws/config
             {{- end }}
+            {{- /* File-based secret mounts: database credentials */}}
+            {{- range .Values.statuslist.secretMounts }}
+            {{- if eq .name "db-credentials" }}
+            - name: APP_DATABASE__PASSWORD_FILE
+              value: {{ .mountPath }}/{{ (index .items 0).path }}
+            {{- end }}
+            {{- end }}
+            {{- /* File-based secret mounts: token signing keys */}}
+            {{- range .Values.statuslist.secretMounts }}
+            {{- if eq .name "token-signing-keys" }}
+            {{- $mountPath := .mountPath }}
+            {{- range .items }}
+            {{- if eq .key "statuslist.crt" }}
+            - name: APP_SERVER__CERT__STORE__CERTIFICATE_PATH
+              value: {{ $mountPath }}/{{ .path }}
+            {{- end }}
+            {{- if eq .key "statuslist.key" }}
+            - name: APP_SERVER__CERT__STORE__SIGNING_KEY_PATH
+              value: {{ $mountPath }}/{{ .path }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
             - name: APP_DATABASE__PASSWORD
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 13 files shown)


helm/chart/templates/deployment.yaml
```diff
@@ -174,6 +177,36 @@ spec:
...
+              value: {{ $mountPath }}/{{ .path }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- /* Watcher poll interval for credential rotation detection */}}
+            {{- if and .Values.statuslist.watcher .Values.statuslist.watcher.pollIntervalSecs }}
+            {{- if ne .Values.statuslist.watcher.pollIntervalSecs "" }}
+            - name: APP_WATCHER__POLL_INTERVAL_SECS
+              value: {{ .Values.statuslist.watcher.pollIntervalSecs | quote }}
```
forkimenjeckayang: Mounting the Secret as files is the correct direction, and these variable names match the contract proposed in [#456](https://github.com/adorsys/status-list-server/issues/456). However, the current server has no `password_file` or watcher configuration, builds the database pool once at startup, and still reads `APP_DATABASE__PASSWORD`. Please describe this as preparatory chart support and make the example explicitly require an application image containing #456. Also define file-versus-environment precedence, or stop injecting `APP_DATABASE__PASSWORD` when the file mount is selected, so the application cannot silently keep using the stale startup value.

helm/chart/templates/external-secrets.yaml
```diff
@@ -47,11 +53,15 @@ spec:
...
-    kind: {{ .Values.externalSecret.spec.secretStoreRef.kind }}
+    kind: {{ .Values.externalSecret.spec.secretStoreRef.kind | default "SecretStore" }}
   target:
     # Fixed name — the Deployment volume references "aws-credentials-secret" directly.
     name: aws-credentials-secret
     creationPolicy: {{ .Values.externalSecret.spec.target.creationPolicy }}
+    {{- with .Values.externalSecret.spec.target.template }}
+    template:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
```
forkimenjeckayang: This copies the application Secret's template onto the separate AWS-credentials ExternalSecret. With the new examples, that second resource fetches only `credentials` and `config`, while the copied template references `db_password`, `api_key`, or `postgres_password`. ESO treats missing template keys as errors, so `aws-credentials-secret` will not reconcile and the Pod cannot mount it. Please do not apply the application target template here; use a dedicated AWS-credentials template if one is needed. Reference: [ESO advanced templating](https://external-secrets.io/main/guides/templating/).

helm/chart/values-rotation-example.yaml
```diff
@@ -0,0 +1,249 @@
...
+          aws-credentials: "{{ .aws_credentials }}"
+    data:
+      - secretKey: postgres-password
+        remoteRef:
+          key: secret/data/statuslist/production/database
+          property: password
+      - secretKey: aws-credentials
+        remoteRef:
+          key: secret/data/statuslist/production/aws
+          property: credentials
```
forkimenjeckayang: This file is described as a complete production example, but none of these three mounted Secrets is created by it. The rendered ExternalSecrets target only `statuslist-secret` and `aws-credentials-secret`, so the Pod cannot mount `statuslist-db-credentials`, `statuslist-token-keys`, or `statuslist-api-keys` unless somebody provisions them separately. In our intended model, please add the required ExternalSecrets or map these files from an existing target Secret. If customer-side provisioning is intentional, document the exact Secret names and required keys as prerequisites. Reference: [Kubernetes Secret volumes](https://kubernetes.io/docs/concepts/storage/volumes/#secret).

helm/chart/templates/deployment.yaml
```diff
@@ -22,6 +22,9 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: {{ .Values.statuslist.service.targetPort | quote }}
         prometheus.io/path: "/metrics"
+        {{- if .Values.externalSecret.enabled }}
+        checksum/secret: {{ include (print $.Template.BasePath "/external-secrets.yaml") . | sha256sum }}
+        {{- end }}
```
forkimenjeckayang, line 27: This checksum covers only the Helm-rendered ExternalSecret manifest, not the Secret data later synchronized by ESO. A customer rotating a value in Vault therefore leaves this checksum unchanged and does not roll the Deployment. That is acceptable when the intended mechanism is mounted-file reload through #456, but the example must not claim that this annotation restarts Pods after an ESO sync. Please describe it only as reacting to Helm template/value changes, or remove it if restarts are not part of the design. References: [Helm checksum pattern](https://helm.sh/docs/v3/howto/charts_tips_and_tricks/#automatically-roll-deployments) and [Kubernetes Secret update behavior](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod).

**Also affected:** `helm/chart/values-rotation-example.yaml:7-12` and `helm/chart/values-rotation-example.yaml:214-220`

helm/chart/values-external-secrets.yaml
```diff
@@ -0,0 +1,144 @@
...
+#       name: statuslist-secret
+#       creationPolicy: Owner
+#       template:
+#         data:
+#           postgres-password: "{{ .password }}"
+#     data:
+#       - secretKey: password
+#         remoteRef:
+#           key: statuslist-postgres-password
+#           version: latest
```
forkimenjeckayang, line 100: If this GCP example is enabled, `version: latest` is silently dropped because `external-secrets.yaml` renders only `remoteRef.key` and `.property`. The same happens to ESO options such as `decodingStrategy`, `conversionStrategy`, and `metadataPolicy`. Please render the configured `remoteRef` map (for example with `toYaml`) so provider-specific mappings survive. Reference: [ESO ExternalSecret API](https://external-secrets.io/latest/api/externalsecret/).

**Underlying template:** `helm/chart/templates/external-secrets.yaml:30-35`

helm/chart/values-rotation-example.yaml
```diff
@@ -0,0 +1,249 @@
...
+# ---
+# Rotation Procedure
+# ---
+#
+# 1. Database Credential Rotation:
+#    a. Update password in external secret store (e.g., Vault)
+#    b. External Secrets Operator syncs to Kubernetes Secret
+#    c. Application detects file change via APP_WATCHER__POLL_INTERVAL_SECS
+#    d. Application reloads credentials without restart
+#    e. Verify: kubectl logs -l app.kubernetes.io/name=status-list-server | grep "credentials reloaded"
```
forkimenjeckayang: This procedure covers ESO delivery but not the database side of credential rotation. Updating Vault does not itself change the password accepted by PostgreSQL, and immediately revoking the old credential can still cause failures during the ESO refresh and watcher delay. Please document the supported rotation order, old/new credential overlap, validation, and rollback expectations. The zero-downtime claim should depend on the pool-swap behavior in #456 and an agreed overlap procedure with the external secret owner.

forkimenjeckayang (commented):
a few comments @Ngha-Boris 
Please check

helm/chart/templates/deployment.yaml
```diff
@@ -174,6 +177,36 @@ spec:
             - name: AWS_CONFIG_FILE
               value: /home/nobody/.aws/config
             {{- end }}
+            {{- /* File-based secret mounts: database credentials */}}
+            {{- range .Values.statuslist.secretMounts }}
+            {{- if eq .name "db-credentials" }}
+            - name: APP_DATABASE__PASSWORD_FILE
+              value: {{ .mountPath }}/{{ (index .items 0).path }}
```
Hermann-Core: **[MED] `(index .items 0).path` assumes `items` is always present — an empty-items mount breaks rendering**

`APP_DATABASE__PASSWORD_FILE` is derived from `(index .items 0).path`. But a `secretMounts` entry may legitimately omit `items` (the volume block at deployment.yaml:240 uses `{{- with .items }}`, so mounting every key with no subPath is supported). In that case `(index .items 0)` fails `helm template` with a "index of empty list" error. The same assumption applies to the `token-signing-keys` loop (lines 190-200). Recommend deriving the file path more defensibly — e.g. resolve the path from the matching `items` entry by key, or fail with a clear message when `items` is empty. As-is this only renders because the documented examples always supply `items`.

helm/chart/templates/deployment.yaml
```diff
@@ -174,6 +177,36 @@ spec:
             - name: AWS_CONFIG_FILE
               value: /home/nobody/.aws/config
             {{- end }}
+            {{- /* File-based secret mounts: database credentials */}}
+            {{- range .Values.statuslist.secretMounts }}
+            {{- if eq .name "db-credentials" }}
```
Hermann-Core: **[MED] Auto-injection is keyed on the free-form, unvalidated `secretMounts[].name`**

The env injection is coupled to the exact magic strings `"db-credentials"` (line 182) and `"token-signing-keys"` (line 189). `secretMounts[].name` is a free-form user string with no validation. An operator naming the mount anything else (e.g. `db-access`, `database-credentials`, `signing-keys`) silently gets the file mounted but **no** `APP_DATABASE__PASSWORD_FILE` / cert-path env — the rotation feature silently degrades with no error. This name contract is only hinted at in values.yaml comments. Recommend validating that a mount named `db-credentials`/`token-signing-keys` exists when file-based rotation is intended, or keying the injection off an explicit option on the mount entry rather than a magic string literal.

Hermann-Core (changes requested):
@Ngha-Boris please check the inline comments below (in addition to the earlier ones from @forkimenjeckayang). Two additional MED findings at head `cc104e0a`.

Ngha-Boris:
Hello @Hermann-Core @forkimenjeckayang please recheck

helm/chart/templates/external-secrets.yaml
```diff
@@ -62,3 +71,47 @@ spec:
         key: {{ .Values.statuslist.aws.credentialsSecret.remoteKey }}
         property: {{ .Values.statuslist.aws.credentialsSecret.configProperty }}
 {{- end }}
+{{- if and .Values.externalSecret.enabled .Values.externalSecret.spec.extraExternalSecrets }}
```
Hermann-Core, line 74: This new `extraExternalSecrets` template block (the most complex logic added in this revision) is never exercised by the render-time CI: `render-helm-templates/action.yml` renders `values.yaml`, `values-production.yaml`, and a battery of `--set` combos, but no case sets `externalSecret.spec.extraExternalSecrets` nor renders `values-rotation-example.yaml`. Nit: add one render case (e.g. `-f helm/chart/values-rotation-example.yaml` or a `--set` extraExternalSecrets[0]...) and assert the extra Secrets (`statuslist-db-credentials`, `statuslist-api-keys`) are emitted so this block is covered before merge. Non-blocking.

Hermann-Core (commented):
@Ngha-Boris this re-review is at the new head `3a508aa`. All findings from my previous CHANGES_REQUESTED review `5069545418` are now resolved, and I've verified that all 6 of forkimenjeckayang's earlier inline comments are also addressed. Two non-blocking hardening suggestions are posted as inline comments below; no blocking issues remain from my side.

helm/chart/templates/deployment.yaml
```diff
@@ -101,6 +104,27 @@ spec:
...
+                  {{- if hasPrefix "/" $pathString }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s must be relative to mountPath, got absolute path %q" $mount.name $name $pathString) }}
+                  {{- end }}
+                  {{- if hasKey $env $name }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s duplicates statuslist.env.%s; define each environment variable only once" $mount.name $name $name) }}
+                  {{- end }}
+                  {{- if hasKey $secretFileEnv $name }}
+                    {{- fail (printf "statuslist.secretMounts fileEnv defines %s more than once" $name) }}
+                  {{- end }}
+                  {{- $_ := set $secretFileEnv $name (printf "%s/%s" $mount.mountPath $pathString) }}
```
Hermann-Core: The `fileEnv` value is a free-form relative path joined onto `mountPath`, but it is not cross-validated against the mount's `items[].path` (or the Secret's keys when `items` is absent). An operator could set `fileEnv.APP_DATABASE__PASSWORD_FILE: pwd` while `items` produces a file named `password`, and the chart renders an env var pointing at a file that is never mounted (the app then silently falls back / errors at startup). The JSON schema could `enum`-match each `fileEnv` value against `items[].path`, or the template could `fail` when a fileEnv path has no matching normalized item path. The new render-time negative tests would be the natural place to lock this down. Non-blocking.

Hermann-Core (dismissed):
LGTM

Hermann-Core:
@forkimenjeckayang please can you check back this ? It is an important ticket blocking others.

helm/chart/values-rotation-example.yaml
```diff
@@ -0,0 +1,320 @@
...
+      # Template to map external secret structure to expected application Secret keys.
+      template:
+        type: Opaque
+        data:
+          postgres-password: "{{ .postgres_password }}"
+    data:
+      - secretKey: postgres-password
+        remoteRef:
+          key: secret/data/statuslist/production/database
+          property: password
```
forkimenjeckayang, line 43: The template reads `.postgres_password`, but the fetched value is exposed as `postgres-password`. ESO keeps the `secretKey` name exactly; it does not replace `-` with `_`, and a missing template key makes reconciliation fail. As a result, this example will not create `statuslist-secret`. Please align the names—for example, fetch it as `secretKey: postgres_password` and keep the template output key `postgres-password`. Reference: [ESO advanced templating](https://external-secrets.io/main/guides/templating/).

helm/chart/templates/deployment.yaml
```diff
@@ -101,6 +104,44 @@ spec:
...
+                  {{- if and (gt (len $items) 0) (not (hasKey $itemPaths $normalizedPath)) }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s references %q, but items mounts only: %s" $mount.name $name $pathString (keys $itemPaths | sortAlpha | join ", ")) }}
+                  {{- end }}
+                  {{- if hasKey $env $name }}
+                    {{- fail (printf "statuslist.secretMounts[%s].fileEnv.%s duplicates statuslist.env.%s; define each environment variable only once" $mount.name $name $name) }}
+                  {{- end }}
+                  {{- if hasKey $secretFileEnv $name }}
+                    {{- fail (printf "statuslist.secretMounts fileEnv defines %s more than once" $name) }}
+                  {{- end }}
+                  {{- $_ := set $secretFileEnv $name (printf "%s/%s" $mount.mountPath $normalizedPath) }}
```
forkimenjeckayang, line 141: This validation catches names already in `statuslist.env`, but `fileEnv` can still reuse variables that the chart adds later. For example, `fileEnv.APP_DATABASE__PASSWORD=password` renders two `APP_DATABASE__PASSWORD` entries: the file path here and the chart's `secretKeyRef` below. The same happens with `APP_WATCHER__POLL_INTERVAL_SECS` and the AWS file variables, while `fileEnv.APP_DATABASE__URL` bypasses the protected-name check entirely. Please validate `fileEnv` against the chart-managed and prohibited names, or build one combined environment map and reject every duplicate before rendering.

helm/chart/templates/external-secrets.yaml
```diff
@@ -1,11 +1,13 @@
...
+{{- /* ClusterSecretStore is cluster-scoped and managed by cluster admins, not rendered by this chart */}}
+{{- $isClusterStore := eq (.Values.externalSecret.spec.secretStoreRef.kind | default "SecretStore") "ClusterSecretStore" }}
+{{- if and (not .Values.secretStore.enabled) (not $isClusterStore) }}
 {{- fail "externalSecret.enabled requires secretStore.enabled: no SecretStore would be rendered, leaving the ExternalSecret with a dangling secretStoreRef" }}
 {{- end }}
 {{- if ne .Values.externalSecret.spec.target.name (include "status-list-server-chart.appSecretName" .) }}
 {{- fail (printf "externalSecret.spec.target.name must be '%s' to match the single application-secret name used by the Deployment and PostgreSQL (postgres.auth.existingSecret)" (include "status-list-server-chart.appSecretName" .)) }}
 {{- end }}
-apiVersion: external-secrets.io/v1beta1
+apiVersion: external-secrets.io/v1
```
forkimenjeckayang, line 10: Using `external-secrets.io/v1` is correct, but it raises the minimum cluster prerequisite and the deployment README currently does not list ESO at all. Please document that ESO and its CRDs must serve `external-secrets.io/v1` before installing this chart, and mention that the operator and CRDs must be upgraded together. Reference: [ESO v0.16 promotion guide](https://github.com/external-secrets/external-secrets/releases/tag/v0.16.0).

forkimenjeckayang (changes requested):
a few comments please check

Ngha-Boris:
Hello @forkimenjeckayang please recheck

forkimenjeckayang (approved):

Hermann-Core (approved):

PR #465, opened 2026-09-01
feat: add support for reloading database credentials from a password file

### LINKED ISSUE #456: File-watch based zero-downtime rotation for database credentials and token signing keys

Ngha-Boris:
## Summary

Implements file-watch based zero-downtime rotation for mounted database credentials and filesystem token signing material.

## Changes

- Added async `FileWatcher` with `notify` support, 500ms debounce, Kubernetes Secret symlink-swap handling, and polling fallback via `APP_WATCHER__POLL_INTERVAL_SECS`.
- Added `APP_DATABASE__PASSWORD_FILE` support.
- Added atomic database pool swapping with `ArcSwap`, retaining the existing healthy pool on reload/validation failure.
- Added graceful old-pool drain/close after successful database rotation.
- Added reloadable filesystem certificate/signing-key provider with atomic last-known-good swapping.
- Added validation for token signing material:
  - private key parses as signing material
  - certificate parses as X.509
  - certificate public key matches signing key
  - key is valid for JWT signing
- Added rotation metrics:
  - `credential_rotation_total{target="database|token_signing_key", outcome="success|failure"}`
  - `credential_rotation_last_success_timestamp{target="database|token_signing_key"}`
- Added structured lifecycle logs for `file_change_detected`, `rotation_started`, `rotation_succeeded`, and `rotation_failed`.
- Updated Helm deployment to mount the application Secret and configure `APP_DATABASE__PASSWORD_FILE` instead of exposing DB password as an env var.
- Added tests for watcher debounce, password-file URL resolution, token material validation, malformed reload retention, and Helm rendering.

## Validation

- `cargo test`
- `cargo test --no-run --features sqlite`
- `cargo test --no-run --features acme`


src/config.rs
```diff
@@ -593,6 +602,10 @@ fn trim_non_empty(value: Option<&str>) -> Option<&str> {
     value.map(str::trim).filter(|value| !value.is_empty())
 }
 
+fn trim_trailing_newline(value: &str) -> &str {
+    value.trim_end_matches(['\r', '\n'])
+}
+
 fn encode_url_part(value: &str) -> String {
```
martcpp: Do you nedd a special function to trim line just call the .trim on the place use need to use it

Cargo.toml
```diff
@@ -64,6 +64,8 @@ opt-level = 1
 # Async runtime and utilities
 tokio = { version = "1.5", features = ["full"] }
 async-trait = "0.1"
+arc-swap = "1.7"
+notify = { version = "8.2", default-features = false }
```
Hermann-Core: Blocking: `notify` with `default-features = false` does **not compile on macOS**. `notify 8.2.0` unconditionally compiles its fsevent module on `target_os="macos"` (`#[cfg(all(target_os = "macos", not(feature = "macos_kqueue")))]`), and `fsevent-sys` is only brought in by the `macos_fsevent` feature — disabled here. I verified with `cargo check --no-default-features --features sqlite`: `error[E0432]: unresolved import 'fsevent_sys'` at `notify-8.2.0/src/fsevent.rs:21`. Linux CI passes (inotify backend is unconditional) but every macOS local build/test — including the new rotation unit tests — is broken. Fix: `notify = { version = "8.2", default-features = false, features = ["macos_fsevent"] }` (fsevent-sys cross-compiles fine on Linux). Needs a Cargo.lock update + a cargo-deny exemption for fsevent-sys.

src/utils/file_watcher.rs
```diff
@@ -0,0 +1,229 @@
...
+    where
+        F: Fn() -> Fut + Send + Sync + 'static,
+        Fut: std::future::Future<Output = ()> + Send + 'static,
+    {
+        let callback = Arc::new(on_change);
+        let paths = self.paths.clone();
+        let (tx, rx) = mpsc::channel(32);
+
+        spawn_notify_watcher(paths.clone(), tx.clone(), target);
+        spawn_poll_watcher(paths.clone(), tx, target, self.poll_interval);
```
Hermann-Core: Design: both the native notify watcher and the polling watcher are **always** started, but issue #456 specifies polling as a *fallback*. On any real change, notify fires immediately and then, ~`poll_interval` (up to 30s) later, the poller — whose fingerprint baseline is never refreshed by the notify path — re-detects the same change and fires a second time. Net effect: **every rotation triggers twice**, causing a redundant pool rebuild / cert reload and doubled rotation metrics/logs. Consider only starting the poller when notify fails to initialize, or updating the poller's fingerprint baseline when notify reports a change.

Hermann-Core (commented):
One blocking build finding (notify on macOS), one high-leverage design issue (double-trigger), one concurrency correctness note, plus minor nits — all detailed in the review body and inline comments.

helm/chart/templates/deployment.yaml
```diff
@@ -159,6 +159,9 @@ spec:
             {{- if not (hasKey $env "APP_DATABASE__NAME") }}
               {{- $_ := set $env "APP_DATABASE__NAME" .Values.postgres.auth.database }}
             {{- end }}
+            {{- if not (hasKey $env "APP_DATABASE__PASSWORD_FILE") }}
+              {{- $_ := set $env "APP_DATABASE__PASSWORD_FILE" "/var/run/status-list-server/secrets/postgres-password" }}
```
Hermann-Core: This must not be hardcoded and must be retrieve by a proper mechanism. See https://github.com/adorsys/status-list-server/pull/462

helm/chart/templates/deployment.yaml
```diff
@@ -174,14 +177,12 @@ spec:
...
-              valueFrom:
-                secretKeyRef:
-                  name: {{ include "status-list-server-chart.appSecretName" . }}
-                  key: postgres-password
           volumeMounts:
             - name: tmp
               mountPath: /tmp
+            - name: app-secrets
+              mountPath: /var/run/status-list-server/secrets
+              readOnly: true
```
Hermann-Core: See comment https://github.com/adorsys/status-list-server/pull/465/changes#r3903842281.

src/utils/file_watcher.rs
```diff
@@ -0,0 +1,229 @@
...
+impl FileWatcher {
+    pub(crate) fn new(paths: Vec<PathBuf>, poll_interval: Duration) -> Self {
+        let mut unique = HashSet::new();
+        let paths = paths
+            .into_iter()
+            .filter(|path| unique.insert(path.clone()))
+            .collect();
+        Self {
+            paths: Arc::new(paths),
+            poll_interval: poll_interval.max(Duration::from_secs(1)),
```
Hermann-Core: Use of magic constant in code, prefer using a typed constant with a reasonable default value for this specific use case.

src/utils/rotation_metrics.rs
Hermann-Core, line 1: we already have a metrics module; you should reuse that instead of creating another one or move the logic here in an appropriate location.

src/config.rs
```diff
@@ -593,6 +602,10 @@ fn trim_non_empty(value: Option<&str>) -> Option<&str> {
     value.map(str::trim).filter(|value| !value.is_empty())
 }
 
+fn trim_trailing_newline(value: &str) -> &str {
+    value.trim_end_matches(['\r', '\n'])
+}
```
Hermann-Core: Isn't this just the `trim` method? If so, could you inline the call to `trim()` instead of introducing a separate function that only delegates to it?

Cargo.toml
```diff
@@ -64,6 +64,8 @@ opt-level = 1
 # Async runtime and utilities
 tokio = { version = "1.5", features = ["full"] }
 async-trait = "0.1"
+arc-swap = "1.7"
```
Hermann-Core:
```suggestion
arc-swap = "1.9"
```

src/setup.rs
```diff
@@ -120,6 +119,254 @@ fn classify_db_connect_error(err: &DbErr) -> &'static str {
...
+                        "token signing material rotation failed"
+                    );
+                }
+            }
+        }
+    });
+}
+
+#[cfg(not(feature = "acme"))]
+fn spawn_reloading_certificate_provider_rotation(
```
Hermann-Core: Having long names is not Rust philosophy, the name should be short and expressive.

src/config.rs
```diff
@@ -776,6 +794,23 @@ impl DatabaseConfig {
         )))
     }
 
+    pub async fn resolved_url_with_password_file(&self) -> Result<SecretString, ConfigError> {
```
Hermann-Core: Could you use a more representative name here? This function doesn't only resolve the URL from the password file; it handles both cases. The name should reflect its actual responsibility rather than implying that it only deals with the password file.

src/utils/file_watcher.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn path_relevant(changed: &Path, watched: &Path) -> bool {
+    changed == watched
+        || changed.starts_with(watched)
+        || watched.starts_with(changed)
+        || changed.file_name().is_some_and(|name| name == "..data")
+        || changed
+            .parent()
+            .is_some_and(|parent| watched.parent() == Some(parent))
```
Hermann-Core: This seems too permissive. The last branch matches any file changed in the same directory as any watched path, not just the watched file itself.
```suggestion
    changed == watched
        || changed.starts_with(watched)
        || watched.starts_with(changed)
        || (changed.file_name().is_some_and(|n| n == "..data")
            && changed.parent() == watched.parent())
```

src/utils/file_watcher.rs
```diff
@@ -0,0 +1,229 @@
...
+                event = "file_change_detected",
+                rotation.target = target,
+                paths = ?paths,
+                "watched secret file changed"
+            );
+            while matches!(tokio::time::timeout(DEBOUNCE, rx.recv()).await, Ok(Some(_))) {}
+            let callback = callback.clone();
+            tokio::spawn(async move {
+                callback().await;
+            });
```
Hermann-Core: Each debounced batch spawns a new task and immediately goes back to `rx.recv()`. If a second burst of changes arrives before the first `callback()`` finishes, you get two invocations running concurrently. For a rotation callback that presumably reads-then-swaps key material, that's a real correctness risk (partial reads, duplicate downstream effects), not just a performance nit.

Fix: serialize invocations instead of fire-and-forget spawning, e.g. track the previous `JoinHandle` and await it before starting the next one.
```rust
let mut in_flight: Option<tokio::task::JoinHandle<()>> = None;
while rx.recv().await.is_some() {
    while matches!(tokio::time::timeout(DEBOUNCE, rx.recv()).await, Ok(Some(_))) {}
    if let Some(prev) = in_flight.take() {
        let _ = prev.await; // wait for prior rotation to finish
    }
    let callback = callback.clone();
    in_flight = Some(tokio::spawn(async move { callback().await; }));
}
```
or route everything through a single dedicated worker task via the channel. Simpler to reason about than juggling handles.

src/utils/file_watcher.rs
```diff
@@ -0,0 +1,229 @@
...
+                    path = %watched.display(),
+                    error = %err,
+                    "file watcher could not watch path; polling remains active"
+                );
+            }
+        }
+
+        loop {
+            std::thread::park();
+        }
```
Hermann-Core, line 127: This works, watcher stays alive because it's still in scope but it's a non-obvious way to keep it alive, and there's no way to ever stop it. `FileWatcher::spawn` gives the caller nothing back (no handle, no cancellation token), so every call permanently leaks one OS thread plus the notify backend thread. Fine if spawn is called exactly once per process lifetime; a latent problem if it's ever called on reconfiguration/reload.

Consider returning a handle with a `shutdown()` (backed by a `CancellationToken` or a oneshot channel checked in the park loop / poll loop), even if you don't wire it up yet, makes the intent explicit and the leak a choice rather than an accident. At minimum, a comment explaining why the loop exists would help the next reader.

src/utils/file_watcher.rs
```diff
@@ -0,0 +1,229 @@
...
+            .is_some_and(|parent| watched.parent() == Some(parent))
+}
+
+async fn fingerprint_all(paths: &[PathBuf]) -> HashMap<PathBuf, Option<u64>> {
+    let mut fingerprints = HashMap::new();
+    for path in paths {
+        fingerprints.insert(path.clone(), fingerprint(path).await);
+    }
+    fingerprints
+}
```
Hermann-Core, line 233: Two independent things worth improving:

- Sequential I/O: paths are fingerprinted one at a time. `futures::future::join_all` (or `FuturesUnordered`) would parallelize the reads, matters more as the number of watched files grows.
- Full content read + hash every poll tick: for change detection, `std::fs::metadata()` (mtime + len) is usually enough and much cheaper than reading and hashing full file bytes each tick. Content hashing is more robust against clock skew / same-second writes, so you might keep it as a fallback when metadata is ambiguous, but as the sole mechanism it's doing more I/O than necessary. Given secrets are typically small this is probably not urgent, just worth knowing the cost scales with file size × poll frequency × file count.

src/utils/file_watcher.rs
Hermann-Core, line 1: **Minor nits**:

- `tokio::time::interval(poll_interval).tick()` fires immediately on the first call, so the very first poll iteration re-fingerprints right after the initial fingerprint_all, harmless (comparison is a no-op) but wasted work. `tokio::time::interval_at(Instant::now() + poll_interval`, `poll_interval`) avoids it.
- The debounce channel carries `()`, so the debouncer can't tell the callback which path changed. It has to treat the whole FileWatcher's path set as dirty. If a FileWatcher ever watches multiple unrelated secrets, that means changing one forces a reload of all of them. Might be intentional given target: `&'static str` suggests one watcher per rotation target, but worth confirming that's always 1:1.
- No doc comments on the pub(crate) API. Worth a line each on new/spawn documenting the debounce/concurrency guarantees, since that's exactly the kind of thing that bit item # 2.
- `poll_interval.max(Duration::from_secs(1))` silently clamps caller input with no log. A `tracing::debug!` when clamping happens would save someone a confused debugging session later.

Hermann-Core (commented):
@Ngha-Boris can you also check the comments below?

Hermann-Core:
## Re-review confirms the prior findings are addressed ✅

I re-reviewed the updated branch (HEAD `ec57b17`) by building and testing on macOS (the platform where the original blocker manifested) and went through the full diff. All the substantive findings from my earlier review are resolved and **verified**:

### Fixed & verified
1. **Blocking — `notify` macOS build break (Cargo.toml:68)** — now `features = ["macos_fsevent"]`. Verified on macOS: `cargo check --features sqlite` **and** `cargo check --features acme` both compile cleanly (previously `E0432 unresolved import 'fsevent_sys'`). The test build also compiles `fsevent-sys v4.1.0` successfully.
2. **Double-trigger from dual watchers** — `spawn_notify_watcher` now returns `bool native_started` (true only when the backend initializes *and* all paths watch successfully), and the poller starts only `if !native_started`. Correct fallback semantics.
3. **Unserialized rotation callbacks** — the debouncer now runs callbacks **serially** (5s debounce drain, then `callback().await` before processing the next event). Locked in by new test `debouncer_serializes_callbacks` (asserts `max_active == 1`), which passes.
4. **`path_relevant` over-matching** — the k8s `..data` symlink case now requires the *same parent directory*, verified by `k8s_data_symlink_events_only_match_same_directory`.

### Validation results
- **Full lib unit suite: 190 passed / 0 failed** on macOS, including all rotation, watcher, cert (last-known-good), metrics_db, and metrics_http tests.
- **Helm integration tests: 7 passed / 0 failed**, incl. new `rendered_chart_does_not_duplicate_watcher_poll_interval`.
- **`ec57b17` config-test correction** is legitimate — it aligns the expectation with the intentional `password.trim()` behavior applied consistently to both env and file-based passwords.

### Out of scope (pre-existing, not a PR regression)
`src/test_utils.rs` does not compile under `--no-default-features --features sqlite` (the `Memory*` types are gated behind the `memory` feature but used unconditionally there). I confirmed this file is **byte-identical to develop** (the PR does not touch it), and CI runs with `--all-features`, so it does not affect this PR. Worth a separate cleanup if you ever test that posture.

This PR looks good to merge from my side — no remaining blockers.

Hermann-Core (dismissed):
All my comments have been addressed. Good work here @Ngha-Boris.

martcpp (changes requested):
Image the following test are missing

src/outbound/cert.rs
```diff
@@ -55,6 +58,97 @@ impl StoreCertificateProvider {
...
+) -> Result<SigningMaterial, StatusListError> {
+    let cert_pem = tokio::fs::read_to_string(cert_path)
+        .await
+        .map_err(|err| StatusListError::Backend(Box::new(err)))?;
+    let signing_key_pem = tokio::fs::read_to_string(key_path)
+        .await
+        .map_err(|err| StatusListError::Backend(Box::new(err)))?;
+    validate_signing_material(&cert_pem, &signing_key_pem)?;
+    Ok(SigningMaterial {
+        certificate_chain: Some(vec![cert_pem]),
```
martcpp: CertificateProvider's two-method port defeats the atomic pair guarantee; tokens can be issued with new key + old chain. → single signing_material() snapshot.

src/utils/metrics.rs
```diff
@@ -53,6 +67,44 @@ pub(crate) fn metrics_test_lock() -> std::sync::MutexGuard<'static, ()> {
         .unwrap_or_else(|poisoned| poisoned.into_inner())
 }
 
+fn rotation_metrics() -> RotationMetrics {
+    cached_instruments(&ROTATION_METRICS, || {
+        let meter = global::meter("status-list-server");
+        RotationMetrics {
+            total: meter
+                .u64_counter("credential_rotation_total")
```
martcpp, line 75: ACME-variant rotation has no atomic swap and is unserialized against the renewal cron; can persist a permanently mismatched pair.

src/config.rs
```diff
@@ -758,6 +771,7 @@ impl DatabaseConfig {
         let url_host = format_database_url_host(host);
         let username = required_config_field(self.username.as_deref(), "database.username")?;
         let password = required_secret_field(self.password.as_ref(), "database.password")?;
+        let password = password.trim();
```
martcpp, line 774: ReloadingCertificateProvider emits PEM where consumers require base64 DER. CWT issuance is broken on -fscert; JWT x5c is non-conformant; chains are collapsed.

src/setup.rs
```diff
@@ -120,6 +119,251 @@ fn classify_db_connect_error(err: &DbErr) -> &'static str {
...
+                event = "rotation_started",
+                rotation.target = TARGET_TOKEN_SIGNING_KEY,
+                "token signing material rotation started"
+            );
+            let result = async {
+                let certificate = tokio::fs::read_to_string(&certificate_path).await?;
+                let signing_key = tokio::fs::read_to_string(&signing_key_path).await?;
+                validate_signing_material(&certificate, &signing_key)
+                    .map_err(|err| color_eyre::eyre::eyre!("{err}"))?;
+                StoreProvisioningStrategy::filesystem(&certificate_path, &signing_key_path)
```
martcpp: provision() → persist_crypto_material persists cert then key sequentially (cert_manager.rs:741-743), so there's an observable window serving new cert + old key — deliverable 3 requires an atomic pair. Separately, this callback can now run concurrently with setup_cert_renewal_scheduler's cron calling the same provision(), with no mutex; two interleaved runs can persist a permanently mismatched pair. Needs a tokio::Mutex on CertManager at minimum, ideally the same atomic-holder treatment as ReloadingCertificateProvider.

src/setup.rs
```diff
@@ -120,6 +119,251 @@ fn classify_db_connect_error(err: &DbErr) -> &'static str {
...
+        color_eyre::eyre::eyre!(
+            "Failed to connect to database (kind={}, {})",
+            classify_db_connect_error(&err),
+            config.database.redacted_target()
+        )
+    })
+}
+
+#[cfg(any(feature = "sqlite", feature = "postgres", feature = "mysql"))]
+fn spawn_database_rotation(config: AppConfig, db: Arc<SwappableDatabaseConnection>) {
```
martcpp, line 166: The acceptance criteria name three test cases: successful rotation, malformed file reload, and atomic swap. Only malformed cert reload exists (cert.rs:236). Nothing exercises this function, SwappableDatabaseConnection::swap, or FileWatcher::spawn end-to-end. Minimum before merge: a postgres-tests integration test doing container → password_file → ALTER USER → rewrite file → assert in-flight queries survive, new pool is active, success metric increments; plus the wrong-password variant asserting the old pool is retained and neither password appears in the rendered error.

martcpp (changes requested):
certificate_chain() and signing_key_pem() are two separate awaits , each doing its own load_full(). A rotation landing between them produces a token signed with the new key carrying the old x5c. The ArcSwap in SigningMaterial makes this look atomic while the consumer still tears. Please add a single signing_material() method to the port returning both from one snapshot, and call it once here

Ngha-Boris:
@martcpp please recheck

martcpp:
> @martcpp please recheck

ci his failing fix that

Ngha-Boris:
Hello @martcpp recheck

martcpp (approved):

Hermann-Core (approved):

PR #470, opened 2026-09-02
feat: ci scheduled re scan

### LINKED ISSUE #413: CI: scheduled re-scan of the deployed image digest
## Summary

The image scan gate in `deploy.yml` runs only on push-to-main and `v*.*.*` tags. There is no scheduled scan of the currently-deployed artifact.

## Why this matters

`docs/supply-chain.md` argues that the image gate's one irreplaceable capability over
`cargo-audit` / `cargo-deny` is that **advisories published after merge trip the gate at
deploy time with no source change**. That is the correct argument for why the job exists at
all  source-level CI cannot cover a time axis.

The implementation does not deliver it. The default branch is `develop` and releases are cut
via release-plz, so the gate runs a handful of times a month. An advisory published the day
after a release sits undetected until someone cuts the next one — at which point it blocks a
release that may itself be the urgent fix for something else.

The supply-chain PR amends that doc section to stop claiming the capability and points here.
This issue is what makes the claim true again.

## Second thing this fixes: `expired_at` expiry lands on the release path

`.trivyignore.yaml` entries carry an `expired_at` date. Nothing warns before one lapses. With
scanning tied to releases, the first symptom of an expired exception is a **blocked production
deploy**, discovered by whoever is trying to ship.

A nightly scan surfaces the lapse days early, off the critical path. Without it, the expiry
mechanism  which is the whole point of the ledger discipline — is a landmine on the release
path.

## Third thing: this is the right home for SARIF

The supply-chain PR restricts SARIF upload to `refs/heads/main`, because code-scanning alerts
filed against tag refs do not surface in the branch view where anyone looks. That leaves
release scans with no durable sink  job summaries and artifacts both expire (90-day default).

A scheduled workflow runs on a branch ref, so alerts land somewhere real. Uploading from here
with `category: container-image` gives the deduplicated, history-preserving, dismissible
record that the release-triggered path cannot.

## Scope

One workflow. Deliberately not three issues  these are the same job, and split apart the
SARIF question gets relitigated standalone.

- `schedule:` trigger, plus `workflow_dispatch:`
- Resolve the digest currently deployed. Reading it from the live cluster is better than
  scanning `:latest`; scanning `:latest` is acceptable as a first cut and should be recorded
  as a known weakness if chosen.
- Run the same gate configuration as `deploy.yml`, so results are comparable
- On failure, open or update a single tracking issue rather than opening one per run
- Surface `.trivyignore.yaml` entries whose `expired_at` is within N days, before they lapse
- Upload SARIF with `category: container-image`

## Design questions for whoever picks this up

- **Cadence.** Nightly is the obvious default. Consider whether it should also run after the  Trivy DB refresh rather than on a fixed clock.
- **Alert fatigue.** A findings set that is stable across runs must not reopen an issue daily.
- **Cluster read access.** Reading the deployed digest needs cluster credentials in a scheduled workflow  a new credential surface. Weigh against scanning `:latest`.

## Acceptance criteria

- [ ] Scheduled workflow scans the deployed digest and reports findings
- [ ] Failure opens or updates a single tracking issue, not one per run
- [ ] `expired_at` entries approaching expiry are surfaced before they lapse
- [ ] SARIF uploaded with `category: container-image` from a branch ref
- [ ] `docs/supply-chain.md` "Relationship to Source-Level Checks" restored to claiming
      post-merge advisory detection, now accurately

ETA: 1 DAY

martcpp:
## Summary

The image scan gate in `deploy.yml` runs only on push-to-main and `v*.*.*` tags. There is no scheduled scan of the currently-deployed artifact.

## Why this matters

`docs/supply-chain.md` argues that the image gate's one irreplaceable capability over
`cargo-audit` / `cargo-deny` is that **advisories published after merge trip the gate at
deploy time with no source change**. That is the correct argument for why the job exists at
all  source-level CI cannot cover a time axis.

The implementation does not deliver it. The default branch is `develop` and releases are cut
via release-plz, so the gate runs a handful of times a month. An advisory published the day
after a release sits undetected until someone cuts the next one — at which point it blocks a
release that may itself be the urgent fix for something else.

The supply-chain PR amends that doc section to stop claiming the capability and points here.
This issue is what makes the claim true again.

## Second thing this fixes: `expired_at` expiry lands on the release path

`.trivyignore.yaml` entries carry an `expired_at` date. Nothing warns before one lapses. With
scanning tied to releases, the first symptom of an expired exception is a **blocked production
deploy**, discovered by whoever is trying to ship.

A nightly scan surfaces the lapse days early, off the critical path. Without it, the expiry
mechanism  which is the whole point of the ledger discipline — is a landmine on the release
path.

## Third thing: this is the right home for SARIF

The supply-chain PR restricts SARIF upload to `refs/heads/main`, because code-scanning alerts filed against tag refs do not surface in the branch view where anyone looks. That leaves release scans with no durable sink  job summaries and artifacts both expire (90-day default).

A scheduled workflow runs on a branch ref, so alerts land somewhere real. Uploading from here with `category: container-image` gives the deduplicated, history-preserving, dismissible record that the release-triggered path cannot.

## Acceptance criteria

- [ ] Scheduled workflow scans the deployed digest and reports findings
- [ ] Failure opens or updates a single tracking issue, not one per run
- [ ] `expired_at` entries approaching expiry are surfaced before they lapse
- [ ] SARIF uploaded with `category: container-image` from a branch ref
- [ ] `docs/supply-chain.md` "Relationship to Source-Level Checks" restored to claiming
      post-merge advisory detection, now accurately


.github/workflows/scheduled-image-scan.yml
```diff
@@ -0,0 +1,933 @@
...
+          # The expiry set is half the change key, not decoration. Keying the comment
+          # on advisories alone made the expiry warning silent in the exact state this
+          # workflow drives itself into: the issue is deliberately left open after
+          # findings clear, so once it exists, an exception entering the 30-day window
+          # on a clean night rewrote the body -- which notifies nobody -- and posted no
+          # comment, on a green run. That is the notice not being given.
+          #
+          # `id@date`, not the day count: the countdown changes every night and would
+          # post a comment every night, which is the alert fatigue this is avoiding.
+          # The set only changes when an entry enters the window or is re-argued.
```
Ngha-Boris, line 700: **Issue: Silent failure risk in SARIF processing loop**

The `jq` command inside the `while` loop could fail silently because the error occurs in a subshell created by the here-string redirection (`<<< "$files"`). With `set -euo pipefail`, errors in the subshell don't propagate to the parent shell.

**Suggestion:** Add explicit error handling:
```bash
while IFS= read -r file; do
    [ -n "$file" ] || continue
    base=$(basename "$file" .sarif)
    rest=${base#image-}
    variant=${rest%-*}
    arch=${rest##*-}
    
    # Capture jq output separately to detect failures
    if ! jq --arg tag "${variant}/${arch}" '
      .runs |= map(
        .results = ((.results // []) | map(
          .message.text = ((.message.text // "") + " (" + $tag + ")")
          | .locations = ((.locations // []) | map(
              .physicalLocation.artifactLocation.uri =
                ($tag + "/" + (.physicalLocation.artifactLocation.uri // "image"))
            ))
        ))
      )' "$file" > "tagged/${variant}-${arch}.sarif"; then
        echo "::error::Failed to process SARIF file: $file"
        exit 1
    fi
done <<< "$files"
```

This ensures that any `jq` failure (malformed input, disk full, etc.) fails the step explicitly rather than producing an incomplete merged SARIF.

.github/workflows/scheduled-image-scan.yml
```diff
@@ -0,0 +1,933 @@
...
+
+      - name: Verdict
+        # Last, and `always()`, so the SARIF, the summary and the issue are all in place
+        # before the run goes red. The red run is the notification; the issue is the
+        # record of findings specifically.
+        #
+        # Red is reserved for what this workflow promises, and the promise is about the
+        # image serving production. Findings surviving the ledger, the deployed variant
+        # not being scanned, and the expiry checker being broken are all failures of
+        # that promise. A *non-deployed* variant being unresolvable is not: it is worth
```
Ngha-Boris, line 850: **Issue: Fragile issue body parsing**

The `sed` pattern assumes the HTML comment format (`<!-- advisories: ... -->`) exists in the issue body. If someone manually edits the tracking issue and accidentally breaks this format, the extraction fails silently and may cause spurious "changed" notifications.

**Suggestion:** Add validation that the extraction succeeded:
```bash
previous=$(printf '%s' "$body" \
    | sed -n '/<!-- advisories: /{s/.*<!-- advisories: \(.*\) -->.*/\1/p;q;}')

# Validate extraction succeeded
if [ -z "$previous" ] && [ -n "$body" ]; then
    echo "::warning::Could not extract previous advisory set from issue body"
    # Consider this a change to be safe
    previous="__UNREADABLE__"
fi
```

This provides a fallback when the expected markers are missing, preventing the comparison logic from producing false positives.

Ngha-Boris (commented):
Hello @martcpp thank you for the PR the implementation is clean, can you please check the two comments below before the merge?

.github/workflows/scheduled-image-scan.yml
martcpp: Thanks  I dug into this, but I don't think the failure mode is reachable here.

A here-string is a redirection, not a pipeline, so `<<< "$files"` creates no
subshell; the loop body runs in the step's own shell. That's why the loop is
written with `<<<` rather than `find … | while`  the latter *would* fork, and
would have the problem you're describing.

.github/workflows/scheduled-image-scan.yml
martcpp: Good instinct, but I think this specific guard would cause the failure it's
meant to prevent, so I'd rather not take it as written.

`[ -z "$previous" ] && [ -n "$body" ]` can't distinguish "marker broken" from
"marker present, advisory set empty" — and the second is the state this workflow
deliberately parks itself in. The body writes `<!-- advisories: ${ADVISORIES} -->`,
and on a clean night `ADVISORIES` is empty, so the marker is
`<!-- advisories:  -->`, which the `sed` parses correctly to the empty string:

Ngha-Boris:
Thank you for replying to my comments
It is good to go for me. Please update the branch before merging

Ngha-Boris (dismissed):

Hermann-Core (approved):
LGTM

Ngha-Boris (approved):

PR #473, opened 2026-09-03
feat: update Helm chart for provider-neutral deployment and enhance configuration options

### LINKED ISSUE #393: Helm: remove provider-specific defaults and add values schema validation
## Objective

Make the Helm chart vendor-neutral and easier to validate by removing AWS/EKS-specific defaults, centralizing environment-specific values, and adding schema validation.

## Scope

- [x] Update `helm/chart/Chart.yaml` description so the chart targets Kubernetes generally, not only AWS EKS.
- [x] Move AWS-specific Service annotations, such as NLB annotations, out of default values and into AWS example overlays.
- [x] Default storage classes to the cluster default where possible, using `""` instead of hardcoded provider-specific names.
- [ ] Centralize hardcoded public domains under `global.domain` or an equivalent chart-wide value.
- [ ] Add `helm/chart/values.schema.json` for common chart values, including provider selectors and mutually constrained options where practical.
- [ ] Keep existing production examples available through explicit values overlays.

## Acceptance Criteria

- [ ] Default chart values render without AWS-specific Service annotations.
- [ ] Default chart values do not require a provider-specific storage class.
- [ ] Domain values are configured from one documented chart-level source.
- [ ] `helm lint` validates the chart against `values.schema.json`.
- [ ] Existing AWS/EKS deployment behavior remains available through an explicit AWS values example.

## Verification

- [ ] `helm lint ./helm/chart`
- [ ] `helm template ./helm/chart`
- [ ] `helm template ./helm/chart -f <aws-example-values>`
- [ ] Schema validation catches invalid provider values and obvious type errors.

Ngha-Boris:
This PR makes the Helm chart provider-neutral by default and moves AWS/EKS-specific behavior into explicit overlays.

**Changes**
- Updates `Chart.yaml` description to target Kubernetes generally instead of AWS EKS.
- Changes the default app image variant from AWS to the filesystem certificate-store variant.
- Removes AWS-specific defaults from `values.yaml`, including:
  - AWS credential mounting by default
  - default AWS SecretStore/ExternalSecret enablement
  - Route53 DNS provider env default
  - hardcoded public domain values
  - provider-specific `high-performance` storage class defaults
- Adds `global.domain` as the chart-wide source for public domain configuration.
- Updates ingress rendering to derive:
  - `statuslist.<global.domain>` for the default host
  - `*.<global.domain>` for default TLS hosts
- Adds configurable Service annotations.
- Adds `values-aws.yaml` as an explicit AWS/EKS overlay, including NLB annotations, AWS image selection, Route53 DNS, Secrets Manager, AWS credentials mount, and AWS storage class.
- Updates production values to explicitly preserve existing AWS deployment behavior.
- Expands `values.schema.json` to validate common chart values, provider selectors, service types/ports, domain format, image digest format, and related constraints.
- Updates Helm documentation to reflect the new neutral defaults and explicit AWS overlay.

**Verification**
- `helm lint ./helm/chart`
- `helm template ./helm/chart`
- `helm template ./helm/chart -f helm/chart/values-aws.yaml`
- `helm template ./helm/chart -f helm/chart/values-production.yaml`
- Confirmed schema validation rejects invalid provider values, bad service port types, and malformed `global.domain` values.


Hermann-Core:
@martcpp @forkimenjeckayang can you check this ticket? we need it to be closed today please, thanks.

helm/chart/templates/deployment.yaml
```diff
@@ -160,8 +160,9 @@ spec:
               {{- fail "statuslist.env.APP_DATABASE__PORT is required; set it to the database service port" }}
             {{- end }}
             {{- if not (hasKey $env "APP_SERVER__DOMAIN") }}
-              {{- if .Values.statuslist.ingress.externalDnsHostname }}
-                {{- $_ := set $env "APP_SERVER__DOMAIN" .Values.statuslist.ingress.externalDnsHostname }}
+              {{- $globalDomain := trimSuffix "." (.Values.global.domain | default "") }}
```
Hermann-Core: Accessing `.Values.global.domain` directly fails with a nil pointer evaluation error when `.Values.global` is omitted or set to `null` (such as when the chart is deployed as a subchart or with `--set global=null`):

```bash
Error: status-list-server/templates/deployment.yaml:163:58: executing "..." at <.Values.global.domain>: nil pointer evaluating interface {}.domain
```

Use `dig` for safe traversal:

```yaml
{{- $globalDomain := trimSuffix "." (.Values | dig "global" "domain" "" | default "") }}
```

helm/chart/templates/ingress.yaml
```diff
@@ -1,4 +1,16 @@
 {{- if .Values.statuslist.ingress.enabled }}
+{{- $globalDomain := trimSuffix "." (.Values.global.domain | default "") }}
```
Hermann-Core: Same nil pointer evaluation risk here as in `deployment.yaml`: if `.Values.global` is nil, `.Values.global.domain` causes template execution to fail.

Please use `dig` for safe traversal:

```yaml
{{- $globalDomain := trimSuffix "." (.Values | dig "global" "domain" "" | default "") }}
```

helm/chart/templates/ingress.yaml
```diff
@@ -1,4 +1,16 @@
...
+{{- $ingressHost := .Values.statuslist.ingress.externalDnsHostname | default "" }}
+{{- if and (eq $ingressHost "") $globalDomain }}
+{{- $ingressHost = printf "statuslist.%s" $globalDomain }}
+{{- end }}
+{{- if eq $ingressHost "" }}
+{{- $ingressHost = "localhost" }}
+{{- end }}
+{{- $tlsHosts := .Values.statuslist.ingress.tls.hosts | default list }}
```
Hermann-Core: When `global.domain` is empty but an operator configures `statuslist.ingress.externalDnsHostname` (e.g. `status.example.com`), `$tlsHosts` remains empty `[]`. Because lines 395–403 guard on `gt (len $tlsHosts) 0`, the `spec.tls` block is omitted entirely despite TLS redirect annotations being enabled by default.

Consider falling back to `$ingressHost` when `tls.hosts` and `global.domain` are unset:

```yaml
{{- $tlsHosts := .Values.statuslist.ingress.tls.hosts | default list }}
{{- if and (eq (len $tlsHosts) 0) $globalDomain }}
{{- $tlsHosts = list (printf "*.%s" $globalDomain) }}
{{- else if and (eq (len $tlsHosts) 0) (ne $ingressHost "localhost") }}
{{- $tlsHosts = list $ingressHost }}
{{- end }}
```

Hermann-Core (changes requested):
@Ngha-Boris I have posted review comments on the templates for your attention.

helm/chart/Chart.yaml
```diff
@@ -1,10 +1,12 @@
...
-description: Helm chart for deploying status list server on AWS EKS
+description: Helm chart for deploying status list server on Kubernetes
 type: application
 version: 0.3.0
 # Default image tag when statuslist.image.tag and .digest are both empty.
-# Release images are always suffixed; the chart default uses the AWS variant.
+# Release images are always suffixed; the release workflow publishes the AWS
+# variant as the default image tag. Provider-specific Kubernetes settings stay
+# in explicit values overlays.
 appVersion: "1.0.1-aws"
```
forkimenjeckayang: The base chart still resolves to the `1.0.1-aws` image. That build contains the AWS Secrets Manager and Route53 paths; with `APP_ENV=production` and no DNS provider value, the application defaults to Route53. This leaves provider-specific executable behavior in the supposedly neutral default while AWS credential mounting is disabled, and it contradicts the PR description's stated `-fscert` default. Please make the base select and configure a runnable provider-neutral variant, or require an explicit image variant, and keep `-aws` only in the AWS/production overlays.

helm/chart/values.yaml
```diff
@@ -337,15 +339,15 @@ postgres:
...
+    storageClass: ""
   service:
     type: ClusterIP
     port: 5432
   image:
     tag: "18.0"
 
 externalSecret:
-  enabled: true
+  enabled: false
```
forkimenjeckayang, line 369: Disabling ESO here leaves the default without any source for `statuslist-secret`: the fallback Secret is also disabled, but the Deployment mounts this Secret and bundled PostgreSQL reads `postgres-password` from it. The documented plain `helm install` therefore renders successfully but its Pods cannot start unless the operator has already created the fixed-name Secret. Please make the existing-Secret path an explicit prerequisite/choice, or make the default self-contained without shipping a real password. CI should also verify that each default non-optional Secret reference has a rendered creator or an explicit existing-Secret mode.

helm/chart/values-aws.yaml
```diff
@@ -0,0 +1,39 @@
...
+  image:
+    tag: "1.0.1-aws"
+  service:
+    # Enable when exposing the service directly through an AWS Network Load Balancer.
+    # Keep ClusterIP when ingress is responsible for external traffic.
+    type: LoadBalancer
+    annotations:
+      service.beta.kubernetes.io/aws-load-balancer-type: "external"
+      service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: "ip"
+      service.beta.kubernetes.io/aws-load-balancer-scheme: "internet-facing"
```
forkimenjeckayang: The comment says direct NLB exposure is optional, but this overlay enables an `internet-facing` LoadBalancer while the inherited NGINX Ingress remains enabled. The documented AWS command therefore creates a second public path directly to the application's HTTP port, bypassing the Ingress TLS layer. Please keep this general AWS overlay on `ClusterIP`, and put direct NLB exposure in a separate explicit overlay (or otherwise make the two exposure modes mutually exclusive and configure TLS/security for the selected path). Reference: [AWS Load Balancer Controller NLB guide](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/nlb/).

helm/chart/values.schema.json
```diff
@@ -17,7 +31,85 @@
...
+              "type": ["string", "null"],
+              "pattern": "^$|^sha256:[a-f0-9]{64}$"
+            }
+          }
+        },
+        "service": {
+          "type": "object",
+          "properties": {
+            "enabled": { "type": "boolean" },
+            "type": { "type": "string", "enum": ["ClusterIP", "NodePort", "LoadBalancer", "ExternalName"] },
```
forkimenjeckayang: The schema now accepts `ExternalName`, but the Service template has no `externalName` value and always renders the application's selector. This cannot represent an ExternalName Service. Please remove `ExternalName` from the enum, or add a required `externalName` setting and render the ExternalName-specific spec without the Pod selector. Reference: [Kubernetes ExternalName Service](https://kubernetes.io/docs/concepts/services-networking/service/#externalname).

helm/chart/templates/_helpers.tpl
```diff
@@ -36,8 +36,9 @@ Common labels
 {{- define "status-list-server-chart.labels" -}}
 helm.sh/chart: {{ include "status-list-server-chart.chart" . }}
 {{ include "status-list-server-chart.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- $appVersion := .Values.statuslist.image.tag | default .Chart.AppVersion }}
+{{- if $appVersion }}
+app.kubernetes.io/version: {{ $appVersion | quote }}
```
forkimenjeckayang: `statuslist.image.tag` is now copied verbatim into a Kubernetes label, but the schema does not enforce the label's 63-character and start/end rules. For example, `_foo` renders successfully here but is invalid as a label value. Please sanitize/truncate the value used only for this label, keep the chart appVersion as the label, or constrain the tag schema to the label rules. Reference: [Kubernetes label syntax](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

helm/chart/values.schema.json
```diff
@@ -2,6 +2,20 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "properties": {
+    "global": {
+      "type": "object",
+      "properties": {
+        "domain": {
+          "type": "string",
+          "description": "Chart-wide public DNS suffix. Leave empty to render localhost/default-neutral hostnames.",
+          "pattern": "^$|^([A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?\\.)+[A-Za-z]{2,63}\\.?$"
```
forkimenjeckayang: This schema is intended to reject malformed domains, but `[A-Za-z]` allows uppercase values. `--set global.domain=Example.COM` passes Helm validation and renders uppercase Ingress/TLS hosts, which Kubernetes rejects under its DNS host validation. Please restrict the pattern to lowercase DNS labels (or normalize before rendering) and cover the accepted value with an API-schema validation test. Reference: [Kubernetes IngressRule host](https://kubernetes.io/docs/reference/kubernetes-api/networking/ingress-v1/#IngressRule).

helm/README.md
```diff
@@ -4,7 +4,7 @@ This guide provides instructions for deploying the Status List Server using the
 
 ## Prerequisites
 
-- Kubernetes cluster (e.g., AWS EKS)
+- Kubernetes cluster
 - Helm 3 installed
 - `kubectl` configured to connect to your cluster
 - External Secrets Operator (ESO) installed when `externalSecret.enabled=true` (the default). The chart renders `external-secrets.io/v1` `ExternalSecret`, `SecretStore`, and `ClusterSecretStore` references, so the cluster CRDs must serve `external-secrets.io/v1` before installing or upgrading this chart. ESO v0.16 promoted these resources to `v1`; upgrade the ESO controller and CRDs together, and if CRDs are managed separately, apply the matching CRD bundle before upgrading the operator.
```
forkimenjeckayang: still says ESO is installed when `externalSecret.enabled=true` “(the default),” but this PR changes the default to `false`.

forkimenjeckayang (dismissed):
a few coments please check

helm/chart/values.yaml
Hermann-Core: Confirmed. With `externalSecret.enabled: false`, `secretStore.enabled: false`, and `fallbackSecret.enabled: false`, a default `helm install` emits no Secret resource. The Deployment mounts `statuslist-secret` and PostgreSQL configures `existingSecret: statuslist-secret`, which immediately triggers `CreateContainerConfigError`. Either `fallbackSecret.enabled` should be `true` with a local default, or `values.yaml` and the docs must explicitly document creating `statuslist-secret` as a prerequisite.

helm/chart/values-aws.yaml
Hermann-Core: Agreed. When `values-aws.yaml` sets `service.type: LoadBalancer` with internet-facing NLB annotations without disabling `ingress.enabled`, it creates an unencrypted public entry point directly to port 8081 alongside the Ingress. `values-aws.yaml` should keep `service.type: ClusterIP` so the Ingress handles traffic, or explicitly disable Ingress if direct NLB is chosen.

helm/chart/values.schema.json
Hermann-Core: Confirmed. Kubernetes Ingress host validation strictly enforces RFC 1123 (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`). Allowing uppercase letters in `global.domain` passes Helm schema checks but causes `kubectl apply` to fail with a validation error on `spec.rules[0].host`.

Ngha-Boris:
Hello @Hermann-Core @forkimenjeckayang please rescheck

helm/chart/templates/ingress.yaml
```diff
@@ -31,10 +41,10 @@ spec:
             name: {{ include "status-list-server-chart.fullname" . }}-service
             port:
```
martcpp, line 45: This drops the entire tls: block on the default render. tls.hosts is now [] and global.domain is "", so $tlsHosts is empty  but tls.secretName: statuslist-tls, cert-manager.io/cluster-issuer, and force-ssl-redirect: "true" all still render. Result: cert-manager never issues a cert (ingress-shim needs the tls: block), and nginx redirects everything to HTTPS to serve its self-signed default.

helm/chart/values.yaml
```diff
@@ -62,17 +68,18 @@ statuslist:
...
+    externalDnsHostname: ""
   env:
     RUST_LOG: "info"
     APP_ENV: "production"
     APP_SERVER__HOST: "0.0.0.0"
     APP_SERVER__PORT: "8000"
-    APP_SERVER__CERT__EMAIL: "info@adorsys.com"
-    APP_SERVER__CERT__ORGANIZATION: "adorsys GmbH & CO KG"
+    APP_SERVER__CERT__EMAIL: ""
+    APP_SERVER__CERT__ORGANIZATION: ""
```
martcpp: This doesn't unset these  it overrides working defaults with invalid values. config.rs:1071 defaults server.cert.email to admin@example.com and config.rs:1073 defaults organization; deployment.yaml emits every $env key unconditionally, so the pod gets APP_SERVER__CERT__EMAIL="". CertConfig.email is a bare String with no validation (config.rs:186), so ACME registration fails at runtime with invalidContact long after a green deploy.

.github/workflows/render-helm-templates/action.yml
martcpp, line 1: These assertions lock in the broken default from values.yaml:350  CI now proves the chart creates no secret source and calls that success. helm template can't see a dangling secret reference, so this gate stays green on a chart that won't start.

helm/chart/values.yaml
```diff
@@ -1,5 +1,10 @@
 global:
-  storageClass: "high-performance"
+  # Chart-wide public DNS suffix. When set, the default ingress host becomes
+  # statuslist.<domain> and the default TLS host becomes *.<domain>.
+  # Leave empty for provider-neutral defaults.
+  domain: ""
+  # Empty string lets Kubernetes use the cluster default StorageClass.
+  storageClass: ""
```
martcpp: global.storageClass is dead. No template in this chart reads it, and the vendored postgres-0.8.2 subchart touches .Values.global only to pass it to common.image for registry resolution  it reads .Values.persistence.storageClass. Setting it does nothing.

helm/chart/Chart.yaml
```diff
@@ -1,11 +1,13 @@
...
-description: Helm chart for deploying status list server on AWS EKS
+description: Helm chart for deploying status list server on Kubernetes
 type: application
 version: 0.3.0
 # Default image tag when statuslist.image.tag and .digest are both empty.
-# Release images are always suffixed; the chart default uses the AWS variant.
-appVersion: "1.0.1-aws"
+# Release images are always suffixed; the chart default uses the filesystem
+# certificate-store variant so base installs are provider-neutral. Cloud-specific
+# variants are selected in explicit values overlays.
```
martcpp, line 9: appVersion: "1.0.1-aws" plus empty image.tag means helm install ./helm/chart with no overlay pulls the AWS variant image onto any cluster, and the _helpers.tpl change stamps app.kubernetes.io/version: "1.0.1-aws" on every object. The one thing that decides which cloud SDK is in the running binary is still AWS-pinned.

martcpp (changes requested):

helm/chart/templates/ingress.yaml
Hermann-Core: Confirmed. Because `global.domain` is empty and `tls.hosts` defaults to `[]`, `spec.tls` is omitted entirely while the Ingress annotations still include `cert-manager.io/cluster-issuer: letsencrypt-prod` and `nginx.ingress.kubernetes.io/ssl-redirect: "true"`. Ingress-shim ignores Ingress resources lacking `spec.tls`, leaving NGINX to redirect HTTP requests to HTTPS and serve its fake default certificate. Either `statuslist.ingress.enabled` should default to `false`, or the SSL-redirect and cluster-issuer annotations should be conditional on TLS being configured.

helm/chart/values.yaml
Hermann-Core: Confirmed. In `src/config.rs`, `CertConfig` defaults `email` to `admin@example.com` and `organization` to `Example Org`. Setting `APP_SERVER__CERT__EMAIL: ""` in `values.yaml` causes `deployment.yaml` to pass an empty env var, overriding the Rust defaults with `""`. At runtime, ACME account registration fails with `invalidContact` because Let's Encrypt rejects empty contact emails. These two keys should be removed from `values.yaml:env` so the application defaults apply unless explicitly set in an overlay.

.github/workflows/render-helm-templates/action.yml
Hermann-Core: Note: in commit `3291848`, Boris enabled `statuslist.fallbackSecret.enabled: true` with a generated 32-character password in `values.yaml:143-147`, and updated `action.yml` (lines 95–106) to assert that `statuslist-secret` IS rendered by default and referenced by both Deployment and PostgreSQL. The default render now produces a self-contained Secret resource.

helm/chart/values.yaml
Hermann-Core: Confirmed. Neither this chart's templates nor the vendored `postgres` subchart read `.Values.global.storageClass`. PostgreSQL strictly reads `.Values.persistence.storageClass`. Setting `global.storageClass` in `values.yaml` is dead configuration unless wired up.

helm/chart/Chart.yaml
Hermann-Core: In commit `3291848`, Boris updated `Chart.yaml` to `appVersion: "1.0.1-fscert"`. However, there is a remaining runtime conflict: `values.yaml` leaves `APP_SERVER__CERT__PROVISIONING_STRATEGY` unset (defaulting to `acme`) and `APP_SERVER__CERT__DNS__PROVIDER` unset (defaulting to `Route53` in production). Because `-fscert` does not include cloud provider SDKs and AWS credentials are not mounted by default, an out-of-the-box install will fail at boot when trying to initialize Route53 ACME challenges. The base defaults must align the image variant with a runnable provisioning strategy (e.g. strategy `store`).

Ngha-Boris:

helm/chart/Chart.yaml
```diff
@@ -1,11 +1,13 @@
 apiVersion: v2
 name: status-list-server
-description: Helm chart for deploying status list server on AWS EKS
+description: Helm chart for deploying status list server on Kubernetes
 type: application
 version: 0.3.0
```
martcpp: still version: 0.3.0. fallbackSecret now defaults on, appVersion changed variant, three ingress annotations disappeared. Consumers get all of that on an unchanged chart version

martcpp (changes requested):
lookup + randAlphaNum breaks GitOps. Needs at minimum a documented caveat.
lookup returns empty under helm template, which is exactly how ArgoCD and Flux render. Every sync generates a fresh 32-char password while postgres keeps the old one in its PVC → auth failure and permanent diff noise. helm upgrade from a CLI does a live lookup, so the interactive path is fine; the GitOps path is not. secret.yaml carries no helm.sh/resource-policy: keep. Add that annotation and a line in the README telling GitOps users to set an explicit password or use ESO.

 values-aws.yaml and new values-aws-nlb.yaml are rendered and asserted; schema negative tests added. helm lint still runs nowhere.||

helm/chart/values-external-secrets.yaml:11 is broken. Fix before merge.
It sets externalSecret.enabled: true and never sets fallbackSecret.enabled: false. With the new default of true, deployment.yaml:2-4 hard-fails: "externalSecret.enabled and statuslist.fallbackSecret.enabled cannot both be true". Every other ESO overlay got the flag (values-aws.yaml:27, values-production.yaml:33, values-rotation-example.yaml:98) this one was missed. It's rendered by no CI job, so nothing catches it. One line, plus add it to the render action.

@Ngha-Boris  fix this minus change thank for your works so far

Ngha-Boris:
@martcpp please check again

src/config.rs
```diff
@@ -1974,10 +1974,7 @@ mod tests {
         }
         if let Some(db_url) = default_config.database.url.as_ref() {
             let db_url = db_url.expose_secret();
-            assert!(
-                !db_url.contains("test_data"),
-                "Default config database URL references test_data: {db_url}"
-            );
+            assert!(!db_url.contains("test_data"));
```
martcpp: config.rs still has the bare assert!(!db_url.contains("test_data")) with the diagnostic message stripped, while the four sibling assertions around it kept theirs. add a message here but don't expose the data

martcpp (changes requested):
values.schema.json domain pattern is lowercase-only, so Example.COM is rejected despite being valid DNS, and action.yml:331 now enshrines that in a test. Normalize with lower rather than reject.
alues-aws.yaml:9 and values-production.yaml:13 still pin tag: "1.0.1-aws" with nothing in the release process to bump them; the two files still duplicate the AWS block.

Ngha-Boris:
Hello @martcpp please take a look

docs/troubleshooting.md
```diff
@@ -560,14 +560,11 @@ and `curl` against the service/NodePort returns `000`/empty reply. The applicati
...
 
-**Root cause:** The container image is built with the `acme` Cargo feature, and under that feature
-the **default `server.cert.provisioning_strategy` is `acme`** (`src/utils/cert_manager/builder.rs:189-190`). With
-strategy `acme`, the application attempts ACME DNS-01 certificate provisioning at startup, which
-requires a DNS provider and credentials (see `dns-providers.md`). `values-local.yaml` and
-`LOCAL_DEPLOYMENT.md` do not override the cert values, so they inherit the production defaults
-from `values.yaml`: `APP_SERVER__CERT__DNS__PROVIDER=route53`,
-`APP_SERVER__CERT__EMAIL=info@adorsys.com`, `APP_SERVER__DOMAIN=statuslist.eudi-adorsys.com`, and
-the Let's Encrypt directory URL, and block before binding the HTTP server.
+**Root cause:** ACME-enabled image variants use `server.cert.provisioning_strategy=acme` by
```
Hermann-Core: This section references `server.cert.provisioning_strategy=acme`, which is an internal Rust trait detail rather than an application configuration property (`src/config.rs` has no such setting).

Describe the behavior generically instead: ACME-enabled image variants perform automated DNS-01 certificate issuance at startup, requiring a DNS provider and valid credentials. Without those credentials, the container blocks before binding the HTTP port. Replicas using static certificate images require certificate and private key files mounted into the container.

Please also update lines 577 to 586 and line 822. Instructions like `grep -i provisioning` or setting `APP_SERVER__CERT__PROVISIONING_STRATEGY=store` refer to non-existent environment variables.

docs/deployment-runbook.md
```diff
@@ -48,37 +48,29 @@ kubectl config use-context minikube
...
-> **Certificate caveat for local runs:** the published images are built with the `acme` feature, so
-> the application's default `server.cert.provisioning_strategy` is `acme` and it attempts ACME
-> DNS-01 at startup. `values-local.yaml` leaves the production cert env (`route53` DNS provider,
-> Let's Encrypt URL, `statuslist.eudi-adorsys.com` domain) in place, which blocks the HTTP server
-> from binding. For a local run, disable or retarget certificate provisioning (e.g. strategy
-> `store` with a self-signed cert/key mounted via `statuslist.secretMounts`), see the entry "Pod
-> `Running` but never binds the HTTP port" in `troubleshooting.md`. If the pod crash-loops with a
-> database pool timeout while PostgreSQL is healthy, see the "pool timed out while waiting for an
-> open connection ... while PostgreSQL is reachable" entry in `troubleshooting.md`.
+> **Certificate caveat for local runs:** the provider-neutral `-fscert` image expects certificate
```
Hermann-Core: Avoid referencing internal provisioning strategy names like `store` here. The `-fscert` image variant is compiled without ACME and always expects static certificate and signing key files.

State this directly: the default `-fscert` image requires certificate and signing key files mounted into the pod. For local runs, mount test credentials via `statuslist.secretMounts` or use an image variant tailored to your environment.

docs/LOCAL_DEPLOYMENT.md
```diff
@@ -15,35 +15,30 @@ minikube start
...
-  --set statuslist.image.tag=latest-aws --set statuslist.image.digest=null
+helm install statuslist-local ./helm/chart -n local -f ./helm/chart/values-local.yaml
 ```
 
-> **Certificates:** the image is built with the `acme` feature, so the app tries ACME DNS-01 at
-> startup unless disabled. `values-local.yaml` inherits the production cert values
-> (`APP_SERVER__CERT__DNS__PROVIDER=route53`, Let's Encrypt URL, adorsys domain) from
-> `values.yaml`, which blocks the HTTP server from binding. Where `--set-string` flags can't fully
-> express the file mounts, use a values override file (see `docs/troubleshooting.md` → "Pod Running
+> **Certificates:** the default `-fscert` image is provider-neutral and expects file-backed signing
```
Hermann-Core: This note refers to `when using the store provisioning strategy`, which leaks internal implementation details. The `-fscert` image always requires static certificate and key files.

In addition, following step 4 with `values-local.yaml` will cause the server to fail at startup with `static certificate mode requires both certificate and signing key`, because `values-local.yaml` does not configure certificate file paths or mounts. Please update `values-local.yaml` or provide a ready-to-use local configuration so that local deployments succeed without immediate container restarts.

helm/README.md
```diff
@@ -118,20 +146,17 @@ statuslist:
     APP_SERVER__DOMAIN: "statuslist.example.com"
 ```
 
-With ACME or a cloud secret backend, the signing material is provisioned by that backend. In that
-case configure `APP_SERVER__DOMAIN` and the backend credentials instead.
+With ACME or a cloud secret backend, configure `APP_SERVER__DOMAIN`, the DNS provider, and the backend credentials instead.
```
Hermann-Core, line 149: In the example above at line 145, `APP_SERVER__CERT__PROVISIONING_STRATEGY: "store"` is passed under `env:`.

This variable does not exist in `AppConfig` or `src/config.rs`. The application relies entirely on `APP_SERVER__CERT__STORE__CERTIFICATE_PATH` and `APP_SERVER__CERT__STORE__SIGNING_KEY_PATH` to load file-based certificates. Please remove `APP_SERVER__CERT__PROVISIONING_STRATEGY` to avoid documenting unused configuration keys.

helm/chart/values-rotation-example.yaml
```diff
@@ -171,10 +177,11 @@ statuslist:
 
   ingress:
     enabled: true
-    externalDnsHostname: statuslist.example.com
+    # Defaults to statuslist.<global.domain>.
```
Hermann-Core, line 178: Note that around line 166 above, `APP_SERVER__CERT__PROVISIONING_STRATEGY: "store"` is defined in `env`. This variable is not mapped to any field in `AppConfig` and has no effect on certificate provisioning. Removing it keeps the rotation example focused on active configuration.

helm/chart/templates/ingress.yaml
```diff
@@ -31,10 +41,10 @@ spec:
...
               number: {{ .Values.statuslist.service.port }}
-  {{- if .Values.statuslist.ingress.tls.secretName }}
+  {{- if and .Values.statuslist.ingress.tls.secretName (gt (len $tlsHosts) 0) }}
```
Hermann-Core, line 47: When `global.domain` is empty and `statuslist.ingress.tls.hosts` is left unset, `$tlsHosts` evaluates to empty (`[]`). Because of `(gt (len $tlsHosts) 0)`, the entire `tls:` block is omitted, even when an operator provides `statuslist.ingress.tls.secretName`.

If `secretName` is set, consider defaulting `$tlsHosts` to `list $ingressHost` even when `$ingressHost` is `localhost`, or retaining the secret name so ingress controllers can bind certificates properly.

helm/chart/templates/secret.yaml
```diff
@@ -1,7 +1,21 @@
...
-{{- fail "statuslist.fallbackSecret is enabled but 'postgres-password' is empty. Provide a non-empty password in statuslist.fallbackSecret.stringData.postgres-password, otherwise the application would be configured with an invalid database credential." }}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace "statuslist-secret" }}
+{{- $existingPg := "" }}
+{{- if and $existing $existing.data }}
+{{- $existingPg = index $existing.data "postgres-password" | default "" }}
+{{- end }}
+{{- if $existingPg }}
+{{- $pg = b64dec $existingPg }}
+{{- else }}
+{{- $pg = randAlphaNum 32 }}
```
Hermann-Core, line 17: Adding `helm.sh/resource-policy: keep` preserves the Secret during Helm CLI upgrades.

For GitOps tools like Argo CD and Flux, client-side template rendering runs without cluster API access, meaning `lookup` always evaluates to empty. Every sync cycle generates a new random 32-character password, creating ongoing drift against existing database storage. Consider adding a clear inline comment above `randAlphaNum 32` noting that GitOps pipelines should set an explicit password or disable `fallbackSecret`.

helm/chart/values.yaml
```diff
@@ -138,8 +133,10 @@ statuslist:
   # (postgres.auth.existingSecret) references the same name, so it is deliberately not
   # independently configurable.
   fallbackSecret:
-    enabled: false
+    enabled: true
     stringData:
+      # Empty means the chart generates a random password. Helm upgrades reuse
```
Hermann-Core, line 158: Note in this comment that GitOps deployments (such as Argo CD or Flux) run `helm template` client-side where `lookup` cannot read the cluster. Leaving `postgres-password` empty will produce a new random password on every reconciliation. Users deploying via GitOps should provide an explicit password or set `fallbackSecret.enabled: false`.

src/config.rs
Hermann-Core: Add a descriptive assertion message here without exposing secret contents, for example:
```rust
assert!(
    !db_url.contains("test_data"),
    "Default config database URL references test_data"
);
```
This aligns with the other assertions in this test block while keeping sensitive values out of CI logs.

.github/workflows/render-helm-templates/action.yml
```diff
@@ -253,27 +336,50 @@ runs:
         if helm template status-list-server helm/chart --set secretStore.provider=raw >/dev/null 2>&1; then
           echo "ERROR: provider=raw with empty secretStore.raw should fail"; exit 1
         fi
+        if helm template status-list-server helm/chart --set statuslist.service.type=ExternalName >/dev/null 2>&1; then
+          echo "ERROR: statuslist.service.type=ExternalName should fail because the chart does not render ExternalName Services"; exit 1
+        fi
+        if helm template status-list-server helm/chart --set-string global.domain=Example.COM >/dev/null 2>&1; then
+          echo "ERROR: uppercase global.domain should fail schema validation"; exit 1
```
Hermann-Core: This test expects uppercase domain names to fail schema validation. Because DNS domain names are case-insensitive per RFC 1035 and RFC 1123, converting domain values to lowercase in templates with `lower` offers a cleaner experience than rejecting valid DNS inputs outright.

Hermann-Core (changes requested):
@Ngha-Boris I reviewed the latest updates on provider-neutral deployment and external secrets management.

The chart structure, lint checks, and test coverage for external secrets show good progress. Before this is ready for production, there are a few drifts, documentation leaks, and template edge cases to address:

1. Remove leaked implementation details and non-existent configuration keys:
   Avoid referencing internal Rust builder traits like `server.cert.provisioning_strategy=acme` in `docs/troubleshooting.md`. Describe the certificate mode generically instead (ACME automated issuance vs filesystem static certificate loading).
   Clean up references to non-existent environment variables like `APP_SERVER__CERT__PROVISIONING_STRATEGY` across `helm/README.md`, `values-rotation-example.yaml`, and `docs/troubleshooting.md`.
2. Prevent startup failures in local deployment:
   The default chart image is now `1.0.1-fscert`, which requires static certificate and key files. `values-local.yaml` currently lacks certificate configuration, which causes new local deployments to fail startup immediately.
3. Handle Ingress TLS without global domain:
   In `templates/ingress.yaml`, when `global.domain` is empty, providing `tls.secretName` still drops the `tls:` block because `$tlsHosts` remains empty.
4. GitOps compatibility for secret generation:
   Document the client-side behavior of `randAlphaNum 32` under Argo CD and Flux, where `lookup` returns empty and generates fresh passwords on every sync.
5. Humanize documentation across all modified files:
   Keep technical descriptions plain, direct, and free of internal jargon or filler.

Detailed inline comments with suggestions have been provided below.

martcpp (dismissed):

Ngha-Boris:
Hello @Hermann-Core @forkimenjeckayang please recheck

martcpp (approved):

Hermann-Core (approved):
@Ngha-Boris I re-checked the latest updates, all previous review points have been addressed. Thank you

PR #493, opened 2026-09-04
fix(ci): fetch repo and pass repository flag for digest artifact download

Hermann-Core:
### Summary

- Adds `actions/checkout` step to the `promote-tags` job in `.github/workflows/deploy.yml` so that a git repository context is present.
- Explicitly adds `-R "$GITHUB_REPOSITORY"` to `gh run download` in both `deploy.yml` and `scheduled-image-scan.yml` to ensure artifact downloading succeeds even without local git repo inference.
- Retains existing semver tag formatting (without `v` prefix) as configured.

### Root Cause
During release tag promotion, the `promote-tags` job failed at `gh run download` with `fatal: not a git repository (or any of the parent directories): .git` because the repo was not checked out and no repository target was specified to the GitHub CLI.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -256,7 +256,7 @@ jobs:
       - name: Download built digest
         env:
           GH_TOKEN: ${{ github.token }}
-        run: gh run download "$GITHUB_RUN_ID" --name "image-digest-${{ matrix.variant.suffix }}" --dir .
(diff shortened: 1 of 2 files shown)


Hermann-Core:
@martcpp @Ngha-Boris @ndefokou @Christiantyemele can you approve?

martcpp (approved):

Christiantyemele (approved):

PR #512, opened 2026-09-09
fix(helm): align chart monitoring with shared kube-prometheus-stack

ndefokou:
## Summary
Aligns the Helm chart's monitoring resources with the cluster-wide
`kube-prometheus-stack` shared platform (installed/owned by `wallet-eks-env`), so
Status List can be the first reliable onboarding candidate for that stack.

## What changed
- **Namespace-scoped PromQL** in `prometheusrule.yaml`
  - Every raw selector now pins `namespace="<release namespace>"`, e.g.
    `{otel_scope_name="status-list-server",namespace=…}`.
  - Every alert expression that reads a recording series now adds a `namespace`
    matcher (e.g. `sli:error_budget:success:30d{namespace=…}`).
  - `StatusListMetricsAbsent` is no longer cluster-global: it is scoped to the
    release namespace, so removing one deployment can no longer keep another
    deployment's absence alert silent.
  - Fixes cross-namespace evaluation and overlapping/wrongly-labelled recording
    series when the chart is installed in more than one namespace.
- **Dashboard provisioning**
  - New `dashboard-configmap.yaml` publishes the generated SLO dashboard as a
    ConfigMap labelled `grafana_dashboard: "1"`, auto-provisioned by the shared
    Grafana sidecar (`searchNamespace: ALL`).
  - The generated `status-list-slo.json` is vendored under
    `helm/chart/observability/dashboards/generated/` so Helm can embed it via
    `.Files.Get`. Enabled in production values.
- **Prometheus annotations made opt-in**
  - `prometheus.io/scrape|port|path` Pod annotations now default to **off**
    (`statuslist.prometheusAnnotations: false`) because the `ServiceMonitor` is the
    authoritative discovery method in kube-prometheus-stack. Enable only for a
    separately configured scraper that relies on annotations.
- **Documentation**
  - `values.yaml` documents the cluster dependencies: CRDs must exist before this
    release on a fresh cluster, and the shared Prometheus must retain **>30 days**
    to represent the 30d SLO windows.
  - Updated `alerting`/`prometheusRule`/`grafanaDashboard` value docs.

## Testing
- `helm lint` passes for all chart variants (default, aws, external-secrets, aws+prod).
- `helm template` render tests added/verified for:
  - namespace-scoped rule selectors and absence alert,
  - the Grafana dashboard ConfigMap (label + embedded JSON), and
  - the opt-in `prometheus.io/*` annotations.

## Notes / follow-ups
- The shared stack (`wallet-eks-env`) serves the `AlertmanagerConfig` CRD as
  `monitoring.coreos.com/v1alpha1` only. Any `AlertmanagerConfig` work (currently on
  the separate webhook-alerting branch `446-…`) must target `v1alpha1`, prefer native
  receivers (e.g. Teams v2 `msteamsv2Configs`), and keep `mattermost` as a generic
  webhook adapter since the pinned CRD has no native Mattermost receiver.


deploy/helm/chart/observability/dashboards/generated/status-list-slo.json
```diff
@@ -0,0 +1,328 @@
...
+              }
+            ]
+          }
+        }
+      },
+      "targets": [
+        {
+          "expr": "sli:request_latency:p95:5m",
+          "legendFormat": "p95",
+          "refId": "A"
```
forkimenjeckayang, line 75: This dashboard is being published to a cluster-wide Grafana, but the query has no namespace matcher and the dashboard defines no namespace variable. Because the recording series retain `namespace`, multiple Status List releases will be mixed into panels with identical legends, and the stat panels will return ambiguous values. Please add a Prometheus-backed `namespace` dashboard variable, filter every query with `{namespace=~"$namespace"}` (or an equivalent single-value matcher), and include the namespace in legends. Make the change in `observability/dashboards/src/generate.mjs` and regenerate both artifacts. This is required for the shared-cluster isolation described in our architecture. Grafana documents namespace/job-style template variables here: [Prometheus template variables](https://grafana.com/docs/grafana/latest/datasources/prometheus/template-variables/).

deploy/helm/chart/templates/prometheusrule.yaml
```diff
@@ -77,7 +77,7 @@ spec:
             runbook_url: "https://github.com/adorsys/status-list-server/blob/develop/observability/README.md"
             dashboard_url: {{ .Values.alerting.dashboardUrl | default "http://localhost:3000/d/status-list-slo" | quote }}
         - alert: StatusListMetricsAbsent
-          expr: absent(http_server_duration_seconds_bucket{otel_scope_name="status-list-server"})
+          expr: absent(http_server_duration_seconds_bucket{otel_scope_name="status-list-server",namespace="{{ .Release.Namespace }}"})
           for: 5m
```
forkimenjeckayang, line 91: This can page for a healthy but fresh/idle deployment because it checks whether a request histogram exists, not whether Prometheus can scrape the target. `/metrics` and health routes are excluded from the application's HTTP metrics middleware, so scrapes themselves do not create this histogram. Please alert on the scrape-generated `up` series instead, covering both a missing target and failed scrapes, for example `absent(up{namespace="{{ .Release.Namespace }}",service="<rendered-service-name>"}) or max(up{namespace="{{ .Release.Namespace }}",service="<rendered-service-name>"}) == 0`. Use the actual templated Service name and add a test for both states. Prometheus uses `up == 0` for unreachable-instance alerts in its [official alerting-rule example](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).

deploy/helm/chart/templates/prometheusrule.yaml
```diff
@@ -100,7 +100,7 @@ spec:
             runbook_url: "https://github.com/adorsys/status-list-server/blob/develop/observability/runbooks/request-latency.md"
             dashboard_url: {{ .Values.alerting.dashboardUrl | default "http://localhost:3000/d/status-list-slo" | quote }}
```
forkimenjeckayang, line 115: The query is now namespace-scoped, but the emitted alert still does not satisfy our shared Alertmanager contract: every alert needs `team: statuslist`, and this actionable alert also needs a useful `description` (ideally including the evaluated value and namespace). The same `team` omission exists on every alert, and nine page/warn alerts in this file lack descriptions. Please update the full rule set and the promtool expectations so routing ownership and notification context are enforced in CI. Prometheus explains how alert labels drive identity/routing metadata and annotations carry descriptions/runbook context in [Alerting rules](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).

deploy/helm/chart/values.yaml
```diff
@@ -323,11 +330,47 @@ serviceMonitor:
...
+#    Helm release, otherwise the API server rejects those objects. Use bounded
+#    readiness gates / a staged pipeline; an upgrade on a cluster where the CRDs
+#    already exist does not prove first-install ordering.
+# 2. Retention: the SLO rules here use 30d windows (sli:error_rate:30d,
+#    sli:token_gen_failure_rate:30d, error-budget gates). The shared Prometheus
+#    must retain >30 days (with enough storage that the retention-size limit does
+#    not truncate history sooner) or these 30d windows are not representable.
+# 3. Discovery label: production sets ServiceMonitor/PrometheusRule `labels.release:
+#    kube-prometheus-stack`, which is exactly what the shared Prometheus requires to
+#    select these objects cluster-wide (label opt-in). Keep it.
```
forkimenjeckayang, line 406: Please add the agreed ingestion decision to this new cluster-dependency section: for this EKS integration, the Status List `/metrics` endpoint discovered by `ServiceMonitor` is the authoritative Prometheus path. The application exports the same instruments through both its pull reader and its production OTLP reader, while `values-production.yaml` still suggests sending OTLP metrics to Prometheus Remote Write. Pointing that pipeline at the shared Prometheus would ingest the same signals twice. Remove/narrow that EKS guidance and state that OTLP metrics may go to a different backend but not back into this Prometheus; traces/logs are independent. The OpenTelemetry Metrics SDK explicitly supports multiple readers, meaning both configured outputs collect the instruments: [Metrics SDK](https://opentelemetry.io/docs/specs/otel/metrics/sdk/)

.github/workflows/render-helm-templates/action.yml
```diff
@@ -519,6 +519,56 @@ runs:
...
+          --output-dir /tmp/rendered/scoped-rules
+        RULES=/tmp/rendered/scoped-rules/status-list-server/templates/prometheusrule.yaml
+        if ! grep -q 'otel_scope_name="status-list-server",namespace="statuslist-prod"' "$RULES"; then
+          echo "ERROR: PrometheusRule raw selectors must be namespace-scoped"; exit 1
+        fi
+        if ! grep -q 'absent(http_server_duration_seconds_bucket{otel_scope_name="status-list-server",namespace="statuslist-prod"})' "$RULES"; then
+          echo "ERROR: StatusListMetricsAbsent must be namespace-scoped, not cluster-global"; exit 1
+        fi
+        if ! grep -q 'sli:error_budget:success:30d{namespace="statuslist-prod"}' "$RULES"; then
+          echo "ERROR: recording-series reads in alert rules must be namespace-scoped"; exit 1
```
forkimenjeckayang, line 616: These greps check three representative strings, but they do not prove the PR's cross-namespace isolation requirement. Please add a behavioral promtool case for the rendered `ns1` rule containing only firing/healthy series from `ns2`: `ns1` SLO alerts must not consume them, and loss of the expected `ns1` scrape target must not be masked by `ns2`. That test will validate the outcome across all relevant rules rather than only the presence of three substrings.

helm/chart/templates/dashboard-configmap.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if and .Values.statuslist.enabled .Values.grafanaDashboard.enabled }}
+{{- /*
+  Publishes the generated SLO dashboard as a ConfigMap so the central Grafana
+  sidecar (kube-prometheus-stack configures `sidecar.dashboards.searchNamespace`
+  + `label: grafana_dashboard / labelValue: "1"`) auto-provisions it.
+
+  Source of truth: observability/dashboards/generated/status-list-slo.json,
+  regenerated from src/generate.mjs (see observability/dashboards/README.md).
+  The chart-local copy under helm/chart/observability/ is kept in sync so Helm
+  can embed it via .Files.Get (Helm only reads files inside the chart tree).
```
forkimenjeckayang: The comment says the chart-local dashboard is kept in sync, but the documented generator writes only `observability/dashboards/generated/status-list-slo.json`, and CI only diffs that original file. A future dashboard update can therefore pass CI while this ConfigMap embeds stale JSON. Please make generation copy/write both artifacts and add a byte-for-byte CI comparison (or establish a single generated source that the chart packaging step consumes) so the deployed dashboard cannot drift.

helm/chart/values.yaml
```diff
@@ -323,11 +330,47 @@ serviceMonitor:
...
+
+# Cluster dependency notes for the shared kube-prometheus-stack integration
+# (managed centrally by wallet-eks-env, not by this chart):
+#
+# 1. CRD install order: the ServiceMonitor, PrometheusRule (and any
+#    AlertmanagerConfig / Grafana dashboard) objects below assume the
+#    kube-prometheus-stack CRDs already exist in the cluster. On a fresh
+#    cluster, install the monitoring stack (which installs the CRDs) BEFORE this
+#    Helm release, otherwise the API server rejects those objects. Use bounded
+#    readiness gates / a staged pipeline; an upgrade on a cluster where the CRDs
```
forkimenjeckayang: Please separate CRD ordering from dashboard discovery here. The dashboard is a `core/v1` ConfigMap, so the API server can accept it without kube-prometheus-stack; it only becomes operational when the Grafana sidecar is present. `ServiceMonitor`, `PrometheusRule`, and `AlertmanagerConfig` are the resources that require their CRDs before this Helm release. The overall deployment order remains stack first, then application, but the reason should be technically accurate.

forkimenjeckayang (changes requested):
a few comments please check

.github/workflows/CI.yml
```diff
@@ -919,16 +979,20 @@ jobs:
...
+        # regenerated artifact cannot land silently. The generator writes the
+        # canonical JSON AND the chart-local copy the Helm ConfigMap embeds; a
+        # byte-for-byte comparison of both guards against a deployed dashboard
+        # that drifts from the reviewed source.
         run: |
           cd observability/dashboards/src
           npm install --no-audit --no-fund
           npm run generate-dashboards
           git diff --exit-code ../generated/status-list-slo.json
+          cmp ../generated/status-list-slo.json ../../../helm/chart/observability/dashboards/generated/status-list-slo.json
```
forkimenjeckayang, line 1023: The generator now writes both artifacts, but this cmp runs after regeneration has already overwritten the chart-local copy. Only the canonical file is checked against Git. I reproduced a committed chart-only stale dashboard: these checks both pass while the chart-local Git diff is nonempty.

Please check both committed artifact paths with git diff --exit-code after generation. The cmp can remain, but cannot replace that second Git check. This completes the original drift finding without removing the chart-packaged dashboard.

helm/chart/values.yaml
```diff
@@ -323,11 +330,61 @@ serviceMonitor:
...
+
+# Cluster dependency notes for the shared kube-prometheus-stack integration
+# (managed centrally by wallet-eks-env, not by this chart):
+#
+# 1. CRD install order: the resources below differ in what the API server
+#    requires. ServiceMonitor, PrometheusRule, and AlertmanagerConfig are CRDs
+#    owned by the kube-prometheus-stack operator, so they can only be created
+#    AFTER those CRDs exist. On a fresh cluster, install the monitoring stack
+#    (which installs the CRDs) BEFORE this Helm release, otherwise the API server
+#    rejects those objects. The Grafana dashboard, by contrast, is a core/v1
```
forkimenjeckayang: Installing the stack first is not sufficient for the optional application route: helm/chart/templates/alertmanagerconfig.yaml:22 renders monitoring.coreos.com/v1beta1, but our pinned kube-prometheus-stack 91.4.0 AlertmanagerConfig CRD serves only v1alpha1. Enabling the documented production alerting example therefore fails on a fresh cluster.

Please use the served v1alpha1 API and validate the enabled optional route against the exact 91.4.0 CRD in the render checks, not merely that an AlertmanagerConfig file exists. Keep application routing optional; the default platform fallback does not require enabling it.

Pinned CRD: https://github.com/prometheus-community/helm-charts/blob/kube-prometheus-stack-91.4.0/charts/kube-prometheus-stack/charts/crds/crds/crd-alertmanagerconfigs.yaml#L20-L25

observability/prometheus/rules/alerting.rules.yml
```diff
@@ -20,25 +20,33 @@ groups:
...
+      # never produces the histogram and a histogram-based check pages falsely.
+      # The `up` series is created by the scrape itself, so `max(up) == 0`
+      # catches a target that is present but failing scrapes while `absent(up)`
+      # catches a target that is gone entirely.
       - alert: StatusListMetricsAbsent
-        expr: absent(http_server_duration_seconds_bucket{otel_scope_name="status-list-server"})
+        expr: |
+          absent(up{otel_scope_name="status-list-server"})
+          or
+          max by (otel_scope_name)(up{otel_scope_name="status-list-server"}) == 0
```
forkimenjeckayang: The standalone up guard retains a label from application instruments, not the scrape target. observability/prometheus/prometheus.yml and prometheus.production.yml configure job=status_list_server without an otel_scope_name target label. A healthy up{job="status_list_server",instance="app:8000"} therefore does not match this selector, so absent() pages continuously.

Please select the configured standalone job/target, update the healthy/failing/absent fixtures in observability/prometheus/tests/alerting.test.yml to use realistic target labels, and adjust the Helm fixture adapter in .github/workflows/CI.yml accordingly. Keep the correctly scoped namespace/Service selector in the Helm rule.

Prometheus-generated up labels: https://prometheus.io/docs/concepts/jobs_instances/

forkimenjeckayang (commented):
a few comments please check

forkimenjeckayang:
You can add @Awambeng to check this too

Awambeng (commented):
@Christiantyemele can you review this. I won't be able to review it

PR #521, opened 2026-09-10
feat: update database password handling and support MySQL

### LINKED ISSUE #498: Decouple Helm chart from Postgres assumptions and support multi-database
## Objective

Decouple the Helm chart from hardcoded PostgreSQL assumptions, support other supported database backends (e.g., MySQL) in their latest versions, rename the database password Secret key to a database-agnostic name (`database-password`) with backward compatibility, and generalize `initContainers`.

## Context

Currently, the Helm chart has tight coupling to PostgreSQL:
1. `initContainers` enforces `wait-for-postgres`, hardcoding `{{ .Release.Name }}-postgres...` and nc check on postgres port.
2. The Secret key for database authentication is hardcoded to `postgres-password` in `fallbackSecret`, `secretMounts`, and deployment helpers.
3. Default database environment variables in `deployment.yaml` (`APP_DATABASE__HOST`, `APP_DATABASE__USERNAME`, `APP_DATABASE__NAME`) default to `.Values.postgres` fields even when MySQL is configured.
4. Network policy egress rules only target pods labeled `app.kubernetes.io/name: postgres`.

## Deliverables

- [x] **Database-agnostic Secret key:**
  - Support `database-password` across `statuslist-secret`, `secretMounts`, and documentation.
  - Retain fallback reading of `postgres-password` from existing cluster Secrets and set both `database-password` and `postgres-password` in `$stringData` to guarantee zero breaking changes for existing deployments and bundled subcharts.
- [x] **Generic initContainers:**
  - Replace `wait-for-postgres` with `wait-for-db` in `values.yaml` and `values-local.yaml`.
  - Introduce `status-list-server-chart.dbHost` helper that dynamically resolves to the configured host (or defaults to `<release>-mysql` or `<release>-postgres` based on `APP_DATABASE__BACKEND`).
- [x] **MySQL backend support:**
  - Add native `mysql:` configuration block in `values.yaml` and `values.schema.json` with latest available versions (e.g. MySQL 8.4 LTS).
  - Update `deployment.yaml` to dynamically default `APP_DATABASE__HOST`, `APP_DATABASE__USERNAME`, and `APP_DATABASE__NAME` according to the active backend (`postgres` or `mysql`).
  - Document latest supported versions for both PostgreSQL (17) and MySQL (8.4).
- [x] **Network policy adaptability:**
  - Update `network-policy.yaml` egress podSelector to match the active backend (`postgres` or `mysql`).
- [x] **Validation & tests:**
  - Update and add assertions in `tests/helm_sensitive_env.rs`.
  - Verify `helm lint ./helm/chart` passes cleanly.

## Acceptance Criteria

- [x] `helm lint ./helm/chart` passes without errors.
- [x] `cargo test --test helm_sensitive_env` passes.
- [x] Deployments with `--set statuslist.env.APP_DATABASE__BACKEND=mysql` correctly template MySQL host, credentials, initContainers, and network policy egress.
- [x] Existing deployments using `postgres-password` continue to work without breaking changes.

1 day

Ngha-Boris:
## Summary

Implements database-agnostic Helm chart support for PostgreSQL and MySQL.

- Adds helper-driven DB defaults for backend, host, port, username, database name, and NetworkPolicy selector.
- Replaces `wait-for-postgres` with generic `wait-for-db`.
- Switches the application password mount to `database-password`.
- Preserves backward compatibility by rendering both `database-password` and legacy `postgres-password`.
- Adds native `mysql:` Helm values/schema defaults using MySQL 8.4 LTS.
- Updates release image feature sets so standard variants include both PostgreSQL and MySQL drivers.
- Updates Helm/backend/troubleshooting/supply-chain docs.
- Expands Helm rendering tests for MySQL and legacy password compatibility.

## Validation

- `helm lint ./helm/chart`
- `cargo test --test helm_sensitive_env`
- `cargo fmt --check`
- `cargo check --workspace --features postgres,mysql,aws`
- `cargo check --workspace --features postgres,mysql,gcp`
- `cargo check --workspace --features postgres,mysql,azure`
- `cargo check --workspace --features postgres,mysql,vault`
- `cargo check --workspace --features postgres,mysql`

### DIFF
--- a/.github/workflows/render-helm-templates/action.yml
+++ b/.github/workflows/render-helm-templates/action.yml
@@ -48,8 +48,8 @@ runs:
           --set secretStore.enabled=true \
           --set secretStore.provider=gcp \
           --set secretStore.gcp.projectID=my-project-id \
-          --set-string 'externalSecret.spec.data[0].secretKey=postgres-password' \
-          --set-string 'externalSecret.spec.data[0].remoteRef.key=statuslist-postgres-password' \
+          --set-string 'externalSecret.spec.data[0].secretKey=database-password' \
+          --set-string 'externalSecret.spec.data[0].remoteRef.key=statuslist-database-password' \
           --set-string 'externalSecret.spec.data[0].remoteRef.version=latest' \
           --set-string 'externalSecret.spec.data[0].remoteRef.decodingStrategy=None' \
           --set-string 'externalSecret.spec.data[0].remoteRef.conversionStrategy=Default' \
--- a/docs/database-backends.md
+++ b/docs/database-backends.md
@@ -45,13 +45,13 @@ Best default for production deployments. PostgreSQL is the safest choice when yo
 - strong transactional guarantees
 - straightforward backup and restore workflows
 
-The Helm chart defaults the bundled PostgreSQL image tag to PostgreSQL `18.6`.
-
 ### MySQL
 
 Good fit when your infrastructure already standardizes on MySQL-compatible services or when you want a production database with familiar operational patterns. For MariaDB, use this same backend setting because the driver path is shared.
 
 The Helm chart provides native MySQL defaults under `mysql:` and uses MySQL `8.4.12` LTS for that block.
+
+For Helm-based MySQL deployments, set `postgres.enabled=false`, `mysql.enabled=true`, and `statuslist.env.APP_DATABASE__BACKEND=mysql`. The chart rejects configurations that enable both database backends or that enable one bundled backend while configuring the application to use the other.
 
 ### SQLite
 
@@ -100,10 +100,10 @@ SET GLOBAL binlog_format = 'ROW';  -- then restart the server
 
 ## Compose Profiles
 
-`docker compose up` starts PostgreSQL by default and builds the container with `postgres,mysql,aws` features enabled. To run the MySQL service instead:
+`docker compose up` starts PostgreSQL by default and builds the container with `postgres,aws` features enabled. To run the MySQL service instead:
 
 ```bash
-docker compose --profile mysql up --build
+FEATURES="mysql,aws" docker compose --profile mysql up --build
 ```
 
 There is no separate MariaDB service because MariaDB uses the same MySQL-driver path. Connect to a MariaDB host by setting `APP_DATABASE__BACKEND=mysql`, `APP_DATABASE__HOST`, `APP_DATABASE__PORT`, `APP_DATABASE__USERNAME`, `APP_DATABASE__PASSWORD`, and `APP_DATABASE__NAME`. For local/custom deployments, a `mysql://` `APP_DATABASE__URL` is still supported when no split database fields are set.
--- a/docs/database-backends.md
+++ b/docs/database-backends.md
@@ -26,7 +26,9 @@ APP_DATABASE__QUERY=sslmode=verify-full&sslrootcert=/var/run/postgres/ca.crt
 
 `APP_DATABASE__URL` remains supported for local and custom deployments. Do not combine it with any split database field; startup rejects that ambiguous configuration.
 
-When deploying the Helm chart, `APP_DATABASE__PASSWORD` is not accepted as a plain `statuslist.env` value. The chart exposes `APP_DATABASE__PASSWORD_FILE` through `statuslist.secretMounts` by default, reading the `postgres-password` key from `statuslist-secret`. For external databases, point the split host/port/backend/name/user fields at the external database and create or sync the configured Secret/key with that database password.
+When deploying the Helm chart, `APP_DATABASE__PASSWORD` is not accepted as a plain `statuslist.env` value. The chart exposes `APP_DATABASE__PASSWORD_FILE` through `statuslist.secretMounts` by default, reading the `database-password` key from `statuslist-secret`. For external databases, point the split host/port/backend/name/user fields at the external database and create or sync the configured Secret/key with that database password.
+
+The Helm chart's default database password key is `database-password`. For compatibility with existing PostgreSQL deployments, chart-managed fallback and External Secrets Operator examples also keep `postgres-password` populated with the same value.
 
 The bundled in-cluster PostgreSQL chart is used without chart-managed database TLS material. For managed or external databases, prefer TLS by setting non-secret query parameters such as `APP_DATABASE__QUERY=sslmode=verify-full&sslrootcert=/var/run/postgres/ca.crt`. The application passes these parameters through after validating that query keys are not credential-like; the referenced CA path must already exist in the container.
 
@@ -43,9 +45,13 @@ Best default for production deployments. PostgreSQL is the safest choice when yo
 - strong transactional guarantees
 - straightforward backup and restore workflows
 
+The Helm chart defaults the bundled PostgreSQL image tag to PostgreSQL `18.6`.
+
 ### MySQL
 
 Good fit when your infrastructure already standardizes on MySQL-compatible services or when you want a production database with familiar operational patterns. For MariaDB, use this same backend setting because the driver path is shared.
+
+The Helm chart provides native MySQL defaults under `mysql:` and uses MySQL `8.4.12` LTS for that block.
 
 ### SQLite
 
@@ -94,10 +100,10 @@ SET GLOBAL binlog_format = 'ROW';  -- then restart the server
 
 ## Compose Profiles
 
-`docker compose up` starts PostgreSQL by default and builds the container with `postgres,aws` features enabled. To run the MySQL service instead:
+`docker compose up` starts PostgreSQL by default and builds the container with `postgres,mysql,aws` features enabled. To run the MySQL service instead:
 
 ```bash
-FEATURES="mysql,aws" docker compose --profile mysql up --build
+docker compose --profile mysql up --build
 ```
 
 There is no separate MariaDB service because MariaDB uses the same MySQL-driver path. Connect to a MariaDB host by setting `APP_DATABASE__BACKEND=mysql`, `APP_DATABASE__HOST`, `APP_DATABASE__PORT`, `APP_DATABASE__USERNAME`, `APP_DATABASE__PASSWORD`, and `APP_DATABASE__NAME`. For local/custom deployments, a `mysql://` `APP_DATABASE__URL` is still supported when no split database fields are set.
--- a/helm/README.md
+++ b/helm/README.md
@@ -38,12 +38,14 @@ For production, pin the exact artifact by digest rather than tag. A digest is va
 * [`chart/values-production.yaml`](chart/values-production.yaml): production delta applied after `values-aws.yaml` by release deployments.
 * `global.domain`: chart-wide public DNS suffix. When set, Ingress defaults derive `statuslist.<global.domain>` and `*.<global.domain>` from this single value. Rendered hostnames are normalized to lowercase.
 * `postgres.persistence.storageClass`: leave as `""` to use the cluster default StorageClass; set explicitly in environment overlays when needed.
+* `postgres.image.tag`: defaults to PostgreSQL `18.6`, the current supported release line used by this chart.
+* `mysql.image.tag`: defaults to MySQL `8.4.12` LTS for MySQL-compatible deployments.
 * `statuslist.image.variant`: selected image variant when no explicit `tag` or `digest` is set (`fscert`, `aws`, `gcp`, `azure`, or `vault`).
 * `statuslist.image.digest`: takes precedence over `statuslist.image.tag` and renders `repository@digest`.
 
 ## Configure Your Secrets
 
-The application Secret is always named `statuslist-secret` and holds the database password under `postgres-password`. The application Deployment and bundled PostgreSQL both consume that same Secret name.
+The application Secret is always named `statuslist-secret` and holds the database password under `database-password`. For backward compatibility with existing clusters and the bundled PostgreSQL chart, the chart-managed fallback Secret and default ExternalSecret also publish the same value under `postgres-password`.
 
 There are two secret delivery modes, and the chart rejects enabling both at once.
 
@@ -227,7 +229,7 @@ statuslist:
         APP_DATABASE__PASSWORD_FILE: password
 ```
 
-`fileEnv` values are relative to `mountPath`, and they work with or without `items`. By default, the chart mounts the application Secret's `postgres-password` key at `/var/run/status-list-server/database/password` and exposes that path through `APP_DATABASE__PASSWORD_FILE`. You can override `statuslist.secretMounts` to point at another Secret or mount path.
+`fileEnv` values are relative to `mountPath`, and they work with or without `items`. By default, the chart mounts the application Secret's `database-password` key at `/var/run/status-list-server/database/password` and exposes that path through `APP_DATABASE__PASSWORD_FILE`. You can override `statuslist.secretMounts` to point at another Secret or mount path. Existing fallback Secrets that only contain `postgres-password` are read during Helm upgrade and rendered back with both keys.
 
 This chart support is preparatory for application images that implement the file-watcher and reload behavior from issue #456. Current images that only read `APP_DATABASE__PASSWORD` at startup still need a rollout after secret changes. The `checksum/secret` annotation only reacts to Helm-rendered ExternalSecret template or value changes; it does not change when External Secrets Operator later syncs new data from Vault, AWS, GCP, or Azure into a Kubernetes Secret.
 
@@ -242,12 +244,12 @@ statuslist:
   fallbackSecret:
     enabled: true
     stringData:
-      postgres-password: ""
+      database-password: ""
 ```
 
-The default chart uses this mode, so a plain `helm install` creates `statuslist-secret`. Leave `postgres-password` empty to have Helm generate a random password; on upgrades, Helm reuses the existing cluster Secret when it can read it. Set a concrete value only for local or disposable environments.
+The default chart uses this mode, so a plain `helm install` creates `statuslist-secret`. Leave `database-password` empty to have Helm generate a random password; on upgrades, Helm reuses an existing `database-password` value, or falls back to an existing legacy `postgres-password` value when it can read the cluster Secret. The rendered Secret always contains both keys with the same value. Set a concrete value only for local or disposable environments.
 
-GitOps caveat: tools such as Argo CD and Flux render charts with `helm template`, where Helm's live `lookup` function cannot read the existing Secret. If `postgres-password` is left empty, each render generates a new password while PostgreSQL may keep the old password in its PVC. GitOps deployments should set an explicit fallback password from their secret-management flow or use ESO mode instead. The fallback Secret is annotated with `helm.sh/resource-policy: keep` so Helm does not delete it on uninstall.
+GitOps caveat: tools such as Argo CD and Flux render charts with `helm template`, where Helm's live `lookup` function cannot read the existing Secret. If `database-password` is left empty, each render generates a new password while PostgreSQL may keep the old password in its PVC. GitOps deployments should set an explicit fallback password from their secret-management flow or use ESO mode instead. The fallback Secret is annotated with `helm.sh/resource-policy: keep` so Helm does not delete it on uninstall.
 
 **External Secrets Operator.** ESO syncs `statuslist-secret` from a configured `SecretStore` or pre-existing `ClusterSecretStore`. Enable ESO mode explicitly:
 
@@ -264,7 +266,7 @@ statuslist:
 
 Provider selection is fail-closed through `values.schema.json` and render-time checks. Unsupported providers, empty `raw: {}`, ESO without a SecretStore, and custom `externalSecret.spec.target.name` values fail before Kubernetes receives manifests.
 
-Common non-secret values under `statuslist.env` are the split database fields (`APP_DATABASE__HOST`, `APP_DATABASE__PORT`, `APP_DATABASE__USERNAME`, `APP_DATABASE__NAME`) and server values (`APP_SERVER__HOST`, `APP_SERVER__PORT`, `APP_SERVER__DOMAIN`). Do not set `APP_DATABASE__PASSWORD` in Helm values; the chart wires the password from the Secret as `APP_DATABASE__PASSWORD_FILE`.
+Common non-secret values under `statuslist.env` are the split database fields (`APP_DATABASE__BACKEND`, `APP_DATABASE__HOST`, `APP_DATABASE__PORT`, `APP_DATABASE__USERNAME`, `APP_DATABASE__NAME`) and server values (`APP_SERVER__HOST`, `APP_SERVER__PORT`, `APP_SERVER__DOMAIN`). Do not set `APP_DATABASE__PASSWORD` in Helm values; the chart wires the password from the Secret as `APP_DATABASE__PASSWORD_FILE`.
 
 ## Use Workload Identity Instead of Mounted Credentials
 
@@ -352,7 +354,7 @@ helm upgrade --install statuslist helm/chart \
   --wait --timeout 10m
 ```
 
-The chart bundles PostgreSQL and an OpenTelemetry collector. To point at an external database, disable the bundled PostgreSQL subchart and set the split `APP_DATABASE__*` fields under `statuslist.env`.
+The chart bundles PostgreSQL and an OpenTelemetry collector. To point at an external database, disable the bundled PostgreSQL subchart and set the split `APP_DATABASE__*` fields under `statuslist.env`. For MySQL, set `statuslist.env.APP_DATABASE__BACKEND=mysql`; if host, port, username, or database name are omitted, the chart defaults them from the `mysql:` values block (`<release>-mysql.<namespace>.svc.cluster.local`, port `3306`, `mysql.auth.username`, and `mysql.auth.database`). This chart does not currently vendor a MySQL subchart, so provide that MySQL Service through your platform, operator, or an overlay.
 
 ## Verify the Deployment
 
--- a/helm/README.md
+++ b/helm/README.md
@@ -38,7 +38,6 @@ For production, pin the exact artifact by digest rather than tag. A digest is va
 * [`chart/values-production.yaml`](chart/values-production.yaml): production delta applied after `values-aws.yaml` by release deployments.
 * `global.domain`: chart-wide public DNS suffix. When set, Ingress defaults derive `statuslist.<global.domain>` and `*.<global.domain>` from this single value. Rendered hostnames are normalized to lowercase.
 * `postgres.persistence.storageClass`: leave as `""` to use the cluster default StorageClass; set explicitly in environment overlays when needed.
-* `postgres.image.tag`: defaults to PostgreSQL `18.6`, the current supported release line used by this chart.
 * `mysql.image.tag`: defaults to MySQL `8.4.12` LTS for MySQL-compatible deployments.
 * `statuslist.image.variant`: selected image variant when no explicit `tag` or `digest` is set (`fscert`, `aws`, `gcp`, `azure`, or `vault`).
 * `statuslist.image.digest`: takes precedence over `statuslist.image.tag` and renders `repository@digest`.
@@ -354,7 +353,7 @@ helm upgrade --install statuslist helm/chart \
   --wait --timeout 10m
 ```
 
-The chart bundles PostgreSQL and an OpenTelemetry collector. To point at an external database, disable the bundled PostgreSQL subchart and set the split `APP_DATABASE__*` fields under `statuslist.env`. For MySQL, set `statuslist.env.APP_DATABASE__BACKEND=mysql`; if host, port, username, or database name are omitted, the chart defaults them from the `mysql:` values block (`<release>-mysql.<namespace>.svc.cluster.local`, port `3306`, `mysql.auth.username`, and `mysql.auth.database`). This chart does not currently vendor a MySQL subchart, so provide that MySQL Service through your platform, operator, or an overlay.
+The chart bundles PostgreSQL and an OpenTelemetry collector. To point at an external database, disable the bundled PostgreSQL subchart and set the split `APP_DATABASE__*` fields under `statuslist.env`. For MySQL, set `postgres.enabled=false`, `mysql.enabled=true`, and `statuslist.env.APP_DATABASE__BACKEND=mysql`; if host, port, username, or database name are omitted, the chart defaults them from the `mysql:` values block (`<release>-mysql.<namespace>.svc.cluster.local`, port `3306`, `mysql.auth.username`, and `mysql.auth.database`). The database backend flags are mutually exclusive. This chart does not currently vendor a MySQL subchart, so provide that MySQL Service through your platform, operator, or an overlay.
 
 ## Verify the Deployment
 
--- a/helm/chart/templates/_helpers.tpl
+++ b/helm/chart/templates/_helpers.tpl
@@ -63,7 +63,7 @@ Create the name of the service account to use
 {{- end }}
 
 {{/*
-Effective name of the Kubernetes Secret the application reads (postgres password).
+Effective name of the Kubernetes Secret the application reads (database password).
 Single supported name: "statuslist-secret" in both ESO mode (ExternalSecret target) and
 fallback mode. The Deployment, PostgreSQL (postgres.auth.existingSecret), and the fallback
 Secret all reference this same name, so it is not independently configurable. ESO mode
@@ -88,11 +88,85 @@ provider). Preference: explicit statuslist.aws.region, then the legacy secretSto
 {{- end }}
 
 {{/*
-Database port helper: returns the configured database port from env.
-Note: APP_DATABASE__PORT is required by deployment.yaml; this helper
-assumes the value exists and does not provide a default.
+Effective database backend for chart-managed defaults.
+*/}}
+{{- define "status-list-server-chart.dbBackend" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- get $env "APP_DATABASE__BACKEND" | default "postgres" | lower }}
+{{- end }}
+
+{{/*
+Database host helper: returns the configured host, or the default in-cluster
+service name for the active backend.
+*/}}
+{{- define "status-list-server-chart.dbHost" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- $backend := include "status-list-server-chart.dbBackend" . }}
+{{- if get $env "APP_DATABASE__HOST" }}
+{{- get $env "APP_DATABASE__HOST" }}
+{{- else if eq $backend "mysql" }}
+{{- printf "%s-mysql.%s.svc.cluster.local" .Release.Name .Release.Namespace }}
+{{- else if eq $backend "postgres" }}
+{{- printf "%s-postgres.%s.svc.cluster.local" .Release.Name .Release.Namespace }}
+{{- else }}
+{{- fail (printf "statuslist.env.APP_DATABASE__BACKEND must be either postgres or mysql for this chart, got %q" $backend) }}
+{{- end }}
+{{- end }}
+
+{{/*
+Database port helper: returns the configured port, or the active backend default.
 */}}
 {{- define "status-list-server-chart.dbPort" -}}
 {{- $env := .Values.statuslist.env | default dict }}
+{{- $backend := include "status-list-server-chart.dbBackend" . }}
+{{- if get $env "APP_DATABASE__PORT" }}
 {{- get $env "APP_DATABASE__PORT" }}
+{{- else if eq $backend "mysql" }}
+{{- .Values.mysql.service.port | toString }}
+{{- else if eq $backend "postgres" }}
+{{- .Values.postgres.service.port | toString }}
+{{- else }}
+{{- fail (printf "statuslist.env.APP_DATABASE__BACKEND must be either postgres or mysql for this chart, got %q" $backend) }}
+{{- end }}
+{{- end }}
+
+{{/*
+Database username helper: returns the configured username, or the active backend default.
+*/}}
+{{- define "status-list-server-chart.dbUsername" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- $backend := include "status-list-server-chart.dbBackend" . }}
+{{- if get $env "APP_DATABASE__USERNAME" }}
+{{- get $env "APP_DATABASE__USERNAME" }}
+{{- else if eq $backend "mysql" }}
+{{- .Values.mysql.auth.username }}
+{{- else if eq $backend "postgres" }}
+{{- .Values.postgres.auth.username }}
+{{- else }}
+{{- fail (printf "statuslist.env.APP_DATABASE__BACKEND must be either postgres or mysql for this chart, got %q" $backend) }}
+{{- end }}
+{{- end }}
+
+{{/*
+Database name helper: returns the configured database name, or the active backend default.
+*/}}
+{{- define "status-list-server-chart.dbName" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- $backend := include "status-list-server-chart.dbBackend" . }}
+{{- if get $env "APP_DATABASE__NAME" }}
+{{- get $env "APP_DATABASE__NAME" }}
+{{- else if eq $backend "mysql" }}
+{{- .Values.mysql.auth.database }}
+{{- else if eq $backend "postgres" }}
+{{- .Values.postgres.auth.database }}
+{{- else }}
+{{- fail (printf "statuslist.env.APP_DATABASE__BACKEND must be either postgres or mysql for this chart, got %q" $backend) }}
+{{- end }}
+{{- end }}
+
+{{/*
+Database pod label value for the chart-managed NetworkPolicy egress selector.
+*/}}
+{{- define "status-list-server-chart.dbPodSelectorName" -}}
+{{- include "status-list-server-chart.dbBackend" . }}
 {{- end }}
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -2,6 +2,16 @@
 {{- if and .Values.externalSecret.enabled .Values.statuslist.fallbackSecret.enabled }}
 {{- fail "externalSecret.enabled and statuslist.fallbackSecret.enabled cannot both be true: choose either External Secrets Operator or the fallback Secret for secret delivery, not both" }}
 {{- end }}
+{{- $databaseBackend := include "status-list-server-chart.dbBackend" . }}
+{{- if and .Values.postgres.enabled .Values.mysql.enabled }}
+{{- fail "postgres.enabled and mysql.enabled cannot both be true: choose a single database backend" }}
+{{- end }}
+{{- if and .Values.postgres.enabled (ne $databaseBackend "postgres") }}
+{{- fail (printf "postgres.enabled=true requires statuslist.env.APP_DATABASE__BACKEND=postgres, got %q" $databaseBackend) }}
+{{- end }}
+{{- if and .Values.mysql.enabled (ne $databaseBackend "mysql") }}
+{{- fail (printf "mysql.enabled=true requires statuslist.env.APP_DATABASE__BACKEND=mysql, got %q" $databaseBackend) }}
+{{- end }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/helm/chart/templates/deployment.yaml
+++ b/helm/chart/templates/deployment.yaml
@@ -159,9 +159,6 @@ spec:
             {{- if hasKey $env "APP_DATABASE__PASSWORD" }}
               {{- fail "statuslist.env.APP_DATABASE__PASSWORD must not be set as a plain env value; use the chart-managed APP_DATABASE__PASSWORD_FILE mounted Secret path" }}
             {{- end }}
-            {{- if or (not (hasKey $env "APP_DATABASE__PORT")) (eq (trim (toString (get $env "APP_DATABASE__PORT"))) "") }}
-              {{- fail "statuslist.env.APP_DATABASE__PORT is required; set it to the database service port" }}
-            {{- end }}
             {{- if not (hasKey $env "APP_SERVER__DOMAIN") }}
               {{- $globalDomain := trimSuffix "." (dig "domain" "" (.Values.global | default dict) | default "" | lower) }}
               {{- if or .Values.statuslist.ingress.externalDnsHostname $globalDomain }}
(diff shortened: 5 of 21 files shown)


Hermann-Core:
@Ngha-Boris, the database backends are supposed to be mutually exclusive. Allowing two database backends to be enabled at the same time is a significant logical issue, as the application should have a single active database backend.

Also, this change appears to expand the scope of the ticket beyond its original purpose, which was to update the deployment charts. Could you please revert the unrelated changes and keep this PR focused on the intended scope?

Ngha-Boris:

tests/helm_sensitive_env.rs
```diff
@@ -147,22 +148,115 @@ fn rendered_chart_uses_split_database_credentials() {
...
-fn rendered_chart_respects_database_backend_override() {
+fn rendered_chart_templates_mysql_backend_defaults() {
     let Some(rendered) = render_helm(&[
         "--set",
         "statuslist.env.APP_DATABASE__BACKEND=mysql",
         "--set",
-        "statuslist.env.APP_DATABASE__PORT=3306",
+        "mysql.enabled=true",
+        "--set",
+        "statuslist.networkPolicy.enabled=true",
```
forkimenjeckayang, line 189: This test proves that MySQL values render, but the same manifest still selects `ghcr.io/adorsys/status-list-server:1.0.1-fscert`. Every published image variant in `deploy.yml` is compiled with `postgres` and without the `mysql` Cargo feature, which is what enables the SQLx MySQL driver. Changing `APP_DATABASE__BACKEND` at runtime cannot add that driver, so the documented MySQL deployment will render successfully but cannot connect. Please publish and select a MySQL-capable image while keeping the backend builds mutually exclusive (for example, separate database-specific variants), and extend the test/release checks to verify that the image selected for a MySQL render was built with `mysql`. Please also update the PR description to match the final image strategy. Reference: [Cargo features are compile-time configuration](https://doc.rust-lang.org/cargo/reference/features.html).

helm/chart/values.yaml
```diff
@@ -207,7 +209,7 @@ statuslist:
       secretName: statuslist-secret
       mountPath: /var/run/status-list-server/database
       items:
-        - key: postgres-password
+        - key: database-password
           path: password
       fileEnv:
         APP_DATABASE__PASSWORD_FILE: password
```
forkimenjeckayang, line 225: This mount breaks an existing installation whose separately managed `statuslist-secret` still contains only `postgres-password`. The fallback template can copy the legacy value and the new default ExternalSecret emits both keys, but neither mechanism runs when the Secret is customer-managed or when an existing custom ESO mapping still emits only the legacy key. Kubernetes requires every key listed under a Secret volume's `items` to exist, so the upgraded Pod will fail to mount this volume. Please provide an upgrade-safe/configurable key migration for every supported secret-delivery mode and add a render/upgrade test for a legacy externally managed Secret. Reference: [Kubernetes Secret volume behavior — all listed keys must exist](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#project-secret-keys-to-specific-file-paths).
WDYT?

helm/chart/values.yaml
```diff
@@ -480,6 +482,27 @@ postgres:
...
+    existingSecret: statuslist-secret
+    secretKeys:
+      passwordKey: database-password
+  service:
+    type: ClusterIP
+    port: 3306
+  image:
+    registry: docker.io
+    repository: mysql
+    tag: "8.4.12"
```
forkimenjeckayang: The configured registry/tag pair is not currently available: `docker manifest inspect mysql:8.4.12` returns `no such manifest: docker.io/library/mysql:8.4.12`, while `mysql:8.4.11` resolves for amd64 and arm64. Please pin a tag that exists in the configured Docker Hub repository, or use the correct registry/repository if 8.4.12 is intentionally sourced elsewhere, and verify it in CI before documenting it. References: [official Docker Hub MySQL tags](https://hub.docker.com/_/mysql/tags) and [MySQL 8.4.12 release notes](https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-12.html).

deploy/helm/chart/templates/_helpers.tpl
```diff
@@ -88,11 +88,85 @@ provider). Preference: explicit statuslist.aws.region, then the legacy secretSto
...
 {{/*
-Database port helper: returns the configured database port from env.
-Note: APP_DATABASE__PORT is required by deployment.yaml; this helper
-assumes the value exists and does not provide a default.
+Effective database backend for chart-managed defaults.
+*/}}
+{{- define "status-list-server-chart.dbBackend" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- get $env "APP_DATABASE__BACKEND" | default "postgres" | lower }}
+{{- end }}
```
forkimenjeckayang, line 96: This lowercases the backend for Helm's comparisons, but the Deployment later renders the original `statuslist.env` value. For example, `APP_DATABASE__BACKEND=MYSQL` passes the chart checks and receives MySQL host/port defaults while the Pod still gets `APP_DATABASE__BACKEND="MYSQL"`; the application enum is `#[serde(rename_all = "lowercase")]` and rejects it. Please either validate the raw value as exactly `postgres`/`mysql` or write the normalized value back into the rendered environment, and cover that behavior with a render test.

forkimenjeckayang (changes requested):
a few comments @Ngha-Boris 
Please check

Ngha-Boris:
Hello @forkimenjeckayang please recheck

helm/chart/templates/deployment.yaml
```diff
@@ -2,6 +2,16 @@
 {{- if and .Values.externalSecret.enabled .Values.statuslist.fallbackSecret.enabled }}
 {{- fail "externalSecret.enabled and statuslist.fallbackSecret.enabled cannot both be true: choose either External Secrets Operator or the fallback Secret for secret delivery, not both" }}
 {{- end }}
+{{- $databaseBackend := include "status-list-server-chart.dbBackend" . }}
+{{- if and .Values.postgres.enabled .Values.mysql.enabled }}
+{{- fail "postgres.enabled and mysql.enabled cannot both be true: choose a single database backend" }}
+{{- end }}
+{{- if and .Values.postgres.enabled (ne $databaseBackend "postgres") }}
+{{- fail (printf "postgres.enabled=true requires statuslist.env.APP_DATABASE__BACKEND=postgres, got %q" $databaseBackend) }}
```
martcpp: The acceptance criterion "`--set statuslist.env.APP_DATABASE__BACKEND=mysql` correctly templates MySQL" fails. `postgres.enabled` defaults to `true` (`values.yaml:469`), so this guard fires:

```
$ helm template sls helm/chart --set statuslist.env.APP_DATABASE__BACKEND=mysql
Error: execution error at (status-list-server/templates/deployment.yaml:10:4): postgres.enabled=true requires statuslist.env.APP_DATABASE__BACKEND=postgres, got "mysql"
```

Adding `mysql.enabled=true` fails too, with "cannot both be true". The tests missed this because `helm_template()` always passes `postgres.enabled=false` (`tests/helm_sensitive_env.rs:76-77`), so the real defaults never render.

The guard is correct, since a template can't override a subchart `condition`. Fix the rest:
- Change the error to name the fix: `set postgres.enabled=false`.
- Update the acceptance criterion and docs to show both flags.
- Add a test that renders `APP_DATABASE__BACKEND=mysql` without the harness's `postgres.enabled=false` and asserts the new message.

deploy/helm/chart/templates/secret.yaml
```diff
@@ -3,21 +3,38 @@
...
 {{- else }}
-{{- /* GitOps renderers such as Argo CD and Flux cannot use lookup here; set an explicit postgres-password or disable fallbackSecret in GitOps deployments. */}}
+{{- if empty $db }}
+{{- /* GitOps renderers such as Argo CD and Flux cannot use lookup here; set an explicit database-password or disable fallbackSecret in GitOps deployments. */}}
 {{- $pg = randAlphaNum 32 }}
 {{- end }}
-{{- $_ := set $stringData "postgres-password" $pg }}
 {{- end }}
+{{- end }}
+{{- if $db }}
```
martcpp, line 34: When the cluster Secret already has both keys, `database-password` always wins and gets copied over `postgres-password`. That undoes a password rotation.

1. After this upgrade, `statuslist-secret` holds `database-password=A` and `postgres-password=A`.
2. An operator rotates `postgres-password` to `B`. That's the key the pod mounts (`values.yaml:224`), and `docs/troubleshooting.md` tells them to rotate it.
3. On the next `helm upgrade`, lookup returns `A` and `B`, line 31 sets `$pg = $db`, and the Secret is written back as `postgres-password=A`.
4. The watcher reloads `A` and the pool fails auth. If `A` was rotated because it leaked, the chart has republished the leaked password.

For explicit `stringData`, lines 8-10 already `fail` on a mismatch; the lookup path doesn't. Apply the same rule to existing keys: `fail` when `$existingDb` and `$existingPg` differ, or take the value from the key named in `statuslist.secretMounts[0].items[0].key`. Don't pick one silently.

.github/workflows/render-helm-templates/action.yml
```diff
@@ -48,8 +57,8 @@ runs:
           --set secretStore.enabled=true \
           --set secretStore.provider=gcp \
           --set secretStore.gcp.projectID=my-project-id \
-          --set-string 'externalSecret.spec.data[0].secretKey=postgres-password' \
-          --set-string 'externalSecret.spec.data[0].remoteRef.key=statuslist-postgres-password' \
+          --set-string 'externalSecret.spec.data[0].secretKey=database-password' \
```
martcpp: This render is a broken deployment, and CI passes it. Helm replaces lists on `--set` instead of merging, so `externalSecret.spec.data` becomes one entry that only emits `database-password`. The Deployment still mounts `postgres-password`, and so does the Postgres subchart (`postgres.auth.secretKeys.passwordKey`):

```
- secretKey: database-password     # ExternalSecret output
- key: postgres-password           # Deployment volume item
```

On a cluster, the pod sits in `ContainerCreating` because the Secret key is missing, and Postgres fails to start. The chart never checks this.

Fix: when `externalSecret.enabled`, gather the keys from `spec.data[].secretKey` and `spec.target.template.data`. `fail` if the mounted key, or `postgres.auth.secretKeys.passwordKey` when Postgres is enabled, isn't among them. Then fix this example to emit both keys.

helm/chart/values.yaml
```diff
@@ -480,6 +494,27 @@ postgres:
   image:
     tag: "18.0"
 
+mysql:
+  # Native MySQL defaults used by deployment.yaml, initContainers, and NetworkPolicy
+  # when APP_DATABASE__BACKEND=mysql. This chart does not vendor a MySQL subchart;
+  # provide the service through your platform/operator or a values overlay.
+  enabled: false
```
martcpp: `mysql.enabled` deploys nothing. `Chart.yaml` has no MySQL dependency, and no template reads `mysql.image.*`, `mysql.auth.existingSecret`, `mysql.auth.secretKeys` or `mysql.service.type`. The only reader of `mysql.enabled` is the validation in `deployment.yaml:6-14`.

Users will set `mysql.enabled=true` expecting a database, like `postgres.enabled`, and get an init container waiting on `<release>-mysql` forever. `render-helm-templates/action.yml:15-22,27` also runs `docker manifest inspect` against Docker Hub for this image, which is never deployed.

Fix, pick one:
- Vendor a MySQL subchart behind `condition: mysql.enabled`.
- Or remove `enabled`, `image`, `existingSecret`, `secretKeys` and `service.type`, and keep only the connection defaults (`auth.username`, `auth.database`, `service.port`), documented as describing an external database. Drop the manifest-inspect step either way.

tests/helm_sensitive_env.rs
```diff
@@ -147,22 +148,184 @@ fn rendered_chart_uses_split_database_credentials() {
...
+    for expected in [
+        "name: wait-for-db",
+        "until nc -z status-list-server-mysql.statuslist.svc.cluster.local 3306; do",
+        "name: APP_DATABASE__BACKEND\n              value: \"mysql\"",
+        "name: APP_DATABASE__HOST\n              value: \"status-list-server-mysql.statuslist.svc.cluster.local\"",
+        "name: APP_DATABASE__PORT\n              value: \"3306\"",
+        "name: APP_DATABASE__USERNAME\n              value: \"mysql\"",
+        "name: APP_DATABASE__NAME\n              value: \"status-list\"",
+        "image: \"ghcr.io/adorsys/status-list-server:1.0.0-mysql-fscert\"",
+        "key: postgres-password",
```
martcpp, line 209: This asserts that a MySQL render mounts `postgres-password`, which is the Postgres coupling the issue set out to remove. The deliverable says `database-password` across `secretMounts`, but the default mount is still `postgres-password` for every backend (`values.yaml:224`). No MySQL install on this chart exists, so there's nothing to stay compatible with.

Fix:
- Default the mount key from the backend: `database-password` for `mysql`, and `postgres-password` for `postgres`.
- Change this assertion to `key: database-password`.
- Add a line to the docs with the chart version where the Postgres default switches to `database-password`.

martcpp (commented):
Six blocking issues: five inline, and this one at PR level.

**Release pipeline change not covered by the issue.**
- `deploy.yml` goes from 5 to 10 images in `build-and-push`, `scan-image` and `promote-tags`.
- `scheduled-image-scan.yml` goes from 5 to 10 nightly variants (20 scan legs across both architectures).
- Build time, GHCR storage and Trivy runs all double.

It's required for MySQL to work, but it needs explicit sign-off or its own PR.

**No image exists yet for the derived MySQL tag.** `deployment.yaml:96` derives `<appVersion>-mysql-<variant>`. Chart `appVersion` `1.0.1` shipped before these variants, so `1.0.1-mysql-fscert` doesn't exist and `backend=mysql` gets `ImagePullBackOff` until the next release. Document that.

**Branch history.** `2f01ac3` and `ee1933a` are identical copies of develop's `299dab7` and `fe43572` (`git range-diff`), followed by a merge. Rebase onto `develop` before merging.

Ngha-Boris:
Hello @Hermann-Core, @martcpp and @forkimenjeckayang before checking again, please pull the recent changes first.

martcpp:
@Ngha-Boris revert the dependency fix you did thank there his s deliciated pr for this that you can just rebase to

Ngha-Boris:
> @Ngha-Boris revert the dependency fix you did thank there his s deliciated pr for this that you can just rebase to

Okay thank you

helm/chart/templates/deployment.yaml
```diff
@@ -289,7 +305,10 @@ spec:
             secretName: {{ .secretName }}
             {{- with .items }}
             items:
-              {{- toYaml . | nindent 14 }}
+              {{- range . }}
+              - key: {{ include "status-list-server-chart.databasePasswordSecretKey" (dict "root" $ "key" .key) }}
+                path: {{ .path }}
+              {{- end }}
```
forkimenjeckayang: This rewrite drops each Secret item's configured `mode`: rendering an item with `mode: 256` (`0400`) now emits only `key` and `path`. Previously `toYaml` preserved it; without a mode Kubernetes defaults to `0644`. Please preserve the item's other fields while changing its key, so an upgrade does not silently relax configured Secret-file permissions. Reference: [Kubernetes Secret-file permissions](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#set-posix-permissions-for-secret-keys).

helm/chart/templates/_helpers.tpl
```diff
@@ -88,11 +88,99 @@ provider). Preference: explicit statuslist.aws.region, then the legacy secretSto
...
+default stays postgres-password for PostgreSQL upgrades, while MySQL defaults to
+the backend-neutral database-password key.
+*/}}
+{{- define "status-list-server-chart.databasePasswordSecretKey" -}}
+{{- $key := .key | default "postgres-password" }}
+{{- if and (eq $key "postgres-password") (eq (include "status-list-server-chart.dbBackend" .root) "mysql") -}}
+database-password
+{{- else -}}
+{{- $key -}}
+{{- end }}
```
forkimenjeckayang: This helper also rewrites explicitly configured keys in operator-managed Secrets. With backend `mysql`, a custom mount selecting `customer-db-secret` / `postgres-password` renders `database-password` instead. That name can legitimately hold a MySQL password; if the selected Secret only has the configured key, the Pod cannot mount it. Please limit backend-dependent key selection to the chart's default database mount and preserve explicit custom mappings, using the same resolved key in ESO validation. Reference: [Kubernetes requires all listed Secret keys to exist](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#project-secret-keys-to-specific-file-paths).

deploy/helm/chart/templates/external-secrets.yaml
```diff
@@ -5,7 +5,34 @@
...
+{{- $appSecretName := include "status-list-server-chart.appSecretName" . }}
+{{- $esoKeys := dict }}
+{{- range .Values.externalSecret.spec.data }}
+{{- $_ := set $esoKeys .secretKey true }}
+{{- end }}
+{{- $targetTemplate := .Values.externalSecret.spec.target.template | default dict }}
+{{- with $targetTemplate.data }}
+{{- range $key, $_ := . }}
+{{- $_ := set $esoKeys $key true }}
+{{- end }}
```
forkimenjeckayang, line 24: These input keys are not necessarily present in the final Secret. With a target template, ESO defaults to `mergePolicy: Replace`, so fetched `spec.data` keys are omitted unless the template emits them. A template containing only `unrelated` currently passes this check although both password keys disappear. Please validate the effective output keys: include fetched keys when no template is used or its policy is `Merge`, and account for the template's emitted keys in `Replace` mode. Reference: [ESO template merge policies](https://external-secrets.io/main/guides/templating/#mergepolicy).

forkimenjeckayang (changes requested):
a few comments please check

Ngha-Boris:
@forkimenjeckayang please recheck

helm/chart/templates/deployment.yaml
```diff
@@ -289,7 +305,10 @@ spec:
             secretName: {{ .secretName }}
             {{- with .items }}
             items:
-              {{- toYaml . | nindent 14 }}
+              {{- range . }}
+              - key: {{ include "status-list-server-chart.databasePasswordSecretKey" (dict "root" $ "key" .key) }}
```
Hermann-Core: Iterating `.items` and reconstructing each entry with only `key` and `path` drops any configured `mode` (e.g. `mode: 256` / `0400`). Kubernetes then falls back to `defaultMode` (`0644`), which silently relaxes file permissions on mounted secrets upon upgrade. Please preserve existing item properties (e.g., via `toYaml` or dictionary merge) when resolving keys.

helm/chart/templates/_helpers.tpl
```diff
@@ -88,11 +88,99 @@ provider). Preference: explicit statuslist.aws.region, then the legacy secretSto
...
+
+{{/*
+Effective Secret item key for database password mounts. The historical chart
+default stays postgres-password for PostgreSQL upgrades, while MySQL defaults to
+the backend-neutral database-password key.
+*/}}
+{{- define "status-list-server-chart.databasePasswordSecretKey" -}}
+{{- $key := .key | default "postgres-password" }}
+{{- if and (eq $key "postgres-password") (eq (include "status-list-server-chart.dbBackend" .root) "mysql") -}}
```
Hermann-Core: This helper unconditionally converts `postgres-password` to `database-password` whenever `APP_DATABASE__BACKEND` is `mysql`. Because `deployment.yaml` applies it across all `statuslist.secretMounts`, any custom or operator-managed Secret that intentionally uses `postgres-password` has its key rewritten, causing volume mount failures in Kubernetes. Please restrict backend-dependent key translation to the chart's default database mount (`statuslist-secret`) and preserve explicit keys configured on custom secret mounts.

deploy/helm/chart/templates/external-secrets.yaml
```diff
@@ -5,7 +5,34 @@
 {{- fail "externalSecret.enabled requires secretStore.enabled: no SecretStore would be rendered, leaving the ExternalSecret with a dangling secretStoreRef" }}
 {{- end }}
 {{- if ne .Values.externalSecret.spec.target.name (include "status-list-server-chart.appSecretName" .) }}
-{{- fail (printf "externalSecret.spec.target.name must be '%s' to match the single application-secret name used by the Deployment and PostgreSQL (postgres.auth.existingSecret)" (include "status-list-server-chart.appSecretName" .)) }}
+{{- fail (printf "externalSecret.spec.target.name must be '%s' to match the single application-secret name used by the Deployment and database subcharts" (include "status-list-server-chart.appSecretName" .)) }}
+{{- end }}
+{{- $appSecretName := include "status-list-server-chart.appSecretName" . }}
+{{- $esoKeys := dict }}
+{{- range .Values.externalSecret.spec.data }}
+{{- $_ := set $esoKeys .secretKey true }}
```
Hermann-Core, line 18: When an ExternalSecret defines `spec.target.template` without an explicit `mergePolicy`, ESO defaults to `Replace`. In `Replace` mode, keys defined only in `spec.data` are not emitted to the final Secret unless referenced in the template. Unioning `spec.data` and `template.data` allows configurations that drop the required database password keys to pass validation. Please check effective keys according to `mergePolicy`: only include `spec.data` keys if `target.template` is absent or specifies `mergePolicy: Merge`.

deploy/helm/chart/values.yaml
```diff
@@ -29,24 +29,33 @@ statuslist:
...
-    # | -vault     | postgres,vault | PostgreSQL database with HashiCorp Vault / OpenBao token material storage | Vault-backed on-prem or multi-cloud deployments |
-    # | -fscert    | postgres | PostgreSQL database with filesystem-mounted token signing keys and certificates | Constrained / air-gapped environments using external secret injection |
+    # | Tag Suffix       | Cargo Features | Description | Primary Use Case |
+    # |------------------|----------------|-------------|------------------|
+    # | -aws             | postgres,aws | PostgreSQL database with AWS Secrets Manager and Route53 DNS support | AWS EKS native deployments |
+    # | -gcp             | postgres,gcp | PostgreSQL database with GCP Secret Manager and Cloud DNS support | GCP GKE native deployments |
+    # | -azure           | postgres,azure | PostgreSQL database with Azure Key Vault and Azure DNS support | Azure AKS native deployments |
+    # | -vault           | postgres,vault | PostgreSQL database with HashiCorp Vault / OpenBao token material storage | Vault-backed on-prem or multi-cloud deployments |
```
Hermann-Core, line 46: Since MySQL image builds were reverted from `deploy.yml`, the `-mysql-*` variants documented in this table are neither built nor published. Documenting 5 non-existent tags here adds unnecessary documentation bloat and misleads operators. Please remove this speculative table and replace it with a concise comment stating that MySQL deployments currently require explicitly supplying `statuslist.image.tag` or `statuslist.image.digest` until MySQL image variants are released.

helm/README.md
```diff
@@ -13,19 +13,24 @@ This guide shows you how to deploy the Status List Server on Kubernetes with the
...
+The server is published as provider-specific PostgreSQL image variants. The chart can also derive MySQL-capable image tags for deployments that provide such an image explicitly or after a release publishes them.
+
+| Image suffix      | Database   | Signing-credential backend             | Best for                             |
+| ----------------- | ---------- | -------------------------------------- | ------------------------------------ |
+| `-aws`            | PostgreSQL | AWS Secrets Manager + Route53 DNS-01   | Running on EKS / using AWS           |
+| `-gcp`            | PostgreSQL | GCP Secret Manager + Google Cloud DNS  | Running on GKE / using GCP           |
+| `-azure`          | PostgreSQL | Azure Key Vault + Azure DNS            | Running on AKS / using Azure         |
+| `-vault`          | PostgreSQL | HashiCorp Vault / OpenBao KV v2        | Operating your own Vault             |
+| `-fscert`         | PostgreSQL | File-based signing key and certificate | Delivering signing material as files |
+| `-mysql-aws`      | MySQL      | AWS Secrets Manager + Route53 DNS-01   | Requires a MySQL image release       |
```
Hermann-Core: Similar to `values.yaml`, this table documents 5 speculative `-mysql-*` image variants that are not built or published by the repository. Please remove these rows to keep documentation focused on active, published variants and avoid operator confusion.

.github/workflows/scheduled-image-scan.yml
```diff
@@ -113,7 +113,7 @@ jobs:
       contents: read
       packages: read
     strategy:
-      # One variant's registry trouble must not withhold the other four's findings.
+      # One variant's registry trouble must not withhold the other variants' findings.
```
Hermann-Core: This workflow still scans the 5 fixed PostgreSQL variants defined in `VARIANTS: aws gcp azure vault fscert`. Changing comments from 'five' to 'variants' without actual matrix changes adds unnecessary review churn. Please revert these comment edits or keep them consistent with the active matrix.

Hermann-Core (changes requested):
@Ngha-Boris I have reviewed the changes and posted inline comments for your attention.

helm/chart/templates/deployment.yaml
```diff
@@ -79,8 +86,13 @@ spec:
           {{- end }}
           {{- $pinned := ne $digest "" }}
           {{- $variant := $image.variant | default "fscert" }}
-          {{- $baseAppVersion := regexReplaceAll "-(aws|gcp|azure|vault|fscert)$" $.Chart.AppVersion "" }}
-          {{- $effectiveTag := $image.tag | default (printf "%s-%s" $baseAppVersion $variant) }}
+          {{- $baseAppVersion := regexReplaceAll "-(mysql-)?(aws|gcp|azure|vault|fscert)$" $.Chart.AppVersion "" }}
+          {{- $imageDatabaseBackend := include "status-list-server-chart.dbBackend" . }}
+          {{- $defaultVariantTag := printf "%s-%s" $baseAppVersion $variant }}
+          {{- if eq $imageDatabaseBackend "mysql" }}
```
Hermann-Core: Only images supporting PostgreSQL are published on GHCR; no MySQL images are published or planned. The chart must not synthesize or derive a `-mysql-<variant>` tag pointing to GHCR, as doing so leads to `ImagePullBackOff`. Please remove the `-mysql-` regex replacement and tag derivation logic here. Operators using MySQL must explicitly supply their own image via `statuslist.image.repository` and `statuslist.image.tag`.

.github/workflows/render-helm-templates/action.yml
```diff
@@ -84,6 +90,13 @@ runs:
         if ! grep -q "image: \"ghcr.io/adorsys/status-list-server:${EXPECTED_AWS_TAG}\"" /tmp/rendered/aws/status-list-server/templates/deployment.yaml; then
           echo "ERROR: AWS overlay must derive the AWS image variant from chart appVersion"; exit 1
         fi
+        helm template status-list-server helm/chart \
```
Hermann-Core: Because no MySQL images are published on GHCR, asserting that a MySQL render produces `ghcr.io/adorsys/status-list-server:${EXPECTED_MYSQL_TAG}` enforces a non-existent GHCR artifact. Please remove `EXPECTED_MYSQL_TAG` and this template verification step.

scripts/verify-image-reference.sh
```diff
@@ -83,6 +83,13 @@ expect "${repo}:${app_version}|Always|"
 # Variant derives a suffixed tag from the chart appVersion without duplicating the release version.
 expect "${repo}:${crate_version}-aws|Always|" \
     --set-string statuslist.image.variant=aws
+expect "${repo}:${crate_version}-mysql-fscert|Always|" \
```
Hermann-Core: GHCR only publishes PostgreSQL image variants. Please remove these assertions expecting `${repo}:${crate_version}-mysql-*` tags.

tests/helm_sensitive_env.rs
```diff
@@ -147,28 +173,212 @@ fn rendered_chart_uses_split_database_credentials() {
...
+
+    for expected in [
+        "name: wait-for-db",
+        "until nc -z status-list-server-mysql.statuslist.svc.cluster.local 3306; do",
+        "name: APP_DATABASE__BACKEND\n              value: \"mysql\"",
+        "name: APP_DATABASE__HOST\n              value: \"status-list-server-mysql.statuslist.svc.cluster.local\"",
+        "name: APP_DATABASE__PORT\n              value: \"3306\"",
+        "name: APP_DATABASE__USERNAME\n              value: \"mysql\"",
+        "name: APP_DATABASE__NAME\n              value: \"status-list\"",
+        "image: \"ghcr.io/adorsys/status-list-server:1.0.0-mysql-fscert\"",
```
Hermann-Core: Please remove this assertion expecting `ghcr.io/adorsys/status-list-server:1.0.0-mysql-fscert`. GHCR only publishes PostgreSQL images; the chart should not assert a non-existent GHCR MySQL tag.

docs/database-backends.md
```diff
@@ -47,6 +49,10 @@ Best default for production deployments. PostgreSQL is the safest choice when yo
 
 Good fit when your infrastructure already standardizes on MySQL-compatible services or when you want a production database with familiar operational patterns. For MariaDB, use this same backend setting because the driver path is shared.
 
+The Helm chart provides external MySQL connection defaults under `mysql:`. It does not deploy MySQL.
+
```
Hermann-Core, line 53: Please remove this paragraph discussing derived `mysql-<variant>` tags such as `1.0.1-mysql-fscert`. Clarify instead that published GHCR images currently only support PostgreSQL, and operators deploying with MySQL must build and supply their own container image.

deploy/helm/README.md
```diff
@@ -13,19 +13,24 @@ This guide shows you how to deploy the Status List Server on Kubernetes with the
...
+If `statuslist.image.tag` and `statuslist.image.digest` are both empty, the chart derives a tag from the chart appVersion, `statuslist.image.variant`, and the active database backend. PostgreSQL keeps the historical shape (`<version>-fscert`); MySQL derives `<version>-mysql-fscert`. Existing chart appVersions shipped before MySQL image variants were published can therefore render a MySQL tag that does not exist yet, causing `ImagePullBackOff`; pin `statuslist.image.tag` or `statuslist.image.digest` to a MySQL-capable image until a release publishes that derived tag. Cloud-specific variants, including `aws`, are selected explicitly through values overlays such as [`chart/values-aws.yaml`](chart/values-aws.yaml) and [`chart/values-production.yaml`](chart/values-production.yaml).
```
Hermann-Core, line 29: Please remove references to derived `<version>-mysql-fscert` GHCR tags and `ImagePullBackOff`. GHCR only publishes PostgreSQL images.

helm/chart/values.yaml
```diff
@@ -29,24 +29,33 @@ statuslist:
...
-    # Used only when tag and digest are empty. The chart derives
-    # <appVersion-without-suffix>-<variant>, so overlays can select a cloud
-    # variant without hardcoding the release version.
+    # Used only when tag and digest are empty. The chart derives the
+    # database-specific release tag from appVersion and this provider variant, so
+    # overlays can select a cloud variant without hardcoding the release version.
     variant: fscert
-    # Image tag with variant suffix. The release workflow publishes 5 multi-arch variants:
+    # Image tag with variant suffix. The current release workflow publishes the
+    # PostgreSQL variants below. MySQL tags follow the same provider suffix shape
```
Hermann-Core: Please remove references to MySQL image tags and the 5 speculative `-mysql-*` variants table. GHCR only publishes the 5 PostgreSQL variants (`-aws`, `-gcp`, `-azure`, `-vault`, `-fscert`).

Hermann-Core (commented):
@Ngha-Boris Additional review note regarding GHCR images:

Only images with PostgreSQL support are published to GHCR. Images providing MySQL support are not published on GHCR and are not currently planned. All derivations, CI checks, test assertions, and documentation referencing published MySQL GHCR image tags (`*-mysql-*`) must be removed to reflect the actual published state. Operators deploying with `APP_DATABASE__BACKEND=mysql` must build and supply their own image via `statuslist.image.repository` / `statuslist.image.tag`.

Ngha-Boris:
@Hermann-Core please recheck

helm/chart/templates/_helpers.tpl
```diff
@@ -88,11 +88,106 @@ provider). Preference: explicit statuslist.aws.region, then the legacy secretSto
...
+{{- define "status-list-server-chart.databasePasswordSecretKey" -}}
+{{- $item := .item | default dict }}
+{{- $mount := .mount | default dict }}
+{{- $key := $item.key | default "postgres-password" }}
+{{- $isDefaultDatabaseMount := and
+      (eq ($mount.name | default "") "database-credentials")
+      (eq ($mount.secretName | default "") (include "status-list-server-chart.appSecretName" .root))
+      (eq ($mount.mountPath | default "") "/var/run/status-list-server/database")
+      (eq ($item.path | default "") "password") }}
+{{- if and $isDefaultDatabaseMount (eq $key "postgres-password") (eq (include "status-list-server-chart.dbBackend" .root) "mysql") -}}
```
martcpp: This quietly changes `key: postgres-password` to `database-password` for MySQL, which breaks existing MySQL installs that manage their own Secret.

Repro: `fallbackSecret.enabled=false`, `externalSecret.enabled=false`, `APP_DATABASE__BACKEND=mysql`. `develop` mounts `postgres-password`; this branch mounts `database-password`. If the Secret only has `postgres-password`, the pod sits in `ContainerCreating` forever, and nothing fails at render time because the key check only runs in ESO mode.

It also means the value an operator writes isn't the value that gets rendered, and it relies on matching four hardcoded fields. Rename the mount path and it silently stops applying.

Suggest dropping the rewrite and adding an explicit value such as `statuslist.database.passwordSecretKey` (default `postgres-password`), with docs telling MySQL users to set `database-password`. If you keep the rewrite, only apply it when `fallbackSecret.enabled=true`, the one case where the chart knows both keys exist. Either way, please add a test for the repro above.

helm/chart/templates/deployment.yaml
```diff
@@ -2,6 +2,13 @@
 {{- if and .Values.externalSecret.enabled .Values.statuslist.fallbackSecret.enabled }}
 {{- fail "externalSecret.enabled and statuslist.fallbackSecret.enabled cannot both be true: choose either External Secrets Operator or the fallback Secret for secret delivery, not both" }}
 {{- end }}
+{{- $databaseBackend := include "status-list-server-chart.dbBackend" . }}
+{{- if hasKey (.Values.mysql | default dict) "enabled" }}
```
martcpp: `hasKey` means `mysql.enabled: false` also fails the render. Overlays often disable components explicitly, so people will hit this. Check the value instead:
```gotemplate
{{- if (.Values.mysql | default dict).enabled }}
```

helm/chart/templates/external-secrets.yaml
```diff
@@ -5,7 +5,38 @@
 {{- fail "externalSecret.enabled requires secretStore.enabled: no SecretStore would be rendered, leaving the ExternalSecret with a dangling secretStoreRef" }}
 {{- end }}
 {{- if ne .Values.externalSecret.spec.target.name (include "status-list-server-chart.appSecretName" .) }}
-{{- fail (printf "externalSecret.spec.target.name must be '%s' to match the single application-secret name used by the Deployment and PostgreSQL (postgres.auth.existingSecret)" (include "status-list-server-chart.appSecretName" .)) }}
+{{- fail (printf "externalSecret.spec.target.name must be '%s' to match the single application-secret name used by the Deployment and database subcharts" (include "status-list-server-chart.appSecretName" .)) }}
+{{- end }}
+{{- $appSecretName := include "status-list-server-chart.appSecretName" . }}
+{{- $esoKeys := dict }}
+{{- $targetTemplate := .Values.externalSecret.spec.target.template | default dict }}
+{{- $hasTargetTemplate := hasKey (.Values.externalSecret.spec.target | default dict) "template" }}
```
martcpp: False positive: `externalSecret.spec.target.template: {}` fails with `must emit key "postgres-password"`, even though the default `spec.data` does emit it.

The cause is a mismatch between `hasKey` here and the `{{- with ...template }}` that renders the block, which skips an empty dict. So this check assumes Replace mode while the real ExternalSecret has no template and keeps the data keys. `templateFrom` is also ignored, so ConfigMap-based templates hit the same error.

Suggest `$hasTargetTemplate := not (empty $targetTemplate)`, and skip the check when `templateFrom` is set.

helm/README.md
```diff
@@ -352,7 +353,7 @@ helm upgrade --install statuslist helm/chart \
...
+The chart bundles PostgreSQL and an OpenTelemetry collector. To point at an external database, disable the bundled PostgreSQL subchart and set the split `APP_DATABASE__*` fields under `statuslist.env`. For MySQL, set `postgres.enabled=false` and `statuslist.env.APP_DATABASE__BACKEND=mysql`; if host, port, username, or database name are omitted, the chart defaults them from the `mysql:` values block (`<release>-mysql.<namespace>.svc.cluster.local`, port `3306`, `mysql.auth.username`, and `mysql.auth.database`). This chart does not vendor a MySQL subchart, so provide that MySQL Service through your platform, operator, or an overlay. Published GHCR images currently only support PostgreSQL, so MySQL deployments must build and supply their own container image through a non-GHCR `statuslist.image.repository` plus `statuslist.image.tag` or `statuslist.image.digest`. In chart `0.4.0`, PostgreSQL still defaults to the legacy `postgres-password` mount key; a future chart version that switches PostgreSQL to `database-password` must call that out in its release notes.
```
martcpp: `Chart.yaml` is still `0.4.0`, but this PR adds new render failures that break existing values: `APP_DATABASE__BACKEND=sqlite` or `memory` now fails (it rendered on `develop`), `mysql.enabled` fails, `postgres.enabled=true` with a non-postgres backend fails, and MySQL now requires a non-GHCR image. Please bump the chart version, update the `0.4.0` references here and in `docs/database-backends.md`, and list these in the release notes. If rejecting sqlite/memory is intentional, say so explicitly.

deploy/helm/chart/values.yaml
```diff
@@ -480,6 +490,17 @@ postgres:
   image:
     tag: "18.0"
 
+mysql:
```
martcpp, line 507: The issue no longer matches the code, but every box is ticked. The issue asks for a versioned `mysql:` block (MySQL 8.4), documented versions (PostgreSQL 17, while the chart ships 18.0), and says `--set statuslist.env.APP_DATABASE__BACKEND=mysql` should render. That exact command now fails by design. External-only MySQL is the right scope, but please update the issue and its AC to describe what was built, or move the rest to a follow-up issue.

helm/chart/templates/network-policy.yaml
```diff
@@ -27,13 +27,13 @@ spec:
...
-      # Scope internal egress to the target pods, not "any pod in the cluster".
-      # PostgreSQL is always a dependency. Additional destinations can be appended via
+      # Scope internal egress to the target database pods, not "any pod in the cluster".
+      # Additional destinations can be appended via
       # statuslist.networkPolicy.egressInternal.
       to:
         - podSelector:
             matchLabels:
-              app.kubernetes.io/name: postgres
+              app.kubernetes.io/name: {{ include "status-list-server-chart.dbPodSelectorName" . }}
```
martcpp: For MySQL, the chart says the database is external, but this rule only allows egress to pods labeled `app.kubernetes.io/name: mysql` in the release namespace. That matches nothing for RDS or Cloud SQL, MySQL in another namespace, or operators such as Oracle's MySQL Operator or Percona that use different labels.

With the NetworkPolicy enabled, `wait-for-db` then loops forever with no error. When the chart doesn't deploy the database, I'd rather require an explicit egress target (for example `statuslist.networkPolicy.databaseEgress`) than guess a label.

helm/chart/templates/_helpers.tpl
```diff
@@ -88,11 +88,106 @@ provider). Preference: explicit statuslist.aws.region, then the legacy secretSto
...
+Database host helper: returns the configured host, or the default in-cluster
+service name for the active backend.
+*/}}
+{{- define "status-list-server-chart.dbHost" -}}
+{{- $env := .Values.statuslist.env | default dict }}
+{{- $backend := include "status-list-server-chart.dbBackend" . }}
+{{- if get $env "APP_DATABASE__HOST" }}
+{{- get $env "APP_DATABASE__HOST" }}
+{{- else if eq $backend "mysql" }}
+{{- printf "%s-mysql.%s.svc.cluster.local" .Release.Name .Release.Namespace }}
```
martcpp: This defaults to `<release>-mysql.<ns>`, a Service the chart never creates, so in most real installs it's wrong and shows up as an init container that never finishes. For the external-only MySQL path, failing with "set `APP_DATABASE__HOST`" would be clearer.

helm/chart/templates/deployment.yaml
```diff
@@ -203,14 +211,21 @@ spec:
             {{- if not (hasKey $env "APP_DATABASE__BACKEND") }}
               {{- $_ := set $env "APP_DATABASE__BACKEND" "postgres" }}
             {{- end }}
+            {{- $backend := include "status-list-server-chart.dbBackend" . }}
+            {{- if and (ne $backend "postgres") (ne $backend "mysql") }}
+              {{- fail (printf "statuslist.env.APP_DATABASE__BACKEND must be either postgres or mysql for this chart, got %q" $backend) }}
+            {{- end }}
             {{- if not (hasKey $env "APP_DATABASE__HOST") }}
-              {{- $_ := set $env "APP_DATABASE__HOST" (printf "%s-postgres.%s.svc.cluster.local" .Release.Name .Release.Namespace) }}
+              {{- $_ := set $env "APP_DATABASE__HOST" (include "status-list-server-chart.dbHost" .) }}
```
martcpp: The helpers and the Deployment disagree on empty strings:
- `APP_DATABASE__HOST: ""` → the init container waits on `r-postgres...`, but the app gets `APP_DATABASE__HOST: ""`.
- `APP_DATABASE__BACKEND: ""` → the chart assumes postgres everywhere, but the app gets an empty backend.

Resolve each field once through its helper and always write that result into `$env`, so the env, init container and NetworkPolicy can't diverge. That would also let `dbHost`, `dbPort`, `dbUsername` and `dbName` collapse into one helper, instead of four copies of the same `fail`.

helm/chart/templates/deployment.yaml
```diff
@@ -80,7 +87,11 @@ spec:
           {{- $pinned := ne $digest "" }}
           {{- $variant := $image.variant | default "fscert" }}
           {{- $baseAppVersion := regexReplaceAll "-(aws|gcp|azure|vault|fscert)$" $.Chart.AppVersion "" }}
-          {{- $effectiveTag := $image.tag | default (printf "%s-%s" $baseAppVersion $variant) }}
+          {{- $defaultVariantTag := printf "%s-%s" $baseAppVersion $variant }}
+          {{- $effectiveTag := $image.tag | default $defaultVariantTag }}
+          {{- if and (eq (include "status-list-server-chart.dbBackend" .) "mysql") (or (eq ($image.repository | default "") "ghcr.io/adorsys/status-list-server") (and (not $image.tag) (not $digest))) }}
```
martcpp: This is an exact string match, so it's easy to get past: `ghcr.io/adorsys/status-list-server/` (trailing slash) passes, and so does any pull-through mirror of the same PostgreSQL-only image, which then crashes at runtime. The chart can't know which features an image was built with. I'd keep only the "tag or digest required" part, put the GHCR note in the docs, and have the app fail clearly at startup when the configured backend's feature isn't compiled in.

helm/chart/values.yaml
```diff
@@ -496,6 +517,10 @@ externalSecret:
       name: statuslist-secret
       creationPolicy: Owner
     data:
+      - secretKey: database-password
```
martcpp: Both entries fetch the same `POSTGRES_PASSWORD` property, which doubles provider API calls on every refresh for every release (and AWS/GCP bill per call). One `data` entry plus `target.template.data` that emits both keys does the same job with a single fetch, like `values-rotation-example.yaml` already does.

scripts/verify-image-reference.sh
```diff
@@ -101,3 +101,20 @@ if helm template status-list-server helm/chart -s templates/deployment.yaml \
     exit 1
 fi
 echo "ok: malformed digest rejected at template time"
+
+if helm template status-list-server helm/chart -s templates/deployment.yaml \
+    --set postgres.enabled=false \
+    --set-string statuslist.env.APP_DATABASE__BACKEND=mysql > /dev/null 2>&1; then
```
martcpp: `> /dev/null 2>&1` plus "did it fail?" means any template error makes these checks pass. Grep stderr for the expected message instead. Related: the MySQL render added to `render-helm-templates/action.yml` only checks the exit code, so it's worth grepping for the host, port, key and podSelector there too.

helm/chart/values-external-secrets.yaml
```diff
@@ -144,3 +153,8 @@ postgres:
     existingSecret: statuslist-secret
     secretKeys:
       passwordKey: postgres-password
+
+mysql:
```
martcpp: This is the same as the defaults in `values.yaml`. Can be removed.

.github/workflows/scheduled-image-scan.yml
```diff
@@ -113,7 +113,7 @@ jobs:
       contents: read
       packages: read
     strategy:
-      # One variant's registry trouble must not withhold the other four's findings.
+      # One of the five variants' registry trouble must not withhold the other variants' findings.
```
martcpp: This wording change (and the matching one in `docs/supply-chain.md`) is unrelated to this issue. Please revert. The extra padding in the README image-variant table is also left over from the removed MySQL variants.

martcpp (changes requested):

Ngha-Boris:
Hello @martcpp please recheck

martcpp (dismissed):

helm/chart/templates/deployment.yaml
Hermann-Core: This was marked resolved, but line 6 still checks `hasKey (.Values.mysql | default dict) "enabled"`. Setting `mysql.enabled: false` in an environment overlay will still trigger `fail`. Please evaluate the boolean value instead of key existence:
```gotemplate
{{- if (.Values.mysql | default dict).enabled }}
```

helm/chart/templates/external-secrets.yaml
Hermann-Core: This was marked resolved without changes. `hasKey` still evaluates to true when `spec.target.template` is empty (`{}`), causing the check to assume `Replace` mode and fail when keys are defined in `spec.data`. Please use `not (empty $targetTemplate)` and skip the check when `templateFrom` is set.

helm/chart/templates/network-policy.yaml
Hermann-Core: This was marked resolved without changes. Because MySQL is not deployed in-cluster by this chart, restricting egress to `app.kubernetes.io/name: mysql` in `.Release.Namespace` matches no pods for external instances (RDS, Cloud SQL, or external namespaces). The `wait-for-db` probe will be blocked when network policies are enforced. Provide a configurable egress target or bypass this in-namespace pod selector for external backends.

helm/chart/templates/_helpers.tpl
Hermann-Core: This was marked resolved without changes. Defaulting `dbHost` to `<release>-mysql.<namespace>.svc.cluster.local` creates a reference to a Service this chart does not deploy, causing `wait-for-db` to poll indefinitely. For external MySQL, fail explicitly at template time when `APP_DATABASE__HOST` is missing.

helm/chart/templates/deployment.yaml
Hermann-Core: This was marked resolved without changes. When `APP_DATABASE__HOST` is set to an empty string (`""`), `hasKey` preserves the empty string in the container env while `dbHost` falls back to the in-cluster service name for `wait-for-db`. Please resolve database connection fields once via helpers, write the resolved values to `$env`, and consolidate the repeated `fail` blocks across the helpers.

helm/chart/values.yaml
Hermann-Core: This was marked resolved without changes. Lines 520-527 still define two distinct `data` entries fetching the same `POSTGRES_PASSWORD` property. Use a single `data` entry with `target.template.data` to emit both `database-password` and `postgres-password` from one remote fetch.

scripts/verify-image-reference.sh
Hermann-Core: This was marked resolved without changes. Directing stderr to `/dev/null` and asserting only non-zero exit codes causes the negative checks to pass on any unrelated template rendering failure. Grep stderr for the specific expected failure message instead.

helm/chart/values-external-secrets.yaml
Hermann-Core: This was marked resolved without changes. Lines 157-161 duplicate the base defaults in `values.yaml` verbatim and should be removed.

.github/workflows/scheduled-image-scan.yml
Hermann-Core: While line 116 was reverted, lines 323, 331, and 952 of `scheduled-image-scan.yml` and `docs/supply-chain.md` still contain the wording modifications. Please revert those edits to keep the diff focused.

helm/README.md
Hermann-Core: This was marked resolved without changes. `Chart.yaml` remains at version `0.4.0` despite adding breaking rendering constraints on existing values. Please bump the chart version and document these changes in the release notes.

helm/chart/templates/_helpers.tpl
Hermann-Core: This was marked resolved without addressing the case where an unmanaged Secret (`fallbackSecret.enabled=false`, `externalSecret.enabled=false`) contains only `postgres-password`. The helper rewrites the mount key to `database-password`, which causes volume mount failures on existing clusters upon upgrade. Either preserve `postgres-password` when fallback secrets are disabled or provide an explicit key override.

Hermann-Core (changes requested):
@Ngha-Boris I have completed the re-review and replied to the open threads for your attention. Please check the comments that have been unresolved.

Ngha-Boris:
> @Ngha-Boris I have completed the re-review and replied to the open threads for your attention. Please check the comments that have been unresolved.

Hello @Hermann-Core all the comments you have addressed please recheck, please don't forget to pull before reviewing because most of the comments you unresolved were already addressed

martcpp (dismissed):
LGTM

Hermann-Core (dismissed):
@Ngha-Boris All outstanding review points have been addressed and verified across code and test suites. Looks good to merge.

Ngha-Boris:
Hello @Hermann-Core @martcpp please re-approve I merged the develop branch into this one.

martcpp (approved):

Hermann-Core (approved):

Hermann-Core:
@forkimenjeckayang can you recheck this so we merge?

forkimenjeckayang (approved):

PR #530, opened 2026-09-11
ci: add Trivy vulnerability database caching with staleness handling

### LINKED ISSUE #405: CI: cache Trivy vulnerability DB across runs  DB fetch is an unguarded dependency in the release path
## Summary

Every run of the `scan-image` job in `deploy.yml` downloads the Trivy vulnerability database
from scratch (~108 MiB, measured ~35s). The DB is cached within a job  three scan steps
share one fetch  but nothing persists it across runs.

## Impact

This is an availability problem, not a latency one.

`scan-image` gates `deploy`. So every production release depends on a third-party download
succeeding at release time. A network blip, an upstream rate limit, or a registry incident
fails the fetch, fails the job, and blocks the deploy.

The failure is also misleading. A DB-fetch failure surfaces as the security gate failing,
which reads as "a vulnerability was found" to whoever is on the release. Expect an hour lost
to diagnosing the wrong thing.

Latency is the secondary benefit; removing the hard dependency is the point.

## Proposed fix

Persist the Trivy cache directory across runs with `actions/cache`, so a failed fetch can
fall back to the last-known-good DB rather than failing the job.

`trivy-action` supports pointing at a cache dir. Design decisions to settle when this is
picked up:

- **Refresh cadence** — how stale is acceptable before a fetch failure should be fatal
- **Cache key** — date-based, or key-on-miss with a restore-key fallback
- **Stale-hit visibility** — a stale DB must be reported in the step summary, since a
  sufficiently old DB silently weakens the gate. Falling back quietly is its own failure mode.

The default posture should be: fetch failure degrades to a stale DB with a loud warning,
rather than blocking the release.

## Why this wasn't in the supply-chain PR

The cache-key design above is a small but real decision surface, and that PR already carries
the `cargo-auditable` change, the build assertion, the scan job, and the deploy gate. Folding
this in would have made it harder to review the parts with production blast radius.

## Acceptance criteria

- [ ] Trivy DB persists across workflow runs
- [ ] A DB-fetch failure degrades to the cached DB instead of failing the job
- [ ] Stale-DB use is visible in the step summary with the DB age
- [ ] Cache staleness bound is documented in `docs/supply-chain.md`

eta: 2days

martcpp:
- Add explicit actions/cache step for Trivy DB in deploy.yml and scheduled-image-scan.yml
- Add ensure-trivy-db.sh script to refresh DB or degrade to cached version with warnings
- Add .trivy.yaml config to skip DB updates during scan
- Document cache strategy and staleness bounds in supply-chain.md
- Cache key uses run ID with restore-keys fallback to most recent DB
- 24-hour staleness bound matches Trivy's up-to-date threshold
- Degrades with warning if cached DB within bound, hard-fails beyond bound


Ngha-Boris (dismissed):

scripts/ensure-trivy-db.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# Fetch failed. Check whether we have a usable cached DB.
+echo "::warning::Failed to refresh Trivy vulnerability database; checking for cached DB..."
+
+# The vulnerability DB lives in ${DB_DIR}/<schema>/trivy.db (or similar).
+# We look for any trivy.db file under the cache dir.
+if [ -d "${DB_DIR}" ] && find "${DB_DIR}" -name 'trivy.db' -type f | grep -q .; then
+  # Found a cached DB. Compute its age.
+  # We take the newest trivy.db as the effective DB.
+  newest_db=$(find "${DB_DIR}" -name 'trivy.db' -type f -printf '%T@ %p\n' | sort -nr | head -1 | cut -d' ' -f2-)
```
Hermann-Core: This line breaks on macOS and BSD systems. GNU `find`'s `-printf '%T@ %p\n'` is not supported by BSD `find` (`find: -printf: unknown primary or operator`). Under `set -euo pipefail`, the script terminates immediately with an error whenever the fallback branch executes.

Similarly on line 60, `stat -c '%Y'` is GNU-specific (BSD `stat` expects `-f '%m'`).

Trivy writes `metadata.json` alongside `trivy.db` in `${DB_DIR}` with `DownloadedAt` and `UpdatedAt` timestamps. If calculating mtime directly from the filesystem, consider a portable fallback:
```bash
if stat -c '%Y' "$newest_db" >/dev/null 2>&1; then
  db_mtime=$(stat -c '%Y' "$newest_db")
else
  db_mtime=$(stat -f '%m' "$newest_db")
fi
```
And for finding `trivy.db` without GNU `-printf`:
```bash
newest_db=$(find "${DB_DIR}" -name 'trivy.db' -type f | head -1)
```
(Trivy stores a single `trivy.db` per schema version in the cache dir).

.github/workflows/deploy.yml
```diff
@@ -218,7 +218,7 @@ jobs:
     permissions:
       contents: read
       packages: read
-      actions: read
+      actions: write
```
Hermann-Core: `actions: write` is unnecessary and violates the Principle of Least Privilege.

`actions/cache` does not authenticate using `GITHUB_TOKEN` permissions; it communicates with the internal GitHub Actions Cache Service via the runner's `ACTIONS_RUNTIME_TOKEN`. Granting `actions: write` to the job provides write access to the GitHub Actions REST API (e.g. canceling runs, approving deployments). This should remain `actions: read`.

.github/workflows/scheduled-image-scan.yml
```diff
@@ -112,6 +112,7 @@ jobs:
     permissions:
       contents: read
       packages: read
+      actions: write
```
Hermann-Core: Similar to `deploy.yml`, `actions: write` is not required for `actions/cache` (which uses the runner's internal Actions Cache Service token). No steps in `rescan` require GitHub Actions API write permissions. Please remove `actions: write` to adhere to least-privilege security standards.

.github/workflows/deploy.yml
```diff
@@ -283,6 +283,38 @@ jobs:
...
+          restore-keys: |
+            trivy-db-${{ runner.os }}-
+
+      - name: Ensure Trivy vulnerability database
+        # Refreshes the DB; on fetch failure, falls back to the cached DB with a
+        # loud warning (and step-summary entry) if the cached DB is within the
+        # 24h staleness bound. Hard-fails only if no DB exists at all.
+        run: bash scripts/ensure-trivy-db.sh ~/.cache/trivy 24
+
+      - name: Save Trivy vulnerability database cache
```
Hermann-Core, line 314: Two issues with this cache save step:

1. **Workflow Dispatch Cache Bypass:** `deploy.yml` only runs on push tags (`v*.*.*`) and `workflow_dispatch`. Restricting cache save to `github.event_name == 'push'` prevents manual dispatch runs (which maintainers use to exercise scans and warm pipeline caches) from updating the cache.
2. **Matrix Race Condition:** `scan-image` runs 5 matrix variants concurrently. All 5 legs execute with the identical key `trivy-db-${{ runner.os }}-${{ github.run_id }}`. All 5 will race to upload ~100MB of cache data, resulting in 4 legs logging `ReserveCacheError: Cache already exists`.

Consider gating the save step to a single matrix leg (e.g. `if: matrix.variant.suffix == 'aws'`) to avoid redundant uploads and collision errors.

.github/workflows/scheduled-image-scan.yml
```diff
@@ -214,6 +215,31 @@ jobs:
...
           done <<< "$resolved"
 
+      - name: Install Trivy
+        # Pinned to the same version trivy-action v0.36.0 uses internally.
+        uses: aquasecurity/setup-trivy@3fb12ec12f41e471780db15c232d5dd185dcb514 # v0.2.6
+        with:
+          version: v0.70.0
+          cache: false
+
+      - name: Restore Trivy vulnerability database cache
```
Hermann-Core, line 224: Using `actions/cache` (which performs an automatic post-job save) across a 5-variant matrix causes all 5 parallel legs to attempt saving to the same key `trivy-db-${{ runner.os }}-${{ github.run_id }}` upon job completion, triggering `ReserveCacheError` across 4 legs.

Consider adopting the pattern in `deploy.yml` by separating into `actions/cache/restore` and `actions/cache/save`, and gating the save step to a single deterministic matrix leg (`if: matrix.variant.suffix == 'aws'`).

.trivy.yaml
```diff
@@ -0,0 +1,2 @@
+db:
+  skip-update: true
```
Hermann-Core: Placing `.trivy.yaml` at the repository root with `db: skip-update: true` introduces unexpected side effects for local development. Any developer invoking `trivy` from the root of the workspace (e.g. `trivy fs .` or local container scans) will silently inherit `skip-update: true` by default.

`aquasecurity/trivy-action` natively supports `skip-db-update: 'true'` as a direct action input (or via `TRIVY_SKIP_DB_UPDATE: "true"` in `env:`). Using the action input directly eliminates the need for a root `.trivy.yaml` file and prevents local scan pollution.

Hermann-Core (changes requested):
@martcpp I have reviewed the changes and posted inline comments for your attention.

martcpp:
@Hermann-Core pls recheck

Hermann-Core (approved):

martcpp:
@Ngha-Boris recheck

Ngha-Boris (approved):

PR #532, opened 2026-09-11
fix(watcher): only rotate on real content changes, not read/open events

### LINKED ISSUE #529: File watcher triggers spurious credential rotation on read/open events (rotation spam)
## Objective

The file watcher reacts to `IN_OPEN`/access events emitted by the application's own reads of watched secret files, producing a self-sustaining rotation loop that spams logs and reconnects to the database roughly every 500ms even when no key rotation occurs. Make the watcher robust so it only reacts to genuine content changes, regardless of the event source.

## Deliverables

- A robust change-detection mechanism in `src/utils/file_watcher.rs` that verifies the watched file's content actually changed (e.g. content fingerprinting) before dispatching the rotation callback, rather than relying solely on event kinds.
- Unit tests covering spurious events (open/access/metadata) and genuine content changes (create/remove/modify-data/rename), including the Kubernetes `..data` symlink swap case.
- Keep or refine the existing event-kind filter so it is consistent with the content-verification approach.

## Acceptance Criteria

- [ ] The watcher does not trigger rotation when a watched file is merely read/opened or its metadata (e.g. atime) changes.
- [ ] The watcher triggers rotation exactly once when a watched file's content actually changes.
- [ ] A Kubernetes secret `..data` symlink swap (content change) triggers rotation.
- [ ] No rotation spam: running the app with a stable secret produces zero `rotation_started` / `file_change_detected` logs over a sustained interval.
- [ ] All existing and new `file_watcher` unit tests pass (`cargo test --lib utils::file_watcher`).

## Dependencies

_None_

Christiantyemele:
## Summary

Fixes the file-watcher rotation spam (issue #529). The `notify` crate's default inotify mask includes `IN_OPEN`, so the app's own reads of a watched secret file emitted events that created a self-sustaining rotation loop — spamming logs and reconnecting to the database roughly every 500ms even when no key rotation occurred.

## Root cause

`notify`'s default inotify watch mask (`WatchMask::OPEN`) fires an event every time a watched file is opened. The watcher's event handler only checked that the changed path matched, not the event kind. So reading the password file during rotation triggered another rotation → read → event → loop. The file itself was never modified (atime/ctime/mtime all stable).

## Fix (two layers)

1. **Event-kind filter** (`is_content_change`): the notify handler now only reacts to `Create`/`Remove`/`Rename`/`Modify(Data)`, ignoring `Access`/`Open` and `Modify(Metadata)` (e.g. atime updates).
2. **Content fingerprint gate** in the debouncer: the callback is only dispatched when the watched file's content actually changed, so even a misclassified event cannot trigger a rotation unless the content really changed.

## Tests

- `access_and_metadata_events_are_not_content_changes` — verifies the event-kind filter.
- `debouncer_skips_callback_when_content_unchanged` — spurious event with unchanged content does not trigger the callback.
- `debouncer_runs_callback_when_content_changed` — a real content change triggers the callback.
- Updated existing debouncer tests to use real files with changing content.

All 7 `file_watcher` unit tests pass; `cargo fmt` and `cargo clippy --lib` are clean.

Closes #529

### DIFF
--- a/src/utils/file_watcher.rs
+++ b/src/utils/file_watcher.rs
@@ -1,4 +1,5 @@
-use notify::{Event, RecommendedWatcher, RecursiveMode, Watcher as _};
+use notify::event::ModifyKind;
+use notify::{Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher as _};
 use std::collections::{HashMap, HashSet, hash_map::DefaultHasher};
 use std::hash::{Hash, Hasher};
 use std::path::{Path, PathBuf};
@@ -73,6 +74,14 @@ fn spawn_notify_watcher(
         let mut watcher = match RecommendedWatcher::new(
             move |event: notify::Result<Event>| match event {
                 Ok(event) => {
+                    // Only react to real content changes. The notify crate's
+                    // default inotify mask includes IN_OPEN, so the app's own
+                    // reads of the watched file would otherwise emit events and
+                    // create a self-sustaining rotation loop. Ignore access/open
+                    // and pure-metadata events (e.g. atime updates).
+                    if !is_content_change(&event.kind) {
+                        return;
+                    }
                     if event_matches(&event, &event_paths) {
                         let _ = tx_events.blocking_send(());
                     }
@@ -168,6 +177,13 @@ fn spawn_debouncer<F, Fut>(
     Fut: std::future::Future<Output = ()> + Send + 'static,
 {
     tokio::spawn(async move {
+        // Baseline content fingerprint. The callback is only dispatched when the
+        // watched file's content actually changes. This is the final safety net
+        // on top of the event-kind filter: even if a spurious event (e.g. a
+        // misclassified `Modify(Any)` or a kubelet `..data` touch) slips through,
+        // it cannot trigger a rotation unless the file content really changed.
+        let mut last_fingerprints = fingerprint_all(&paths).await;
+
         while rx.recv().await.is_some() {
             tracing::info!(
                 event = "file_change_detected",
@@ -176,6 +192,19 @@ fn spawn_debouncer<F, Fut>(
                 "watched secret file changed"
             );
             while matches!(tokio::time::timeout(DEBOUNCE, rx.recv()).await, Ok(Some(_))) {}
+
+            let current = fingerprint_all(&paths).await;
+            if current == last_fingerprints {
+                tracing::debug!(
+                    event = "file_change_ignored",
+                    rotation.target = target,
+                    paths = ?paths,
+                    "file change event did not alter content; skipping rotation"
+                );
+                continue;
+            }
+            last_fingerprints = current;
+
             let callback = callback.clone();
             callback().await;
         }
@@ -211,6 +240,24 @@ fn path_relevant(changed: &Path, watched: &Path) -> bool {
         || watched.starts_with(changed)
         || (changed.file_name().is_some_and(|name| name == "..data")
             && changed.parent() == watched.parent())
+}
+
+/// Whether an event kind represents an actual content change worth reacting to.
+///
+/// The `notify` crate's default inotify mask includes `IN_OPEN`/`IN_ACCESS`, so
+/// the application's own reads of a watched file emit events. Reacting to those
+/// would create a self-sustaining rotation loop (read -> event -> rotate ->
+/// read -> ...). Only `Create`/`Remove`/`Rename` and `Modify(Data)` indicate a
+/// real change; `Access` and `Modify(Metadata)` (e.g. atime updates) do not.
+fn is_content_change(kind: &EventKind) -> bool {
+    matches!(
+        kind,
+        EventKind::Create(_)
+            | EventKind::Remove(_)
+            | EventKind::Modify(ModifyKind::Data(_))
+            | EventKind::Modify(ModifyKind::Name(_))
+            | EventKind::Modify(ModifyKind::Any)
+    )
 }
 
 async fn fingerprint_all(paths: &[PathBuf]) -> HashMap<PathBuf, Option<u64>> {
@@ -270,10 +317,14 @@ mod tests {
 
     #[tokio::test]
     async fn debouncer_coalesces_bursty_changes() {
+        let dir = TempDir::new();
+        let path = dir.path.join("secret");
+        tokio::fs::write(&path, "v1").await.expect("write initial");
+
         let rotations = Arc::new(AtomicUsize::new(0));
         let seen = rotations.clone();
         let (tx, rx) = mpsc::channel(8);
-        let paths = Arc::new(vec![PathBuf::from("/tmp/secret")]);
+        let paths = Arc::new(vec![path.clone()]);
 
         spawn_debouncer(
             paths,
@@ -287,6 +338,11 @@ mod tests {
             "test",
         );
 
+        // Let the debouncer establish its baseline fingerprint.
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        // One content change followed by a burst of events coalesces into one callback.
+        tokio::fs::write(&path, "v2").await.expect("write update");
         tx.send(()).await.expect("first event");
         tx.send(()).await.expect("second event");
         tokio::time::sleep(Duration::from_millis(700)).await;
@@ -296,11 +352,15 @@ mod tests {
 
     #[tokio::test]
     async fn debouncer_serializes_callbacks() {
+        let dir = TempDir::new();
+        let path = dir.path.join("secret");
+        tokio::fs::write(&path, "v1").await.expect("write initial");
+
         let active = Arc::new(AtomicUsize::new(0));
         let max_active = Arc::new(AtomicUsize::new(0));
         let completed = Arc::new(AtomicUsize::new(0));
         let (tx, rx) = mpsc::channel(8);
-        let paths = Arc::new(vec![PathBuf::from("/tmp/secret")]);
+        let paths = Arc::new(vec![path.clone()]);
 
         spawn_debouncer(
             paths,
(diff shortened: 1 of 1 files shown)


src/utils/file_watcher.rs
```diff
@@ -168,6 +177,13 @@ fn spawn_debouncer<F, Fut>(
     Fut: std::future::Future<Output = ()> + Send + 'static,
 {
     tokio::spawn(async move {
+        // Baseline content fingerprint. The callback is only dispatched when the
+        // watched file's content actually changes. This is the final safety net
+        // on top of the event-kind filter: even if a spurious event (e.g. a
+        // misclassified `Modify(Any)` or a kubelet `..data` touch) slips through,
+        // it cannot trigger a rotation unless the file content really changed.
+        let mut last_fingerprints = fingerprint_all(&paths).await;
```
martcpp: Redue this comment

martcpp (changes requested):
Just reduce the comments whatntyencode ddoes it very easy to see thanks overla.otyer think are okay @Christiantyemele thanks

src/utils/file_watcher.rs
```diff
@@ -73,6 +74,11 @@ fn spawn_notify_watcher(
         let mut watcher = match RecommendedWatcher::new(
             move |event: notify::Result<Event>| match event {
                 Ok(event) => {
+                    // Only react to real content changes; the app's own reads
+                    // emit IN_OPEN events that would otherwise loop rotation.
+                    if !is_content_change(&event.kind) {
```
Hermann-Core, line 87: Filtering out `Access` events here is the key fix that breaks the self-sustaining feedback loop on Linux where the application reading secrets triggered immediate follow-up inotify events.

In Kubernetes environments with projected secrets (Vault/ESO via Secret volume mounts), kubelet updates secret projections by creating `..data_tmp` and atomically renaming it to `..data`. That operation generates `ModifyKind::Name(_)` and `Create(_)`, which will pass this filter as intended and be cleanly handled by the debouncer's content fingerprint check.

src/utils/file_watcher.rs
```diff
@@ -168,6 +174,10 @@ fn spawn_debouncer<F, Fut>(
     Fut: std::future::Future<Output = ()> + Send + 'static,
 {
     tokio::spawn(async move {
+        // Only dispatch when content actually changed; guards against spurious
+        // events that slip past the event-kind filter.
+        let mut last_fingerprints = fingerprint_all(&paths).await;
```
Hermann-Core, line 187: 1. **Simplification & Performance**: `fingerprint_all` uses a `JoinSet` to spawn tasks onto the Tokio runtime and collects into a `HashMap<PathBuf, Option<u64>>`. In `status-list-server`, `paths` is always a small slice (1–2 files: database password or cert+key). Spawning Tokio tasks, cloning `PathBuf` keys, and hashing map entries on every debounce settle and polling tick introduces unnecessary runtime scheduling and heap allocation overhead. Sequential reads into a `Vec<Option<...>>` would be simpler, faster, and eliminate ~20 lines of async task-joining ceremony.

2. **Security & Hardening**: `fingerprint` currently uses `std::collections::hash_map::DefaultHasher`, which produces a 64-bit integer (`u64`) with an unspecified algorithm (SipHash 1-3). A 64-bit hash has a birthday collision bound of $2^{32}$. If two secret revisions ever collide, rotation would be silently suppressed (`current == last_fingerprints`), leaving the server with stale credentials. Using SHA-256 (already available in the workspace) ensures collision resistance and deterministic cross-platform hashing.

src/utils/file_watcher.rs
```diff
@@ -176,6 +186,19 @@ fn spawn_debouncer<F, Fut>(
                 "watched secret file changed"
             );
             while matches!(tokio::time::timeout(DEBOUNCE, rx.recv()).await, Ok(Some(_))) {}
+
+            let current = fingerprint_all(&paths).await;
+            if current == last_fingerprints {
```
Hermann-Core, line 193: Currently, line 186 emits `INFO: watched secret file changed` immediately when *any* event arrives from the channel, *before* debouncing settles and *before* checking whether the content actually changed. If an event fires (e.g., directory metadata changes or an atime touch), the logs output:
```text
INFO: watched secret file changed
DEBUG: file change event did not alter content; skipping rotation
```
For external log aggregators (e.g. Datadog, Grafana Loki, CloudWatch) alerting on `file_change_detected`, this creates false-positive alerts indicating secret rotations when none occurred.

Moving the `INFO` log to line 201 (right before `callback().await`) ensures `INFO` is only emitted when genuine content change has been verified and rotation is actually executing.

src/utils/file_watcher.rs
```diff
@@ -213,6 +236,19 @@ fn path_relevant(changed: &Path, watched: &Path) -> bool {
             && changed.parent() == watched.parent())
 }
 
+/// True for real content changes (Create/Remove/Rename/Modify(Data)); false
+/// for Access/Open and metadata events that the app's own reads emit.
+fn is_content_change(kind: &EventKind) -> bool {
```
Hermann-Core, line 249: There is an edge case with relative paths that can disable native inotify:
1. In `watch_roots`: for relative paths like `PathBuf::from("server-cert.pem")`, `path.parent()` returns `Some(Path::new(""))`. Calling `watcher.watch("", ...)` fails with `No path was found. about [""]`, setting `all_watched = false` (line 125) and forcing the server to disable inotify and fall back to polling.
2. In `path_relevant`: if `watched` is relative (`watched.parent() == Some("")`) and `changed` is absolute (`/etc/secrets/..data`), `changed.parent() == watched.parent()` evaluates to `false`.

Consider normalizing paths to absolute paths in `FileWatcher::new` without resolving symlinks:
```rust
let paths = paths
    .into_iter()
    .map(|p| if p.is_absolute() { p } else { std::env::current_dir().map(|cwd| cwd.join(&p)).unwrap_or(p) })
    .filter(|p| unique.insert(p.clone()))
    .collect();
```

src/utils/file_watcher.rs
```diff
@@ -325,15 +374,122 @@ mod tests {
...
+        // A real content change must trigger the callback.
+        tokio::fs::write(&path, "new").await.expect("write update");
+        tx.send(()).await.expect("send event");
+        tokio::time::sleep(Duration::from_millis(1200)).await;
+
+        assert_eq!(rotations.load(Ordering::SeqCst), 1);
+    }
+
+    #[test]
+    fn access_and_metadata_events_are_not_content_changes() {
```
Hermann-Core, line 457: The unit tests here nicely cover event kinds and debouncer timing. To ensure long-term regression safety for Kubernetes deployments, consider also adding a test that creates real directory symlinks (`..data -> ..dir1`), performs an atomic symlink swap (`..data_tmp -> ..dir2` then rename to `..data`) with unchanged content (verifying 0 callbacks), followed by a swap with changed content (verifying exactly 1 callback).

Hermann-Core (commented):
Hi @Christiantyemele,

Thanks for putting together this PR! I've reviewed your changes and verified the behavior against our live Kubernetes replication environment (with HashiCorp Vault and External Secrets Operator).

The dual-layer mechanism (event-kind filtering + debouncer content fingerprinting) correctly resolves the core rotation storm caused by inotify `Access` read/open loops on Linux. I've left a few inline comments on log ordering, async task simplification, cryptographic hashing, and relative path handling for your attention.

src/utils/file_watcher.rs
Christiantyemele: Thanks — trimmed. The baseline-fingerprint comment is now a single short note (the full reasoning lives in the PR description), and the other inline comments were cut down to one or two short lines each.

src/utils/file_watcher.rs
Christiantyemele: Agreed — filtering out Access/Open is what breaks the self-sustaining loop, and the Create/Modify(Name) events from the kubelet ..data_tmp -> ..data rename pass the filter and are settled by the debouncer fingerprint gate. That k8s path is also covered by the new fingerprint_tracks_content_through_atomic_symlink_swap test.

src/utils/file_watcher.rs
Christiantyemele: Both applied in 9c0f6df. fingerprint_all now does sequential reads into a Vec<Option<[u8; 32]>> (no JoinSet/task spawning), and fingerprint uses sha2::Sha256 instead of the 64-bit DefaultHasher, removing the 2^32-order birthday collision risk.

src/utils/file_watcher.rs
Christiantyemele: Fixed in 9c0f6df. The INFO `file_change_detected` log now fires only after the fingerprint comparison confirms the content actually changed, right before the rotation callback runs — no more false-positive rotation alerts from spurious events.

src/utils/file_watcher.rs
Christiantyemele: Fixed in 9c0f6df. FileWatcher::new now joins relative paths onto the current working directory (without resolving symlinks) before dedup, so watch_roots never produces an empty path and path_relevant compares absolute parents consistently.

src/utils/file_watcher.rs
Christiantyemele: Added fingerprint_tracks_content_through_atomic_symlink_swap in 9c0f6df: it builds a real ..data -> ..dir1 tree, performs an atomic ..data_tmp -> ..data swap with identical content (fingerprint unchanged) and then a swap to changed content (fingerprint differs). The 0/1-callback gating on that fingerprint is already asserted by debouncer_skips_callback_when_content_unchanged and debouncer_runs_callback_when_content_changed.

src/utils/file_watcher.rs
```diff
@@ -350,6 +501,42 @@ mod tests {
         ));
     }
 
+    #[tokio::test]
+    async fn fingerprint_tracks_content_through_atomic_symlink_swap() {
+        use std::os::unix::fs::symlink;
```
Hermann-Core, line 507: **Nit (Portability):**
Since `std::os::unix::fs::symlink` is Unix-specific, consider adding `#[cfg(unix)]` to this test function so non-Unix CI/targets don't hit an unresolved import:
```rust
#[tokio::test]
#[cfg(unix)]
async fn fingerprint_tracks_content_through_atomic_symlink_swap() {
    use std::os::unix::fs::symlink;
    // ...
}
```

Hermann-Core (dismissed):
Hi @Christiantyemele, excellent work addressing all the review points! Left one minor non-blocking nit regarding `#[cfg(unix)]` on the new symlink test. Overall, this is solid, well-structured, and ready to go. LGTM!

src/utils/file_watcher.rs
Hermann-Core: Since `std::os::unix::fs::symlink` is Unix-specific, consider adding `#[cfg(unix)]` to this test function so non-Unix CI/targets don't hit an unresolved import.

Hermann-Core (commented):
..

src/utils/file_watcher.rs
Christiantyemele: Applied in 3478cc1 — added `#[cfg(unix)]` above `#[tokio::test]` on fingerprint_tracks_content_through_atomic_symlink_swap so non-Unix targets skip the test instead of failing on the `std::os::unix::fs::symlink` import.

Hermann-Core (approved):

martcpp (approved):

PR #554, opened 2026-09-16
feat(cache): implement Redis cache support and update configuration options

### LINKED ISSUE #167: Implement LRU status-list cache with in-memory and Redis backends
## Objective

Provide a pluggable LRU cache for recently used status-list records, with the backend chosen at compile time via Cargo features (`memory` or `redis`).

## Context

The server currently keeps status-list records in an in-process `moka` cache. For HA / multi-replica deployments, we need a cache abstraction that can be backed either by an in-process LRU cache or by Redis, selected when the binary is built.

## Deliverables

- [ ] Define a `Cache` trait exposing `get`, `put`, and `invalidate(list_id)` under a new domain ports module.
- [ ] `cache-memory` feature: LRU-capable in-memory implementation using `moka`.
- [ ] `cache-redis` feature: Redis-backed implementation with configurable TTL and invalidation.
- [ ] Ensure only one backend is active at a time; fail compilation if both or default to memory.
- [ ] Replace the current direct `moka` usage in `get_status_list.rs` and `update_status.rs` with the trait.
- [ ] Preserve the existing cache-invalidation semantics on `PATCH` updates.
- [ ] Add Prometheus-style hit/miss metrics for the cache.
- [ ] Add unit/integration tests for both backends.
- [ ] Update `docker-compose.yml`, `.env.template`, and Helm values with the new cache options

1 day

Ngha-Boris:
**Summary**

Implements a pluggable status-list cache selected at compile time via Cargo features.

**Changes**

- Added a domain cache port: `domain::ports::cache::Cache`.
- Added cache backend features:
  - `cache-memory` default backend using `moka`.
  - `cache-redis` backend using Redis with TTL and key prefixing.
- Added compile-time guards so exactly one cache backend is enabled.
- Wired cache construction through the composition root.
- Preserved existing PATCH invalidation semantics.
- Shared Prometheus-style cache hit/miss metrics across both backends.
- Added Redis config:
  - `APP_CACHE__REDIS_URL`
  - `APP_CACHE__REDIS_KEY_PREFIX`
- Updated `.env.template`, `docker-compose.yml`, and Helm values.
- Added backend tests, including Redis tests that run live when `TEST_REDIS_URL` is set.

**Validation**

- `cargo test --features cache-memory --no-default-features --features memory`
- `cargo test --no-default-features --features memory,cache-redis --no-run`
- `cargo test --no-default-features --features memory,cache-redis outbound::cache::redis_tests`
- Confirmed `cache-memory + cache-redis` fails compilation intentionally with the mutual-exclusion error.


src/outbound/cache.rs
```diff
@@ -100,6 +100,181 @@ impl StatusListCache for MokaStatusListCache {
...
+#[cfg(feature = "cache-redis")]
+impl RedisStatusListCache {
+    pub async fn new(
+        redis_url: &str,
+        ttl_secs: u64,
+        max_capacity: u64,
+        key_prefix: impl Into<String>,
+    ) -> Result<Self, StatusListError> {
+        let client = redis::Client::open(redis_url).map_err(redis_error)?;
+        let connection = redis::aio::ConnectionManager::new(client)
```
martcpp: a slow or unreachable Redis hangs every status-list read.
`ConnectionManager::new` uses the redis-rs defaults, and in 0.32.7 both `response_timeout` and `connection_timeout` are `None` (`src/aio/connection_manager.rs:92-93`). `Service::get_status_list` only falls back to the DB when the cache returns an error, and without a timeout it never does. I paused the Redis container in a local test and `get()` was still blocked after 10s.

Fix: use `ConnectionManager::new_with_config` with `set_response_timeout` and `set_connection_timeout` (configurable, default around 100-250ms). Add a test that pauses Redis and checks that `get()` returns an error within the timeout.

This also makes Redis a startup dependency: if it's down, `build_state` blocks right here. For an optional cache, I'd rather the server start and log a warning.

Cargo.toml
```diff
@@ -79,6 +81,7 @@ tower_governor = "0.8"
 governor = "0.10"
 
 moka = { version = "0.12", features = ["future"] }
+redis = { version = "0.32", features = ["tokio-comp", "connection-manager"], optional = true }
```
martcpp: rediss://` doesn't work.

Only `tokio-comp` and `connection-manager` are enabled. Without a TLS feature, redis-rs rejects `rediss://` with "can't connect with TLS, the feature is not enabled" (`src/connection.rs:391`). ElastiCache with in-transit encryption, Azure Cache (non-TLS port off by default) and Memorystore with TLS all need it, and those are the platforms our aws/azure/gcp images target.

Fix: add `tokio-rustls-comp`.

src/setup.rs
```diff
@@ -526,7 +529,30 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
         (provider, None)
     };
 
-    let status_list_cache = MokaStatusListCache::new(config.cache.ttl, config.cache.max_capacity);
+    #[cfg(any(feature = "cache-memory", not(feature = "cache-redis")))]
+    let status_list_cache: Arc<dyn crate::domain::ports::StatusListCache> = Arc::new(
+        MokaStatusListCache::new(config.cache.ttl, config.cache.max_capacity),
+    );
+
+    #[cfg(all(feature = "cache-redis", not(feature = "cache-memory")))]
```
martcpp, line 553: CI never compiles this block.

`--all-features` turns on both cache features, so the Moka branch wins. The only no-default-features check in `CI.yml` is `--features memory`. So this block, the only code that actually selects Redis, never gets built in CI. It compiles today (I ran `cargo clippy --no-default-features --features memory,postgres,cache-redis -- -D warnings`), but nothing keeps it that way. Add that command to CI.

There's also no way to run this from what we publish: `deploy.yml` has no image variant with `cache-redis`, and the chart can't select one. Either add a variant or document that users have to build their own image.

Dockerfile
```diff
@@ -50,7 +50,11 @@ RUN --mount=type=bind,source=src,target=src \
         arm64) RUST_TARGET="aarch64-unknown-linux-musl" ;; \
         *) echo "Unsupported architecture: ${TARGETARCH:-unset}" && exit 1 ;; \
     esac; \
-    cargo auditable build --locked --release --target=${RUST_TARGET} --features "${FEATURES}"; \
+    if printf '%s' ",${FEATURES}," | grep -q ',cache-redis,'; then \
```
martcpp: backend selection fails silently.

`cargo build --features postgres,cache-redis` produces a Moka binary, because `cache-memory` is a default feature and wins. Nothing says so, and `APP_CACHE__REDIS_URL` is just ignored. Operators will think their replicas share invalidation when each one actually has its own cache.

This grep works around that, but it's fragile:
- `FEATURES="postgres, cache-redis"` (with a space) doesn't match, so you silently get a Moka image.
- The hardcoded `memory` duplicates the Cargo defaults, so any default feature added later is silently missing from Redis images.

Fix:
- Log the active backend at startup.
- Fail at startup if `cache.redis_url` is set but the binary wasn't built with Redis.
- Replace the grep with something explicit: either let `cache-redis` win when both features are on (check the `setup.rs` tests if you go that way), or add a dedicated `CACHE_BACKEND` build arg.

src/outbound/cache.rs
```diff
@@ -100,6 +100,181 @@ impl StatusListCache for MokaStatusListCache {
...
+impl StatusListCache for RedisStatusListCache {
+    async fn get(&self, list_id: &str) -> Result<Option<StatusListRecord>, StatusListError> {
+        use redis::AsyncCommands;
+
+        let key = self.key(list_id);
+        let mut connection = self.connection.clone();
+        let cached: Option<String> = connection.get(key).await.map_err(redis_error)?;
+        let metrics = cache_metrics();
+        if let Some(value) = cached {
+            self.touch(&mut connection, list_id).await?;
```
martcpp: a hit is counted before the value is decoded.

`touch` runs and `hits` goes up, and only then can `serde_json::from_str` fail. The service falls back to the DB, but the metric records a hit and the bad entry stays until its TTL runs out.

This will happen on the first deploy that changes the JSON shape of `StatusListRecord`. Old and new pods share the same keys, can't read each other's entries, and keep overwriting each other for the whole rollout.

Fix: decode first. If decoding fails, delete the key, count a miss, and return `Ok(None)`. Also put a schema version in the key (e.g. `...:v1:<list_id>`).

src/outbound/cache.rs
```diff
@@ -100,6 +100,181 @@ impl StatusListCache for MokaStatusListCache {
...
+                .await
+                .map_err(redis_error)?;
+            metrics
+                .misses
+                .add(1, &[KeyValue::new("cache", "status_list")]);
+            Ok(None)
+        }
+    }
+
+    async fn put(&self, record: StatusListRecord) -> Result<(), StatusListError> {
```
martcpp, line 306: 1. Replica A misses the cache and reads the old record from the DB.
2. Replica B commits a PATCH and invalidates the key.
3. A writes the old record back.

Every replica now serves the pre-PATCH statuses for up to `cache.ttl` (300s by default), so a revoked credential still looks valid for 5 minutes. This race existed with Moka too, but with Redis it hits every replica instead of one.

`updated_at` already only goes up (`next_updated_at` in `service.rs`), so this is fixable. Only write when the incoming `updated_at` is newer than the stored one (a small Lua script). On invalidate, store a short-lived marker with the committed `updated_at` instead of a plain `DEL`.

src/outbound/cache.rs
```diff
@@ -100,6 +100,181 @@ impl StatusListCache for MokaStatusListCache {
...
+            key_prefix,
+            lru_key,
+        })
+    }
+
+    fn key(&self, list_id: &str) -> String {
+        format!("{}{}", self.key_prefix, list_id)
+    }
+
+    fn recency_score() -> f64 {
```
martcpp: - Every read is also a write (`ZADD` on a hit, `ZREM` on a miss). Reads take 2 round trips; puts take 2-4.
- The pipelines aren't atomic. `SETEX` can succeed while `ZADD` fails, and that key then escapes eviction.
- `get` can re-add an id right after another replica's `invalidate`. The leftover id counts toward capacity.
- In `evict_if_needed`, `ZCARD` and `ZRANGE` are separate calls. If another replica evicts in between, this call removes newer entries, possibly the one it just wrote.
- Scores come from each pod's own clock, so a pod whose clock runs slow always has its entries evicted first.

Redis already handles this with a TTL plus `maxmemory-policy volatile-lru`. If we really need a per-entry cap, do get+touch and put+evict in Lua using `redis.call('TIME')`.

src/outbound/cache.rs
```diff
@@ -100,6 +100,181 @@ impl StatusListCache for MokaStatusListCache {
...
+            .await
+            .map_err(redis_error)?;
+        Ok(())
+    }
+}
+
+#[cfg(feature = "cache-redis")]
+#[async_trait]
+impl StatusListCache for RedisStatusListCache {
+    async fn get(&self, list_id: &str) -> Result<Option<StatusListRecord>, StatusListError> {
```
martcpp, line 265: with `ttl = 0`, `put` returns early but `get` still sends `GET` + `ZREM` on every request. The docs say 0 disables the cache, so skip Redis entirely in that case.

src/outbound/cache.rs
```diff
@@ -100,6 +100,181 @@ impl StatusListCache for MokaStatusListCache {
...
+        ttl_secs: u64,
+        max_capacity: u64,
+        key_prefix: impl Into<String>,
+    ) -> Result<Self, StatusListError> {
+        let client = redis::Client::open(redis_url).map_err(redis_error)?;
+        let connection = redis::aio::ConnectionManager::new(client)
+            .await
+            .map_err(redis_error)?;
+        let key_prefix = key_prefix.into();
+        let lru_key = format!("{key_prefix}__lru");
```
martcpp: the sorted set lives at `{prefix}__lru`, and records live at `{prefix}{list_id}`. Publish only accepts UUIDs, but GET doesn't validate the id. So `GET /status-lists/__lru` runs `GET` against the sorted set, gets `WRONGTYPE`, and logs a warning every time. Give them separate namespaces, e.g. `{prefix}meta:lru` and `{prefix}rec:{id}`.

docker-compose.yml
```diff
@@ -108,6 +124,8 @@ services:
         condition: service_started
       localstack:
         condition: service_healthy
+      redis:
```
martcpp: the default compose build is `postgres,aws`, which gives a Moka binary. Compose still starts Redis every time, waits for it to be healthy, and passes a `REDIS_URL` the app ignores. That makes it look like Redis is in use when it isn't. Put Redis behind a compose `profile`. Also drop the `6379:6379` host port: the app reaches Redis over the compose network, and Redis has no password here.

deploy/helm/chart/values.yaml
```diff
@@ -104,8 +104,15 @@ statuslist:
     # USERNAME, or NAME are omitted, deployment.yaml derives them from the active
     # backend's native values below (postgres/mysql).
     # Status-list record cache TTL in seconds. "0" disables this cache.
+    # Cache backend is selected at image build time. cache-memory is the default,
+    # and wins when all features are enabled. Docker builds automatically disable
+    # Cargo defaults when the image FEATURES build arg contains cache-redis.
+    # For cache-redis images, provide APP_CACHE__REDIS_URL via env or a
```
martcpp: the only way I found to pass the Redis URL is plain `statuslist.env`, which leaves the password in clear text in the values file and the Deployment. `_FILE` support only exists for `APP_DATABASE__PASSWORD_FILE`. Add `APP_CACHE__REDIS_URL_FILE` or a `secretKeyRef` option.

src/config.rs
```diff
@@ -1126,7 +1131,9 @@ fn base_builder() -> Result<ConfigBuilder<DefaultState>, ConfigError> {
         .set_default("azure_keyvault.vault_url", Option::<String>::None)?
         .set_default("azure_keyvault.secrets_cache_ttl", 300)?
         .set_default("cache.ttl", 5 * 60)?
-        .set_default("cache.max_capacity", 100)?
+        .set_default("cache.max_capacity", 1000)?
```
martcpp, line 1177: this raises the built-in default from 100 to 1000. Moka deployments that don't set it will now keep 10x as many records, and status lists can be large. Fine if that's intended, but put it in the changelog.

src/outbound/cache.rs
```diff
@@ -189,3 +364,98 @@ mod tests {
         }
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "cache-redis")]
+mod redis_tests {
```
martcpp, line 480: tests:
- No test runs two cache instances on the same prefix and checks that an invalidate from one clears the other. That's the whole point of this feature.
- Nothing covers a Redis outage, a corrupt entry, or `max_capacity = 0`.
- No `Service`-level test runs against Redis.
- These tests need Docker and Docker Hub access, and there's no opt-in gate like `postgres-tests`, so `cargo test --all-features` fails without them (I got a 401 locally). Add a `redis-tests` feature and document `TEST_REDIS_URL`.

src/test_utils.rs
```diff
@@ -120,6 +124,31 @@ impl crate::domain::ports::CertificateProvider for TestCertProvider {
     }
 }
 
+#[derive(Default)]
```
martcpp, line 125: why replace Moka with a hand-written `HashMap` cache here? Integration tests no longer exercise the real in-memory adapter. If there's a reason, leave a comment saying so; otherwise put Moka back. Also, the `HashMap as StdHashMap` alias on line 27 only exists to get around the `acme`-gated import. Just import `HashMap` once.

martcpp (changes requested):

Ngha-Boris:
@martcpp please check again

deploy/helm/chart/values.yaml
```diff
@@ -45,7 +45,8 @@ statuslist:
     # | -azure           | postgres,azure | PostgreSQL database with Azure Key Vault and Azure DNS support | Azure AKS native deployments |
     # | -vault           | postgres,vault | PostgreSQL database with HashiCorp Vault / OpenBao token material storage | Vault-backed on-prem or multi-cloud deployments |
     # | -fscert          | postgres | PostgreSQL database with filesystem-mounted token signing keys and certificates | Constrained / air-gapped environments using external secret injection |
-    # Tag format: <version>-<suffix> (e.g., 1.2.0-aws, 1.2.0-gcp)
+    # | -fscert-redis    | postgres,cache-redis | PostgreSQL database with filesystem-mounted token signing keys and Redis cache | HA / multi-replica cache invalidation |
```
martcpp: This variant can't actually be selected. `values.schema.json` still has `"variant": { "enum": ["fscert", "aws", "gcp", "azure", "vault"] }`, so `statuslist.image.variant: fscert-redis` fails schema validation and the install is rejected. The only way to get this image is to pin `statuslist.image.tag` by hand, which contradicts this table.

Fix: add `fscert-redis` to the enum. The regex on `deployment.yaml:89` already handles the suffix.

deploy/helm/chart/values.yaml
```diff
@@ -104,8 +105,16 @@ statuslist:
     # USERNAME, or NAME are omitted, deployment.yaml derives them from the active
     # backend's native values below (postgres/mysql).
     # Status-list record cache TTL in seconds. "0" disables this cache.
+    # Cache backend is selected at image build time. cache-memory is the default,
+    # and wins when all features are enabled. Docker builds select Redis with
+    # CACHE_BACKEND=redis. For cache-redis images, provide APP_CACHE__REDIS_URL
+    # via statuslist.secretEnv or APP_CACHE__REDIS_URL_FILE via secretMounts.
```
martcpp: NetworkPolicy blocks Redis, and the failure is silent. `templates/network-policy.yaml:31-64` only allows egress to the DB port, 443, 53 and OTLP. `networkPolicy.egressInternal` only adds peers to the **database port** rule, so there's no way to open 6379 (or 6380 for TLS).

With `networkPolicy.enabled=true`, an `fscert-redis` pod can't reach Redis. The connect fails at startup, `setup.rs:592` logs a warning, and the pod comes up healthy with the cache switched off. So the HA setup this PR is for looks fine and quietly has no shared cache.

Fix: add a cache egress rule (a `cacheEgress` value, same shape as `databaseEgress`, plus a configurable port), and document it next to these env vars.

src/setup.rs
```diff
@@ -526,7 +531,76 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+                    Ok(cache) => {
+                        tracing::info!(
+                            cache.backend = "redis",
+                            cache.ttl_secs = config.cache.ttl,
+                            cache.max_capacity = config.cache.max_capacity,
+                            "status-list cache backend selected"
+                        );
+                        Arc::new(cache)
+                    }
+                    Err(error) => {
```
martcpp: Once the cache is disabled here, it stays disabled for the life of the process.

`ConnectionManager` reconnects on its own, but only after it has been constructed. If this `new()` fails, we install `DisabledStatusListCache` and never try Redis again. A 30-second Redis blip during a rolling restart leaves every pod running uncached until someone restarts them, and the only signal is one warning line at startup.

Starting up instead of crashing is the right call. What's missing is a way back:
- retry in the background and swap the cache in once Redis answers, or
- export the backend as a metric (or a `backend` label on the existing hit/miss counters) so `cache disabled` is alertable instead of something you infer from a low hit ratio.

src/outbound/cache.rs
```diff
@@ -100,6 +120,271 @@ impl StatusListCache for MokaStatusListCache {
...
+
+                local now = redis.call('TIME')
+                local score = tonumber(now[1]) * 1000000 + tonumber(now[2])
+                redis.call('ZADD', KEYS[2], score, ARGV[4])
+
+                local overflow = redis.call('ZCARD', KEYS[2]) - max_capacity
+                if overflow > 0 then
+                    local victims = redis.call('ZRANGE', KEYS[2], 0, overflow - 1)
+                    for _, victim in ipairs(victims) do
+                        redis.call('DEL', ARGV[1] .. victim)
```
martcpp: Marker keys aren't counted against `max_capacity`. Each entry is really two keys (`rec:v1:<id>` plus `meta:v1:updated:<id>`), and eviction here deletes only the record. Leaving the marker behind is correct, since it has to outlive the record to reject stale fills, but it means `max_capacity: 1000` can hold up to ~2000 keys, and marker count is bounded by distinct ids seen within TTL rather than by capacity. Worth a comment saying that's deliberate, and worth mentioning in the capacity docs so nobody sizes Redis off `max_capacity` alone.

src/outbound/cache.rs
```diff
@@ -100,6 +120,271 @@ impl StatusListCache for MokaStatusListCache {
...
+        if self.ttl_secs == 0 || self.max_capacity == 0 {
+            return Ok(());
+        }
+
+        let key = self.key(&record.list_id);
+        let marker_key = self.marker_key(&record.list_id);
+        let value = serde_json::to_string(&record).map_err(cache_error)?;
+        let mut connection = self.connection.clone();
+        let list_id = record.list_id.clone();
+        let _: () = redis::cmd("EVAL")
```
martcpp: This sends the whole script body on every `put`, and the touch and corrupt-delete paths do the same. Load the scripts once and use `EVALSHA` (`redis::Script` handles the `NOSCRIPT` reload for you). Small win per call, but `put` is on the read-fill path for every cache miss.

src/outbound/cache.rs
```diff
@@ -100,6 +120,271 @@ impl StatusListCache for MokaStatusListCache {
...
+    async fn get(&self, list_id: &str) -> Result<Option<StatusListRecord>, StatusListError> {
+        use redis::AsyncCommands;
+
+        if self.ttl_secs == 0 || self.max_capacity == 0 {
+            return Ok(None);
+        }
+
+        let key = self.key(list_id);
+        let mut connection = self.connection.clone();
+        let cached: Option<String> = connection.get(key).await.map_err(redis_error)?;
```
martcpp: A hit still costs 2 round trips, `GET` here and then the touch script. Folding both into one script (`GET`, and `ZADD` only when it hits) would halve the latency on the hot path and drop the extra write.

While you're here: the hit/miss counters still have no `backend` label and there's no counter for Redis errors or timeouts. With the timeouts now in place, a Redis that times out on every call looks identical to a cold cache on the dashboards.

src/outbound/cache.rs
```diff
@@ -189,3 +474,304 @@ mod tests {
...
+        tokio::time::sleep(Duration::from_millis(2)).await;
+        cache.put(record("c")).await.expect("put c");
+
+        assert!(cache.get("a").await.expect("get a").is_some());
+        assert_eq!(cache.get("b").await.expect("b was evicted"), None);
+        assert!(cache.get("c").await.expect("get c").is_some());
+    }
+
+    #[tokio::test]
+    async fn redis_cache_shared_prefix_invalidation_clears_other_instance() {
```
martcpp, line 482: The marker logic is the most important thing in this PR and nothing tests it. This test proves invalidation crosses instances, but not that a stale read-fill is rejected.

I checked it by hand and it does work: with the marker at `updated_at=5`, a `put` carrying `updated_at=1` is dropped, and a `put` at `5` is stored. Please add exactly that as a test, so a future change to the  it can't silently bring back the 5-minute stale-revocation window.

martcpp (changes requested):

Ngha-Boris:
Hello @martcpp please recheck

deploy/helm/chart/templates/network-policy.yaml
```diff
@@ -43,6 +44,13 @@ spec:
         {{- with .Values.statuslist.networkPolicy.egressInternal }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
+    {{- if not (empty $cacheEgress) }}
```
martcpp: This fixes the mechanism but leaves the failure reachable by default, and `cacheEgress` defaults to `[]`.

I rendered it: `--set statuslist.image.variant=fscert-redis --set statuslist.networkPolicy.enabled=true` succeeds and emits a NetworkPolicy with egress ports 8000, 5432, 443, 53, 53, 8081 and no 6379. The pod then can't reach Redis, startup logs a warning, the pod goes healthy, and the cache is off.

The chart already fails hard for the same shape of mistake two lines up: mysql with an empty `databaseEgress`. Do the same here, e.g. fail when `networkPolicy.enabled` and `image.variant` ends in `-redis` and `cacheEgress` is empty. Otherwise the default path for the new image is the exact silent misconfiguration this value was added to prevent.

src/outbound/cache.rs
```diff
@@ -11,6 +13,86 @@ use crate::domain::{
 
 const HIT_METRIC: &str = "status_list_cache_hits";
 const MISS_METRIC: &str = "status_list_cache_misses";
+#[cfg(feature = "cache-redis")]
+const ERROR_METRIC: &str = "status_list_cache_errors";
```
martcpp: Nothing consumes this metric. `status_list_cache_errors` appears nowhere under `deploy/` - no recording rule, no alert, no runbook line, unlike the hit/miss SLIs which have both in `prometheusrule.yaml`.

As it stands, "Redis was unreachable at startup so this pod is running with no cache" is still something you find out by noticing a low hit ratio. Add an alert on `status_list_cache_errors_total{operation="startup"}` (or at least a line in the cache-hit-ratio runbook pointing at it), otherwise the counter only helps after someone already suspects the problem.

deploy/helm/chart/values.schema.json
```diff
@@ -197,7 +211,14 @@
             "enabled": { "type": "boolean" },
             "ingress": { "type": "array" },
             "egressInternal": { "type": "array" },
-            "databaseEgress": { "type": "array" }
+            "databaseEgress": { "type": "array" },
+            "cachePort": {
+              "type": ["integer", "string"],
```
martcpp: This union type makes the range check skippable. In JSON Schema, `pattern` only applies to strings and `minimum`/`maximum` only to numbers, so a quoted value is checked for digits but not for range.

Rendered against this chart: `--set-string statuslist.networkPolicy.cachePort=99999` passes validation and emits `port: 99999`, which the API server rejects. The integer form `--set ...cachePort=99999` is correctly refused.

Either drop the string form, or use a pattern that encodes the range so both forms agree.

deploy/helm/chart/templates/network-policy.yaml
martcpp, line 61: Requesting a guard here before merge. `cacheEgress` defaults to `[]`, so the default path for the new image is broken: `--set statuslist.image.variant=fscert-redis --set statuslist.networkPolicy.enabled=true` renders successfully with egress ports 8000, 5432, 443, 53, 53, 8081 and no 6379. The pod can't reach Redis, startup logs a warning, the pod reports healthy, and the cache is off, so an HA deployment looks correct while every replica runs uncached.

The chart already fails hard for the identical mistake two lines up (mysql with empty `databaseEgress`). Please do the same: fail when `networkPolicy.enabled` is true, `image.variant` ends in `-redis`, and `cacheEgress` is empty. This is the one item I'd hold the PR for. It's a few lines of template, and the failure mode is silent.

src/outbound/cache.rs
martcpp, line 15: This counter isn't wired to anything. `status_list_cache_errors` appears nowhere under `deploy/`, no recording rule, no alert, no runbook line, while the hit/miss SLIs have both in `prometheusrule.yaml`.

The result is that "this pod started with no cache" is still only discoverable by noticing a low hit ratio, which is the gap the counter was added to close. Please add an alert on `status_list_cache_errors_total{operation="startup"}`, or at minimum a line in the cache-hit-ratio runbook pointing at it.

deploy/helm/chart/values.schema.json
martcpp: The union type lets the range check be bypassed. In JSON Schema `pattern` applies only to strings and `minimum`/`maximum` only to numbers, so a quoted value is checked for digits but never for range.

Verified against this chart: `--set-string statuslist.networkPolicy.cachePort=99999` passes validation and renders `port: 99999`, which the API server then rejects. The integer form is correctly refused. Please either drop the string form or use a pattern that encodes the 1-65535 range so both forms agree.

martcpp (changes requested):

Ngha-Boris:
Hello @martcpp please can you recheck'

deploy/helm/chart/templates/network-policy.yaml
Hermann-Core: Since no additional suffix or image variant (such as `fscert-redis`) should be created for Redis in either CI or deploy workflows, Redis should compose as other backends without special treatment or dedicated image variants. This `-redis` variant guard and dedicated variant logic should be removed.

src/outbound/cache.rs
Hermann-Core: With the simplification to standard Redis commands and native key expiration via TTL, cache backend operations should log failures at `warn` level and degrade gracefully to persistent storage without blocking the request, avoiding the need for custom startup error counters or dedicated alerts.

deploy/helm/chart/values.schema.json
Hermann-Core: Because Redis will compose like other backends without special image variant suffixes or custom cache network policy branches, `cachePort` should be removed from `values.schema.json`.

src/domain/ports/cache.rs
```diff
@@ -0,0 +1,28 @@
+//! Status-list cache port.
+
+use async_trait::async_trait;
+
+use crate::domain::models::status_list::{StatusListError, StatusListRecord};
+
+/// Pluggable cache interface for recently used status-list records.
+#[async_trait]
+pub trait Cache: Send + Sync + 'static {
```
Hermann-Core: The cache port should not have a dedicated module (`src/domain/ports/cache.rs`). It should be defined directly in `src/domain/ports.rs` alongside all other secondary ports (`StatusListRepo`, `CredentialRepo`, etc.).

Additionally, the cache port contract must contain only cache-level methods (such as generic key-value `get`, `put`/`set`, and `delete`/`remove`), not operations like `invalidate_after_update`. Domain concerns (such as serializing and deserializing records, key namespacing, and TTL resets) belong inside the adapter implementation (`src/outbound/cache.rs`).

Cargo.toml
```diff
@@ -4,8 +4,11 @@ version = "1.0.1"
 edition = "2024"
 
 [features]
-default = ["memory"]
+default = ["memory", "cache-memory"]
 memory = []
+cache-memory = []
+cache-redis = ["dep:redis"]
```
Hermann-Core: The cache backend should be selectable through features like other backends in the workspace (`postgres`, `sqlite`, `mysql`). Please rename `cache-redis` to `redis`, and keep `default = ["memory"]` without introducing artificial `cache-memory` or `cache-redis` feature gates.

All backends must compile cleanly together under `--all-features` without mutually exclusive compile-time conflicts.

.github/workflows/CI.yml
```diff
@@ -221,6 +221,9 @@ jobs:
       - name: Memory-only Build
         run: cargo check --no-default-features --features memory
 
+      - name: Redis cache backend build
```
Hermann-Core: The CI workflow should remain untouched and should not contain a dedicated configuration, step, or matrix variant for Redis. Redis must be treated like all other backends and validated via `--all-features`.

Please remove the custom `Redis cache backend build` step, the `fscert-redis` matrix variant, and the conditional build logic.

.github/workflows/deploy.yml
```diff
@@ -106,19 +106,28 @@ jobs:
...
             description: "PostgreSQL database with Azure Key Vault and Azure DNS support"
           - suffix: vault
             features: "postgres,vault"
+            cache_backend: "memory"
             description: "PostgreSQL database with HashiCorp Vault / OpenBao token material storage"
           - suffix: fscert
             features: "postgres"
+            cache_backend: "memory"
             description: "PostgreSQL database with filesystem-mounted token signing keys and certificates"
+          - suffix: fscert-redis
```
Hermann-Core: No additional suffix or image variant (such as `fscert-redis`) should be created in CI or deploy workflows. Redis should compose as other backends without special treatment or dedicated image tags, and `deploy.yml`, `scheduled-image-scan.yml`, and `Dockerfile` should be kept untouched without `CACHE_BACKEND` build arguments.

Dockerfile
```diff
@@ -50,7 +52,11 @@ RUN --mount=type=bind,source=src,target=src \
         arm64) RUST_TARGET="aarch64-unknown-linux-musl" ;; \
         *) echo "Unsupported architecture: ${TARGETARCH:-unset}" && exit 1 ;; \
     esac; \
-    cargo auditable build --locked --release --target=${RUST_TARGET} --features "${FEATURES}"; \
+    case "${CACHE_BACKEND}" in \
+        memory) cargo auditable build --locked --release --target=${RUST_TARGET} --features "${FEATURES},cache-memory" ;; \
```
Hermann-Core: Please revert `ARG CACHE_BACKEND`, `ARG REQUIRED_FEATURES`, and the conditional build switch. Redis should not require a dedicated binary build path; it should compile cleanly alongside other features via `--all-features` and be selected at runtime via configuration.

src/config.rs
```diff
@@ -979,6 +979,53 @@ pub struct CacheConfig {
     /// Setting this to 0 disables caching entirely.
     pub ttl: u64,
     pub max_capacity: u64,
+    /// Redis connection URL used by binaries built with the `cache-redis` feature.
+    #[serde(default)]
+    pub redis_url: Option<SecretString>,
```
Hermann-Core: Redis configuration should follow the existing structure and philosophy established by `DatabaseConfig`:
1. Use split fields (`host`, `port`, `username`, `password`, `password_file`, `database`, `tls`) rather than accepting a raw `redis_url` or `redis_url_file`.
2. Construct and resolve the connection URL internally (for example via `resolved_url()` / `load_resolved_url()`), setting default ports (6379 for plain, 6380 for TLS) and appropriate schemes (`redis://` / `rediss://`).
3. The Redis key prefix should be chosen and hardcoded internally (e.g. `status-list-server:status-list:`). It should not be user-configurable to avoid cluttering configuration values.
4. Avoid exposing internal timeout knobs like `redis_response_timeout_ms` and `redis_connection_timeout_ms`. Safe defaults (e.g. response timeouts and reconnect retries) should be configured internally in `ConnectionManagerConfig`.
5. Keep existing configuration defaults intact (e.g. `cache.max_capacity` should remain 100).

src/outbound/cache.rs
```diff
@@ -11,6 +13,86 @@ use crate::domain::{
...
+#[cfg(feature = "cache-redis")]
+const CACHE_SCHEMA_VERSION: &str = "v1";
+
```
Hermann-Core, line 24: The custom Lua scripts and sorted set (`ZADD`, `ZCARD`, `ZRANGE`, `ZREM`) LRU emulation should be removed. Redis provides native TTL key expiration.

The cache backend should be initialized with the configured cache TTL so that all inserted status list entries are automatically expired and removed by Redis upon expiration (using standard `SET ... EX <ttl>`, `GET`, and `DEL` commands via `redis::aio::ConnectionManager`). Additionally, any failure to interact with the cache backend must log at `warn` level and degrade gracefully to persistent storage without blocking the request.

src/setup.rs
```diff
@@ -526,7 +531,77 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
         (provider, None)
     };
 
-    let status_list_cache = MokaStatusListCache::new(config.cache.ttl, config.cache.max_capacity);
+    #[cfg(any(feature = "cache-memory", not(feature = "cache-redis")))]
```
Hermann-Core: Cache backend selection in `setup.rs` should follow the same pattern used for database backends:
1. Define a `CacheBackend` enum on `CacheConfig` (e.g. `Memory`, `Redis`) selectable at runtime.
2. Select the backend via a match expression gated by `#[cfg(feature = "redis")]`. If `CacheBackend::Redis` is configured but the `redis` feature was not compiled in, return a descriptive error.
3. This allows `--all-features` to compile both Moka and Redis backends cleanly into a single binary, selecting the desired backend based on configuration rather than mutually exclusive compile-time features.

deploy/helm/chart/templates/network-policy.yaml
```diff
@@ -2,9 +2,14 @@
 {{- $otel := index .Values "opentelemetry-collector" | default dict -}}
 {{- $backend := include "status-list-server-chart.dbBackend" . -}}
 {{- $databaseEgress := .Values.statuslist.networkPolicy.databaseEgress | default list -}}
+{{- $cacheEgress := .Values.statuslist.networkPolicy.cacheEgress | default list -}}
+{{- $imageVariant := .Values.statuslist.image.variant | default "fscert" -}}
 {{- if and (eq $backend "mysql") (empty $databaseEgress) }}
 {{- fail "statuslist.networkPolicy.databaseEgress must be set when NetworkPolicy is enabled with APP_DATABASE__BACKEND=mysql because this chart does not deploy the database" }}
 {{- end }}
+{{- if and (hasSuffix "-redis" $imageVariant) (empty $cacheEgress) }}
+{{- fail "statuslist.networkPolicy.cacheEgress must be set when NetworkPolicy is enabled with a Redis cache image variant" }}
```
Hermann-Core: Since the `fscert-redis` variant suffix and dedicated Redis image variants should not exist, the conditional checks for image variants ending in `-redis` and dedicated `cachePort`/`cacheEgress` network policy branches should be removed. Redis should compose naturally as a standard backend.

Hermann-Core (changes requested):
@Ngha-Boris Please review the architectural and configuration requirements below:

1. Cache Port Definition and Contract: Keep the cache port inside `src/domain/ports.rs` (remove `src/domain/ports/cache.rs`). The port trait must only declare cache-level methods (generic `get`, `set`/`put`, `delete`/`remove`) without leaking mutation timestamps (`invalidate_after_update`). Domain concerns (serialization, TTL handling, and key formatting) belong inside the adapter layer.
2. Feature Selection: Name the cargo feature `redis` (matching database backends `postgres`, `mysql`, `sqlite`) rather than `cache-redis`/`cache-memory`. Ensure `--all-features` compiles all backends cleanly without mutually exclusive preprocessor blocks.
3. CI, Deploy Workflows, and Image Variants: Revert modifications to `.github/workflows/CI.yml`, `.github/workflows/deploy.yml`, `.github/workflows/scheduled-image-scan.yml`, and `Dockerfile`. Redis must compose like any other backend without dedicated CI jobs, build arguments, or image variant suffixes (no `fscert-redis`).
4. Redis Configuration: Use split fields (`host`, `port`, `username`, `password`, `password_file`, `database`, `tls`) and assemble the connection URL internally as done for `DatabaseConfig`. Hardcode the Redis key prefix internally rather than exposing it as a configuration knob, and use safe internal defaults for timeouts and retries instead of crowding the configuration surface.
5. Native TTL Expiration and Error Resilience: Replace the custom Lua scripts and sorted-set LRU emulation with standard Redis operations (`GET`, `SET ... EX <ttl>`, `DEL`). Redis natively expires keys when initialized with the cache TTL. A failed cache operation must never block the request path; it should log at `warn` level and fall back to persistent storage.

Ngha-Boris:
Hello @Hermann-Core and @martcpp please recheck this

deploy/helm/chart/values.yaml
```diff
@@ -103,9 +103,19 @@ statuslist:
...
+    # Redis requires APP_CACHE__HOST plus optional split fields below; the app
+    # assembles redis:// or rediss:// internally.
+    APP_CACHE__BACKEND: "memory"
     APP_CACHE__TTL: "300"
-    APP_CACHE__MAX_CAPACITY: "1000"
+    # Maximum status-list records retained by the in-memory backend. Redis uses
+    # native key expiration based on APP_CACHE__TTL.
+    APP_CACHE__MAX_CAPACITY: "100"
+    APP_CACHE__HOST: ""
+    APP_CACHE__PORT: ""
```
martcpp: APP_CACHE__PORT: "" and APP_CACHE__DATABASE: "" are rendered into the pod. config 0.15 only turns a missing value into None, so "" becomes Some("") and "".parse::<u16>() fails. Config::load() errors on a default install even with APP_CACHE__BACKEND=memory. Either remove these keys from values.yaml, or add .ignore_empty(true) to the Environment source at src/config.rs:1146 (check the other env vars if you do). Please add a test that loads config with the chart's default env.

src/setup.rs
```diff
@@ -526,7 +536,67 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+                        }
+                        Err(error) => {
+                            record_redis_cache_error("startup");
+                            tracing::warn!(
+                                error = ?error,
+                                cache.backend = "disabled",
+                                cache.target = %config.cache.redacted_redis_target(),
+                                "Redis status-list cache unavailable at startup; continuing with cache disabled"
+                            );
+                            Arc::new(DisabledStatusListCache)
```
martcpp: If Redis can't be reached at startup, this pod gets DisabledStatusListCache for its whole life, and that cache's invalidate() does nothing. A PATCH handled by this pod commits to the DB but never deletes the Redis key, so every healthy replica keeps serving the old status list for up to cache.ttl. A revoked credential can stay valid for up to 5 minutes. Please don't swap in a disabled cache for Redis: keep the ConnectionManager reconnecting lazily, or retry the connection in the background, and make sure invalidation still reaches Redis.

src/outbound/cache.rs
```diff
@@ -100,6 +161,121 @@ impl StatusListCache for MokaStatusListCache {
...
+        if self.ttl_secs == 0 {
+            return Ok(());
+        }
+
+        let key = self.key(&record.list_id);
+        let value = serde_json::to_string(&record).map_err(cache_error)?;
+        let mut connection = self.connection.clone();
+        let _: () = connection
+            .set_ex(key, value, self.ttl_secs)
```
martcpp: SET EX overwrites unconditionally. Sequence: a reader misses and reads version N from the DB; a writer commits N+1 and runs DEL; the reader then stores N here. Every replica then serves the stale entry for a full TTL. Commit 0ae1b09 fenced this with an updated_at marker and a Lua put; that was removed in 3f9508c. Please restore a compare-on-updated_at put, or have invalidate write a short-lived marker that blocks older puts.

src/outbound/cache.rs
```diff
@@ -189,3 +365,214 @@ mod tests {
...
+    use super::*;
+    use crate::domain::models::credential::Issuer;
+    use crate::domain::models::status_list::StatusList;
+    use testcontainers_modules::testcontainers::{
+        ContainerAsync, GenericImage,
+        core::{IntoContainerPort, WaitFor},
+        runners::AsyncRunner,
+    };
+
+    fn record_at(list_id: &str, updated_at: i64) -> StatusListRecord {
```
martcpp: record_at(.., updated_at) looks left over from the removed fencing. Remove it, or use it in a race test once fencing is back.

Cargo.toml
```diff
@@ -4,8 +4,11 @@ version = "1.0.1"
 edition = "2024"
 
 [features]
-default = ["memory"]
+default = ["memory", "cache-memory"]
 memory = []
+cache-memory = []
```
martcpp: cache-memory = [] doesn't gate any code: moka is always compiled in, and there is no compile_error! when both backends are enabled. The backend is actually picked at runtime by APP_CACHE__BACKEND. That is a reasonable design, but it doesn't meet the issue's compile-time requirement. Please either update the issue or the acceptance criteria to say "runtime selection" and remove this feature, or gate MokaStatusListCache behind it and add a compile_error! when both backends are on. The second option also means changing CI's --all-features runs.

src/outbound/cache.rs
```diff
@@ -100,6 +161,121 @@ impl StatusListCache for MokaStatusListCache {
...
+
+        let key = self.key(list_id);
+        let mut connection = self.connection.clone();
+        let cached: Option<String> = connection
+            .get(&key)
+            .await
+            .map_err(|error| redis_operation_error("get", error))?;
+        let metrics = cache_metrics();
+        if let Some(value) = cached {
+            match serde_json::from_str(&value) {
```
martcpp: Whatever is decoded here goes straight into JWT/CWT signing. Please at least check record.list_id == list_id and treat a mismatch as a corrupt entry. Ideally also HMAC the payload, and require TLS plus auth for Redis when APP_ENV=production. Right now, write access to Redis is enough to get the server to sign any status list.

martcpp (changes requested):
Egress only allows the DB port, 443 and 53. egressInternal peers are attached to the DB-port rule, so adding them never opens 6379/6380. With networkPolicy.enabled=true and APP_CACHE__BACKEND=redis, Redis is blocked and each pod quietly falls back to no cache (see the setup.rs:587 comment). Please add a cache egress rule, with its port taken from APP_CACHE__PORT or the default for APP_CACHE__TLS. Make the chart fail when the backend is redis and no cache egress peer is set. also update runbook line 53

Ngha-Boris:

.github/workflows/CI.yml
Hermann-Core: Please remove this dedicated `Redis cache backend build` step and revert the `,cache-redis` additions in the matrix below. Workflows should not be modified for Redis; the backend compiles cleanly alongside other features under `--all-features`.

.github/workflows/deploy.yml
```diff
@@ -105,19 +105,19 @@ jobs:
       matrix:
         variant:
           - suffix: aws
-            features: "postgres,aws"
+            features: "postgres,aws,cache-redis"
```
Hermann-Core: Please revert the `,cache-redis` additions across the matrix variants. Deploy workflows should remain untouched; Redis should compose as other backends without dedicated image matrix changes.

Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN --mount=type=cache,target=/root/.cargo/registry,id=registry-cache-${TARGETPL
     env -u CARGO_BUILD_TARGET cargo install --locked --root /usr/local cargo-auditable@${CARGO_AUDITABLE_VERSION}; \
     env -u CARGO_BUILD_TARGET cargo install --locked --root /usr/local rust-audit-info@${RUST_AUDIT_INFO_VERSION}
 
-ARG FEATURES="postgres,aws"
+ARG FEATURES="postgres,aws,cache-redis"
```
Hermann-Core: Please revert `ARG FEATURES="postgres,aws,cache-redis"` back to `ARG FEATURES="postgres,aws"`. The Dockerfile should remain untouched.

Cargo.toml
```diff
@@ -6,6 +6,8 @@ edition = "2024"
 [features]
 default = ["memory"]
 memory = []
+cache-redis = ["dep:redis"]
```
Hermann-Core: Please rename the cargo feature from `cache-redis` to `redis` (and `redis-tests = ["redis"]`), matching the convention used for other backends in the workspace (`postgres`, `mysql`, `sqlite`).

src/domain/ports/cache.rs
Hermann-Core: The cache port should not reside in a separate module (`src/domain/ports/cache.rs`). Please define `StatusListCache` directly in `src/domain/ports.rs` alongside other secondary ports. Additionally, keep the trait focused strictly on cache operations (`get`, `put`, `invalidate`) without leaking domain mutation timestamps (`invalidate_after_update`).

src/setup.rs
```diff
@@ -526,7 +536,60 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+        CacheBackend::Redis => {
+            if config.cache.ttl == 0 {
+                tracing::info!(
+                    cache.backend = "disabled",
+                    cache.reason = "ttl_zero",
+                    "status-list cache disabled"
+                );
+                Arc::new(DisabledStatusListCache)
+            } else {
+                #[cfg(feature = "cache-redis")]
```
Hermann-Core: Update the feature gating here and above to `#[cfg(feature = "redis")]` / `#[cfg(not(feature = "redis"))]` once the cargo feature is renamed to `redis`.

src/outbound/cache.rs
```diff
@@ -100,6 +193,204 @@ impl StatusListCache for MokaStatusListCache {
     }
 }
 
+#[cfg(feature = "cache-redis")]
+#[derive(Clone)]
+pub struct RedisStatusListCache {
+    client: redis::Client,
+    connection: Arc<tokio::sync::Mutex<Option<redis::aio::ConnectionManager>>>,
```
Hermann-Core: Avoid wrapping `ConnectionManager` in `Arc<tokio::sync::Mutex<Option<...>>>` and locking it on every operation. `redis::aio::ConnectionManager` already multiplexes requests concurrently and reconnects automatically. It can be held directly on `RedisStatusListCache` so concurrent tasks clone the manager handle without serializing on a mutex.

src/outbound/cache.rs
```diff
@@ -11,6 +13,46 @@ use crate::domain::{
...
+const CACHE_SCHEMA_VERSION: &str = "v1";
+#[cfg(feature = "cache-redis")]
+const REDIS_KEY_PREFIX: &str = "status-list-server:status-list:";
+#[cfg(feature = "cache-redis")]
+const REDIS_RESPONSE_TIMEOUT: Duration = Duration::from_millis(250);
+#[cfg(feature = "cache-redis")]
+const REDIS_CONNECTION_TIMEOUT: Duration = Duration::from_millis(250);
+
+#[cfg(feature = "cache-redis")]
+static REDIS_PUT_SCRIPT: LazyLock<redis::Script> = LazyLock::new(|| {
```
Hermann-Core: Please replace the custom Lua scripts (`REDIS_PUT_SCRIPT`, `REDIS_INVALIDATE_SCRIPT`) and marker keys with standard Redis commands (`SET ... EX <ttl>`, `GET`, `DEL`). Redis natively expires keys after the configured TTL.

Hermann-Core (changes requested):
@Ngha-Boris Thanks for the updates! The configuration improvements in `src/config.rs` (split fields, internal URL assembly, hardcoded key prefix, safe internal timeouts) and the clean removal of the `fscert-redis` Helm variant are great steps forward.

Please address the remaining items below to align with the workspace architecture:

1. Revert CI, deploy, and Dockerfile changes: Keep `.github/workflows/CI.yml`, `.github/workflows/deploy.yml`, and `Dockerfile` untouched. The Redis backend should not have dedicated CI build steps or appended feature flags in release image variants; it will be built and verified naturally under `--all-features`.
2. Feature naming: Rename the cargo feature from `cache-redis` to `redis` in `Cargo.toml` and conditional compilation gates to match other backend features (`postgres`, `mysql`, `sqlite`).
3. Cache port location and contract: Remove `src/domain/ports/cache.rs` and keep the cache port definition directly in `src/domain/ports.rs` alongside other secondary ports. Keep the trait strictly focused on cache operations (`get`, `put`, `invalidate`) without leaking domain mutation timestamps (`invalidate_after_update`).
4. Avoid connection mutex serialization: In `src/outbound/cache.rs`, avoid locking an `Arc<tokio::sync::Mutex<Option<ConnectionManager>>>` on every cache call. `redis::aio::ConnectionManager` is designed to be cloned and multiplexes concurrent requests internally.
5. Standard Redis operations and native TTL: Replace the custom Lua scripts and marker keys with standard Redis commands (`GET`, `SET ... EX <ttl>`, `DEL`), relying on Redis native TTL expiration.

Ngha-Boris:

src/outbound/cache.rs
Hermann-Core: This remains unresolved in the latest revision: `put` still uses unconditional `SET EX`. A reader can miss and load version N, a PATCH can commit N+1 and `DEL` the key, then the reader can write N back for the full TTL. That can make a revocation stale on every replica. Restore version-aware write fencing (or an invalidation marker that rejects older `updated_at` values) and add the race regression test.

docker-compose.yml
Hermann-Core: The Redis service is now correctly profile-gated, but the app build still defaults to `postgres,aws,cache-redis`. The feature was renamed to `redis`, so `docker compose build` now fails with `the package status-list-server does not contain this feature: cache-redis`. Change this to `redis`; update the same stale feature name in README.md and .env.template as well.

src/setup.rs
Hermann-Core: The current revision avoids installing `DisabledStatusListCache` by returning an error from `build_state`, but that makes Redis a hard startup dependency. It also contradicts the new startup alert and runbook, which say a pod continues with the cache disabled; no server starts to expose that counter. Keep cache failures on the bounded, warn-and-fallback path, for example with a lazily reconnecting adapter, while ensuring PATCH invalidation still reaches Redis after recovery.

Hermann-Core:
@Ngha-Boris I reopened the relevant resolved threads and left review comments for you to check.

martcpp:
@Ngha-Boris some conflict here too

Ngha-Boris:
Thank you @martcpp and @Hermann-Core for your reviews so far, can you please re-check

deploy/helm/chart/templates/deployment.yaml
martcpp: Add the HA end-to-end PATCH test and the setup wiring tests; use one shared test container.

src/outbound/cache.rs
```diff
@@ -100,6 +212,187 @@ impl StatusListCache for MokaStatusListCache {
...
+    }
+
+    async fn invalidate(&self, list_id: &str) -> Result<(), StatusListError> {
+        if self.ttl_secs == 0 {
+            return Ok(());
+        }
+
+        let mut connection = self.connection("invalidate").await?;
+        let marker = crate::domain::service::current_unix_timestamp();
+        let _: i32 = REDIS_INVALIDATE_SCRIPT
```
martcpp: The marker uses the writer's wall clock, but updated_at is an OCC version (max(now, prev + 1)), not wall time. So a fill can still get through after an invalidation. Example, all within second 100: a reader loads updated_at = 100, a PATCH commits 101 and writes marker = 100, then the reader's put passes (100 < 100 is false). The pre-PATCH list is then cached for the full TTL on every replica. Bursts of updates and clock skew between replicas cause the same thing.

Fix: make the marker the committed updated_at by passing it through invalidate. invalidate_after_commit already receives &existing, so it's available. OCC guarantees the old value is strictly lower, so no clock is involved. Please also add a test that uses record_at(now). The current test uses updated_at = 1, so it can't catch this.

.env.template
```diff
@@ -160,10 +160,24 @@ AWS_SECRET_ACCESS_KEY=test
 # APP_VAULT__TIMEOUT_SECS=30
 
 # Cache configuration
-# TTL for the status list record cache. 
+# Runtime backend: memory or redis. Redis requires a binary built with the
```
martcpp: his isn't true. The Dockerfile has ARG FEATURES="postgres,aws", and none of the deploy.yml image variants include redis. Only docker-compose adds it. If a Helm user sets APP_CACHE__BACKEND=redis, the pod will crash with "binary was not built with the redis feature".

Since the backend is chosen at runtime, please add redis to the release image features. Otherwise, remove this claim and the Helm Redis docs.

deploy/helm/chart/templates/prometheusrule.yaml
```diff
@@ -207,6 +207,19 @@ spec:
             description: 'Current cache hit ratio: {{ `{{ $value | humanizePercentage }}` }}'
             runbook_url: "https://github.com/adorsys/status-list-server/blob/develop/deploy/observability/runbooks/cache-hit-ratio.md"
             dashboard_url: {{ .Values.alerting.dashboardUrl | default "http://localhost:3000/d/status-list-slo" | quote }}
+        - alert: RedisCacheStartupError
+          expr: sum by (namespace)(status_list_cache_errors_total{otel_scope_name="status-list-server",operation="startup"}) > 0
```
martcpp: Two problems with this alert:

It fires on a raw counter being > 0. After one startup error it stays firing until the pod restarts, even once the reconnect has worked. The summary also says "cache disabled", but the cache reconnects on its own.
A Redis outage while the pods are running isn't covered at all. When get fails, it counts as neither a hit nor a miss, so the hit ratio becomes NaN and CacheHitRatioLow stays quiet.
Suggest replacing it with sum by (namespace)(rate(status_list_cache_errors_total{...}[5m])) > 0 and for: 5m, covering all operations. Also count failed gets as misses. The same change applies to alerting.rules.yml:189.

Cargo.toml
```diff
@@ -6,6 +6,8 @@ edition = "2024"
 [features]
 default = ["memory"]
 memory = []
+redis = ["dep:redis"]
```
martcpp: The issue asks for the backend to be chosen at compile time (cache-memory / cache-redis, with compilation failing if both are enabled). This PR chooses it at runtime with APP_CACHE__BACKEND. Moka is always compiled in, the memory feature is the storage backend rather than a cache, and CI builds with --all-features. The trait and the handler changes were also already on develop.

I think runtime selection is the better design, but the issue checklist says otherwise. Please update the issue to match what was built (and record the decision), or change the code to match the issue.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Bump the Helm chart to `0.5.0` for render-time validation changes.
+- Increase the built-in status-list cache capacity default from 100 to 1000 entries; deployments can override with `APP_CACHE__MAX_CAPACITY`.
```
martcpp: This says the capacity default went from 100 to 1000, but this PR does the reverse. values.yaml and .env.template go from 1000 to 100, so every Helm deploy on the memory backend gets a cache 10× smaller when it upgrades. Please restore 1000, or fix this entry and explain why it went down.

deploy/helm/chart/values.yaml
```diff
@@ -147,9 +147,16 @@ statuslist:
...
+    # Status-list record cache. Backend is selected at runtime: memory or redis.
+    # Redis requires APP_CACHE__HOST plus optional split fields below; the app
+    # assembles redis:// or rediss:// internally.
+    APP_CACHE__BACKEND: "memory"
     # Status-list record cache TTL in seconds. "0" disables this cache.
     APP_CACHE__TTL: "300"
-    APP_CACHE__MAX_CAPACITY: "1000"
+    # Maximum status-list records retained by the in-memory backend. Redis uses
+    # native key expiration based on APP_CACHE__TTL.
+    APP_CACHE__MAX_CAPACITY: "100"
```
martcpp: develop has 1000 here. Was this change intended? See the CHANGELOG comment.

src/outbound/cache.rs
```diff
@@ -100,6 +212,187 @@ impl StatusListCache for MokaStatusListCache {
...
+            redis::aio::ConnectionManager::new_with_config(self.client.clone(), manager_config),
+        )
+        .await
+        .map_err(|_| redis_timeout_error(operation, REDIS_CONNECTION_TIMEOUT))?
+        .map_err(|error| redis_operation_error(operation, error))?;
+        self.connection.store(Some(Arc::new(connection.clone())));
+        Ok(connection)
+    }
+
+    fn key(&self, list_id: &str) -> String {
```
martcpp: Put the list_id in a hash tag ({list_id}) so the record key and the marker key land in the same slot; otherwise the scripts fail with CROSSSLOT on Redis Cluster, and changing the key format later means a migration.

src/outbound/cache.rs
```diff
@@ -11,6 +13,65 @@ use crate::domain::{
...
+        r#"
+        local marker = redis.call('GET', KEYS[2])
+        local updated_at = tonumber(ARGV[2])
+        if marker and updated_at < tonumber(marker) then
+            return 0
+        end
+
+        local current = redis.call('GET', KEYS[1])
+        if current then
+            local decoded = cjson.decode(current)
```
martcpp: This decodes up to ~1 MB of JSON inside Redis just to read updated_at; store it as its own hash field (HSET key v <json> u <ts>) and HGET only u.

src/outbound/cache.rs
```diff
@@ -11,6 +13,65 @@ use crate::domain::{
...
 const HIT_METRIC: &str = "status_list_cache_hits";
 const MISS_METRIC: &str = "status_list_cache_misses";
+#[cfg(feature = "redis")]
+const ERROR_METRIC: &str = "status_list_cache_errors";
+#[cfg(feature = "redis")]
+const CACHE_SCHEMA_VERSION: &str = "v1";
+#[cfg(feature = "redis")]
+const REDIS_KEY_PREFIX: &str = "status-list-server:status-list:";
+#[cfg(feature = "redis")]
+const REDIS_RESPONSE_TIMEOUT: Duration = Duration::from_millis(250);
```
martcpp: Make the timeouts configurable, and add single-flight plus a short circuit breaker to connection() so an outage doesn't add up to 500 ms to every GET.

src/outbound/cache.rs
```diff
@@ -100,6 +212,187 @@ impl StatusListCache for MokaStatusListCache {
...
+
+        let key = self.key(list_id);
+        let mut connection = self.connection("get").await?;
+        let cached: Option<String> = connection
+            .get(&key)
+            .await
+            .map_err(|error| redis_operation_error("get", error))?;
+        let metrics = cache_metrics();
+        if let Some(value) = cached {
+            match serde_json::from_str::<StatusListRecord>(&value) {
```
martcpp: The server signs whatever this returns, so anyone who can write to Redis can un-revoke credentials; HMAC the payload, or document a dedicated ACL user limited to status-list-server:*.

src/config.rs
```diff
@@ -975,10 +992,130 @@ pub struct AzureKeyVaultConfig {
...
+    pub password: Option<SecretString>,
+    /// File containing the Redis password, for secret-mounted deployments.
+    #[serde(default)]
+    pub password_file: Option<PathBuf>,
+    /// Redis database number.
+    #[serde(default)]
+    pub database: Option<u8>,
+    /// Use TLS (`rediss://`) for the Redis connection.
+    #[serde(default)]
+    pub tls: bool,
```
martcpp: Add cache.ca_file for private-CA Redis (production requires TLS) and a configurable key prefix so environments sharing a Redis don't collide.

src/outbound/cache.rs
```diff
@@ -100,6 +212,187 @@ impl StatusListCache for MokaStatusListCache {
...
+    }
+}
+
+#[cfg(feature = "redis")]
+#[async_trait]
+impl StatusListCache for RedisStatusListCache {
+    async fn get(&self, list_id: &str) -> Result<Option<StatusListRecord>, StatusListError> {
+        use redis::AsyncCommands;
+
+        if self.ttl_secs == 0 {
```
martcpp: This can't run because setup already sends ttl=0 to DisabledStatusListCache; remove it and use Disabled for memory with ttl=0 as well.

src/setup.rs
```diff
@@ -526,7 +536,69 @@ async fn build_state_impl(config: &AppConfig) -> EyeResult<BuildStateResult> {
...
+                );
+                Arc::new(DisabledStatusListCache)
+            } else {
+                #[cfg(feature = "redis")]
+                {
+                    let redis_url = config.cache.load_resolved_redis_url().await?;
+                    let cache =
+                        RedisStatusListCache::new(redis_url.expose_secret(), config.cache.ttl)
+                            .map_err(|error| {
+                                record_redis_cache_error("startup");
```
martcpp: new() already counts this failure as connect, so this counts it twice.

src/outbound/cache.rs
```diff
@@ -11,6 +13,65 @@ use crate::domain::{
 
 const HIT_METRIC: &str = "status_list_cache_hits";
 const MISS_METRIC: &str = "status_list_cache_misses";
+#[cfg(feature = "redis")]
+const ERROR_METRIC: &str = "status_list_cache_errors";
+#[cfg(feature = "redis")]
+const CACHE_SCHEMA_VERSION: &str = "v1";
```
martcpp: Add a test that pins the serialized shape of StatusListRecord so a field change forces a version bump.

src/outbound/cache.rs
```diff
@@ -100,6 +212,187 @@ impl StatusListCache for MokaStatusListCache {
...
+impl RedisStatusListCache {
+    pub fn new(redis_url: &str, ttl_secs: u64) -> Result<Self, StatusListError> {
+        let client = redis::Client::open(redis_url)
+            .map_err(|error| redis_operation_error("connect", error))?;
+        Ok(Self {
+            client,
+            connection: Arc::new(ArcSwapOption::from(None)),
+            ttl_secs,
+            record_prefix: format!("{REDIS_KEY_PREFIX}rec:{CACHE_SCHEMA_VERSION}:"),
+            marker_prefix: format!("{REDIS_KEY_PREFIX}meta:{CACHE_SCHEMA_VERSION}:updated:"),
```
martcpp: Take the version out of the marker key (it's just an integer) so invalidations still cover old keys after a schema bump.

martcpp (changes requested):
Document the maxmemory-policy recommendation, the new backend metric label, and statuslist.secretEnv in the CHANGELOG.

deploy/helm/chart/templates/deployment.yaml
Hermann-Core: Reproduced on the current head: `cargo test --no-default-features --features memory,redis-tests --lib outbound::cache -- --nocapture` fails 2/10 tests. `redis_cache_corrupt_entry_is_miss_and_deleted` and `redis_cache_mismatched_list_id_is_miss_and_deleted` time out in `cache.connection("test")` after 250 ms while the per-test containers start concurrently. Please use a shared Redis fixture (or otherwise remove the startup race) so this required coverage is reliable in CI.

src/outbound/cache.rs
Hermann-Core: There is a second fencing gap to cover with the same fix: the marker expires after `min(60, cache.ttl)`. A request which read N before the PATCH but is delayed until after that marker expires can still repopulate N for a full cache TTL. Keep a version fence for at least every possible in-flight fill (preferably a durable per-list version), and add a regression test for the expiry path as well as the same-second OCC case.

CHANGELOG.md
Hermann-Core: This file must not be updated manually, please revert all the changes made here.

docker-compose.yml
```diff
@@ -56,6 +56,20 @@ services:
     networks:
       - status-list-network
 
+  redis:
+    image: redis:8.4-alpine
```
Hermann-Core:
```suggestion
    image: redis:8.10-alpine
```

README.md
```diff
@@ -158,6 +158,19 @@ Verify zero-infrastructure in-memory compilation:
...
+```bash
+cargo clippy --no-default-features --features memory,postgres,redis -- -D warnings
+```
+
+Run Redis cache integration tests with either Docker/testcontainers or an
+existing Redis endpoint:
+
+```bash
+TEST_REDIS_URL=redis://localhost:6379/0 cargo test --no-default-features --features memory,redis-tests outbound::cache
+```
```
Hermann-Core: Please avoid cluttering the README with Redis-specific commands. Redis should be treated consistently with the other backends, with its detailed setup and operational instructions documented in the appropriate location.

The README should remain clean, concise, and focused on providing an overview, with links to more detailed documentation where appropriate.

Cargo.toml
```diff
@@ -79,6 +81,7 @@ tower_governor = "0.8"
 governor = "0.10"
 
 moka = { version = "0.12", features = ["future"] }
+redis = { version = "0.32", features = ["tokio-comp", "tokio-rustls-comp", "connection-manager"], optional = true }
```
Hermann-Core: To keep this consistent with other dependencies that have multiple configuration options, I suggest using the table-style dependency declaration.
```suggestion
[dependencies.redis]
version = "1.7"
features = ["tokio-comp", "tokio-rustls-comp", "connection-manager"]
optional = true
```

Hermann-Core (changes requested):

docker-compose.yml
martcpp: this is out of scope @Hermann-Core

Ngha-Boris:
Hello @martcpp and @Hermann-Core please recheck

src/outbound/cache.rs
Hermann-Core: Required: the durable marker must also be monotonic. `REDIS_INVALIDATE_SCRIPT` unconditionally writes `ARGV[1]`; two committed updates can invalidate out of order (N+2 writes its marker, then delayed N+1 lowers it), allowing a delayed N+1 read-fill to be cached after N+2. Atomically retain `max(existing_marker, committed_updated_at)` and add a reverse-order invalidation regression test.

Cargo.toml
Hermann-Core: Required: the current implementation and ADR use runtime `APP_CACHE__BACKEND`, but this PR description and linked issue #167 still claim compile-time `cache-memory`/`cache-redis` selection, mutual exclusion, and the old Redis URL variables. Please update both before merge so the release and review record matches the shipped configuration contract.

src/outbound/cache.rs
Ngha-Boris: addressed

Cargo.toml
Ngha-Boris: addressed

Ngha-Boris:
Hello @Hermann-Core all your concerns have been addressed please recheck

PR #556, opened 2026-09-16
fix: prevent 304 Not Modified when client JWT has expired

### LINKED ISSUE #482: HTTP cache revalidation: prevent 304 Not Modified when client JWT has expired
## Objective

Ensure `GET /api/v1/status-lists/{list_id}` does not return `304 Not Modified` with an empty body when the client's previously cached Status List Token has expired.

## Context

When a client queries a status list with an `If-None-Match: <etag>` header, the server validates the ETag against the current database version. If the list has not changed, the server returns `304 Not Modified` with an empty body.

However, if the client's cached JWT token has reached its `exp` time, the client receives no body and is left with an expired, unusable token. Per RFC 9110 §8.8.1, a weak validator should change when the cached representation is no longer an acceptable substitute, and Draft-21 §8.2–8.3 requires valid tokens for relying party verification.

## Deliverables

- [x] Update conditional revalidation logic in `src/server/handlers/status_list/get_status_list.rs` and `conditional.rs` to take token expiration into account.
- [x] Factor token expiration window into ETag generation or evaluate token expiration before returning 304.
- [x] When the cached token would be expired, bypass the 304 response and return a fresh `200 OK` with a newly signed Status List Token.
- [x] Add automated tests with simulated time advancement verifying that revalidation with an expired token yields a fresh 200 OK response.

## Acceptance Criteria

- [x] Revalidation of an expired token returns `200 OK` with a newly signed, valid token instead of `304 Not Modified`.
- [x] Revalidation within the token's validity period continues to return `304 Not Modified` efficiently.

## References

- `src/server/handlers/status_list/get_status_list.rs`
- `src/server/handlers/status_list/utils/conditional.rs`
- Arc review: `arc-review/finalReview/application-functional-bugs/README.md` (PoC 4)
- RFC 9110 §8.8.1 & Draft-21 §8.2–8.3

ndefokou:
## Summary

Closes #482. Ensures `GET /api/v1/status-lists/{list_id}` does not return `304 Not Modified` with an empty body when the client's previously cached Status List Token has reached its `exp` time.

When a client revalidates an unchanged list with `If-None-Match`, the server previously validated only the content ETag and returned `304` with no body, leaving the client with an expired, unusable token. Per RFC 9110 §8.8.1, a weak validator must change when the cached representation is no longer an acceptable substitute, and Draft-21 §8.2–8.3 requires relying parties to have a valid token to verify against.

## Changes

- **`src/server/handlers/status_list/utils/conditional.rs`**
  - Added a new `ConditionalResponse::StaleRevalidation` variant signalling that the ETag matches but the cached token has expired.
  - Added `cached_token_expiry(updated_at, token_exp_secs)`, anchoring the cached token's validity window at `updated_at + token_exp_secs` — the same horizon the historical snapshot uses for `iat`.
  - `evaluate_if_none_match` / `evaluate_conditional_request` now take the current time and the token expiry; an ETag match inside the validity window returns `NotModified`, while a match past the window returns `StaleRevalidation`. `If-Modified-Since` handling is unchanged.

- **`src/server/handlers/status_list/get_status_list.rs`**
  - Passes the current timestamp and `token_exp_secs` into the conditional evaluation.
  - Routes `StaleRevalidation` (alongside `Modified`) to the existing fresh-token path, returning a `200 OK` with a newly signed Status List Token instead of a `304`.

- **Tests**
  - `conditional.rs`: unit tests for the expiry gate (expired match → stale, valid match → not modified, expired wildcard, no-match still modified, `cached_token_expiry`, precedence with stale).
  - `get_status_list.rs`:
    - `test_conditional_request_expired_token_returns_fresh_200` — expired-token revalidation returns `200 OK` with a non-empty freshly signed body.
    - `test_conditional_request_if_none_match_within_validity_returns_304` — revalidation within validity still returns `304` (efficient revalidation preserved).

## Acceptance Criteria

- ✅ Revalidation of an expired token returns `200 OK` with a newly signed, valid token instead of `304 Not Modified`.
- ✅ Revalidation within the token's validity period continues to return `304 Not Modified`.

## References

- RFC 9110 §8.8.1
- Draft-21 §8.2–8.3

## Validation

- `cargo build`
- `cargo clippy --lib` (no warnings)
- `cargo fmt --check` (clean)
- `cargo test --lib status_list` — 108 passed, 0 failed

Note: no changes to `etag.rs` (the content validator stays stable so valid revalidation remains efficient), `token.rs`, routes, or configuration.


src/server/handlers/status_list/utils/conditional.rs
```diff
@@ -9,19 +9,27 @@ const IMF_FIXDATE: &[time::format_description::BorrowedFormatItem<'static>] = fo
 pub(crate) enum ConditionalResponse {
     NotModified,
     Modified,
+    StaleRevalidation,
+}
+
+/// Cached-token validity bound for a given record version.
+pub(crate) fn cached_token_expiry(updated_at: i64, token_exp_secs: u64) -> i64 {
```
martcpp: This isn't the cached token's expiry.

Live tokens are signed with `iat = now` at request time â€” `utils/token.rs:136`:

```rust
let iat = OffsetDateTime::now_utc().unix_timestamp();
(iat, iat + state.token_exp_secs as i64)
```

So the client's token expires at `fetch_time + token_exp_secs`, where `fetch_time >= updated_at` and is unbounded above â€” a client can fetch a year after the last write. The ETag can't recover it for you either; it's content-only by design (`utils/etag.rs:4-14`, pinned by `test_generate_etag_updated_at_independence`). The server has no record of when the client fetched, so this is estimating an unknowable quantity from a proxy that isn't correlated with it.

Net effect with the default `token_exp_secs = 900`: for any list not written to in the last 15 minutes, `now >= updated_at + 900` is **always** true, so every conditional GET returns `StaleRevalidation` â†’ full re-signed body. A list at rest is the normal steady state for a revocation list â€” that's the case the 304 path exists for, and it's now permanently off. Acceptance criterion 2 fails in production even though the tests pass (their records are milliseconds old).

Cost per request goes from a ~200-byte header-only 304 to: PEM parse + P-256 sign + gzip over the whole bitstring + `spawn_blocking` hop + full body egress.

Two smaller things on these same lines:

- `token_exp_secs as i64` wraps negative above `i64::MAX`, and `saturating_add` doesn't help because the cast happens first. `i64::try_from(token_exp_secs).unwrap_or(i64::MAX)` if this survives.
- The name asserts the thing that isn't true. If the anchor stays, `record_version_expiry_floor` describes what it actually computes.

src/server/handlers/status_list/utils/conditional.rs
```diff
@@ -68,9 +84,16 @@ pub(crate) fn evaluate_conditional_request(
...
+    now: i64,
+    token_exp_secs: u64,
 ) -> ConditionalResponse {
     if if_none_match.is_some() {
-        return evaluate_if_none_match(if_none_match, current_etag);
+        return evaluate_if_none_match(
+            if_none_match,
+            current_etag,
+            now,
+            cached_token_expiry(updated_at, token_exp_secs),
```
martcpp: The other direction â€” the original bug still reproduces here.

`updated_at` bumps on every write regardless of whether the content changed. `domain/service.rs:180-181` has no "did anything actually change?" guard, and `StatusList::update(vec![])` returns `self.clone()` (`domain/models/status_list.rs:121-123`), so an empty-statuses update is an explicit no-op that still moves the timestamp. The ETag, being content-only, doesn't move with it.

```
t=0     client GETs             updated_at=0    etag=E   token exp=900
t=800   issuer no-op update     updated_at=800  etag=E   (unchanged)
t=1000  client revalidates E â†’ expiry computed as 800+900=1700 â†’ 304
```

304 with an empty body, client left holding a token that died at t=900. That's PoC 4 verbatim. Re-setting an index to the value it already holds does the same thing.

Related: `next_updated_at(prev, now) = now.max(prev + 1)` (`domain/service.rs:306-308`) can push `updated_at` past `now` under write bursts, which widens this window further.

src/server/handlers/status_list/utils/conditional.rs
```diff
@@ -68,9 +84,16 @@ pub(crate) fn evaluate_conditional_request(
...
+        );
     }
     evaluate_if_modified_since(if_modified_since, updated_at)
```
martcpp: The `If-Modified-Since` path never got the freshness gate.

`evaluate_if_modified_since` (lines 56-80) still returns `NotModified` purely on `updated_at <= client_timestamp`. A client revalidating with IMS alone gets a 304 with an expired token 100% of the time, under all conditions â€” the identical bug, same endpoint, same handler, untouched.

Not hypothetical: `test_conditional_request_if_modified_since_returns_304` (`get_status_list.rs:570`) proves it's a supported, exercised flow. The deliverable says "update conditional revalidation logic in `conditional.rs` to take token expiration into account" â€” this is the other half of it.

Worth flagging: the bucketed-ETag approach I suggested in the review body won't fix this for free either, since `Last-Modified` can't carry the bucket. IMS needs its own gate regardless of which direction you go.

src/server/handlers/status_list/get_status_list.rs
```diff
@@ -609,6 +612,115 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_conditional_request_expired_token_returns_fresh_200() {
+        let token_id = uuid::Uuid::new_v4().to_string();
+        let mut app_state = test_app_state(None).await;
+        // Simulate the client's cached token having reached its exp time: with a
+        // zero-length validity window the cached token (updated_at + 0) is already
+        // expired at revalidation time, so the server must bypass the 304.
+        app_state.token_exp_secs = 0;
```
martcpp: This test can't fail.

With `token_exp_secs = 0`, `utils/token.rs:137` signs `exp = iat + 0` â€” **the "fresh" token this test accepts is already expired at issuance.** The assertions are only `status == OK` and `!body.is_empty()`, so a server handing back a stone-dead token passes. Acceptance criterion 1 says "newly signed, **valid** token"; that's unverified. Decode the JWT and assert `claims.exp > now`.

Worse: under `exp_secs = 0` the implementation is indistinguishable from `fn evaluate(..) -> Modified`. This test passes for *any* implementation that returns 200.

It also isn't the "simulated time advancement" the deliverable asked for â€” it's a degenerate config that can't occur meaningfully in production. The test that would have caught the regression I flagged on `conditional.rs:16` is the one that's missing: realistic `token_exp_secs = 900`, record aged past the window, client token still valid â†’ must return 304.

You can't write that today because `now` is read from the global clock inside the handler (line 90), which is presumably why you reached for `exp_secs = 0` in the first place. Injecting a clock into `AppState` is the prerequisite, and it's the thing that unblocks the test the issue actually asked for.

Minor: this also quietly depends on `>=` vs `>` at `conditional.rs:49` â€” with `exp_secs = 0`, `now == expiry` in the same-second case, so tightening that boundary later breaks this test in a way that won't be obvious.

src/server/handlers/status_list/utils/conditional.rs
```diff
@@ -30,13 +38,21 @@ pub(crate) fn evaluate_if_none_match(
...
-            return ConditionalResponse::NotModified;
+            return not_modified_or_stale(now, expiry);
         }
     }
 
     ConditionalResponse::Modified
 }
 
+fn not_modified_or_stale(now: i64, expiry: i64) -> ConditionalResponse {
+    if now >= expiry {
```
martcpp: No skew margin and no minimum-remaining-lifetime floor.

`now >= expiry` hands back a 304 when one second of validity remains. Two problems: the relying party's clock may be ahead of yours, so it rejects a token you just certified as fresh; and even with perfect clocks the client has to turn around and refetch immediately, which is thrash.

The token already carries `ttl = token_ttl_secs` (`utils/token.rs:294`), which means exactly "how long you may cache this before refetching". So the natural rule is 304 only when `expiry - now > token_ttl_secs + skew`, not `expiry - now > 0`.

(The `>=` itself is correct for JWT `exp` semantics per RFC 7519 Â§4.1.4 â€” it's the zero margin I'm flagging, not the operator.)

src/server/handlers/status_list/get_status_list.rs
```diff
@@ -104,7 +107,7 @@ pub async fn get_status_list(
             ],
         )
             .into_response()),
-        ConditionalResponse::Modified => {
+        ConditionalResponse::Modified | ConditionalResponse::StaleRevalidation => {
```
martcpp: `StaleRevalidation` is a distinction without a difference â€” this is its only consumer and it's merged straight into `Modified`. Nothing else in the crate inspects the variant.

Either collapse it into `Modified` and drop the variant, or earn it: emit `conditional_revalidation_total{outcome="not_modified"|"stale"|"modified"}` plus a `tracing::debug!` on this arm.

As it stands there is no way to see from production that the 304 path has gone quiet â€” which is precisely the signal you'd need to catch the regression I flagged on `conditional.rs:16`. Given this change can silently multiply the endpoint's CPU and egress cost, a counter isn't optional.

martcpp (changes requested):
Requesting changes. The bug is real and you fixed it in the right place, but the freshness model is anchored to the wrong timestamp, and that breaks things in both directions.

The core problem in one line: **live tokens are signed with `iat = now` at request time** (`utils/token.rs:136`), not `iat = updated_at`:

```rust
let iat = OffsetDateTime::now_utc().unix_timestamp();
(iat, iat + state.token_exp_secs as i64)
```

So the client's cached token expires at `fetch_time + token_exp_secs`, and `updated_at + token_exp_secs` is simply not that number. The commit message says the anchor was borrowed from the snapshot path â€” that's correct *there* (`domain/service.rs:311` sets `iat = record.updated_at`), but it doesn't carry over to live tokens.

Consequences are inline below. Short version: for any list not written to in 15 minutes (the default `exp`), every conditional GET now returns a full re-signed body instead of a 304 â€” so acceptance criterion 2 fails in production. And a no-op write still lets an expired token get a 304, so the original bug survives.

I ran `cargo test --lib conditional` â€” 31 pass, 0 fail. The suite is green because nothing in it ages a record past the window or checks that the returned token is actually valid.

## Suggested fix

Take the issue's first option â€” put the validity window *in* the ETag instead of guessing it:

```rust
let window = token_exp_secs as i64;
let bucket = now - now.rem_euclid(window);
// sign with (bucket, bucket + 2 * window)
// etag = W/"<content-hash>-<bucket>"
```

An ETag match then **proves** the token was issued in the current window, so you know its `exp` rather than estimating it. Inside the window â†’ 304, token provably valid. Window rolls over â†’ ETag differs â†’ the existing `Modified` path returns 200. No new enum variant, no `now`/`token_exp_secs` threaded through two layers, no clock in the evaluator. It's smaller than what's on the branch now.

Bonus: every client in a window gets a byte-identical body, so CDNs can actually collapse requests. Today `iat` is per-request, so every single response is unique and nothing downstream can dedupe.

## Separate issues â€” the fix above will NOT address these

These are real but out of scope for this PR; they want their own tickets.

1. **OpenAPI documents none of this.** `docs/openapi.yaml:330-445` lists 200/404/406/429/500/503 for `getStatusList` â€” no 304, no `If-None-Match`/`If-Modified-Since` parameters, no `ETag`/`Last-Modified`/`Cache-Control` response headers. A client integrating from the spec cannot discover conditional GET exists. Line 428 also claims `Vary` is "Always `Accept-Encoding`" while the handler sends `Accept, Accept-Encoding`.

2. **No config validation.** `config.rs:1130-1131` doesn't enforce `token_exp_secs > 0` or `token_ttl_secs <= token_exp_secs`. Set TTL 3600 / EXP 900 and a shared cache serves an expired token for 45 minutes without ever revalidating â€” same bug class as this issue, one layer out. And `token_exp_secs = 0` (which this PR's test leans on) makes every token expired at birth and empties the `iat <= time < exp` snapshot window.

3. **`Cache-Control: max-age={ttl}, immutable`** (`get_status_list.rs:269`) contradicts this PR's own premise that the representation changes. `immutable` per RFC 8246 promises it won't. The bucketing scheme above would make it honest; otherwise drop it.

4. **Keypair reparse per request.** `utils/token.rs:143` parses the PKCS#8 PEM on every token build. That was tolerable when most revalidations stopped at 304 â€” it isn't once this path is hot.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

src/server/handlers/status_list/utils/conditional.rs
```diff
@@ -9,6 +9,45 @@ const IMF_FIXDATE: &[time::format_description::BorrowedFormatItem<'static>] = fo
 pub(crate) enum ConditionalResponse {
     NotModified,
     Modified,
+    /// A 304 must not be answered even though the representation is unchanged:
```
martcpp: make it one line

src/server/handlers/status_list/utils/conditional.rs
martcpp, line 1: cut down the comment of this file while still making your point

src/server/handlers/status_list/utils/etag.rs
```diff
@@ -1,13 +1,24 @@
 use crate::domain::models::status_list::{StatusListRecord, StatusListSnapshot};
 use sha2::{Digest, Sha256};
 
-pub(crate) fn generate_etag(record: &StatusListRecord) -> String {
+/// Generate the weak ETag for the *live* status list representation.
```
martcpp: also this comment too

martcpp (changes requested):
add the following test case thanks
- the CWT path (only JWT is tested)
- the gzip path in the expired case
- the E == 0 and ttl ≥ E configs
- revalidation exactly at we − ttl and at we, on the handler path
- headers on the ExpiredToken 200 (ETag, Cache-Control, Vary)
- content changed and window rolled over
- metric labels

src/server/handlers/status_list/get_status_list.rs
```diff
@@ -33,6 +67,23 @@ pub async fn get_status_list(
...
+    let now = OffsetDateTime::now_utc().unix_timestamp();
+    get_status_list_at(State(state), list_id, query_result, headers, now).await
+}
+
+/// Request-time implementation of [`get_status_list`] with an explicit `now`.
+///
+/// `now` is injected (rather than read from the clock inside) so tests can
+/// advance the clock deterministically and pin the token-expiry revalidation
+/// behaviour.
+#[tracing::instrument(skip_all, fields(list_id = %list_id), err(Debug))]
```
martcpp: remove the duplicate instrument so errors aren't logged twice.

martcpp (changes requested):

src/server/handlers/status_list/utils/conditional.rs
```diff
@@ -67,12 +121,33 @@ pub(crate) fn evaluate_conditional_request(
...
-        return evaluate_if_none_match(if_none_match, current_etag);
+        // `current_etag` is keyed to the current token validity window (see
+        // etag::generate_etag). A match means the client fetched within this
+        // window, so in the worst case its token still has `window_end - now` of
+        // validity left. Require that to exceed `token_ttl_secs` so a 304 never
+        // certifies a token that is (or will imminently become) unusable under
+        // clock skew; otherwise serve a freshly signed token.
+        let matches =
+            evaluate_if_none_match(if_none_match, current_etag) == ConditionalResponse::NotModified;
+        let runway = window_end.saturating_sub(now);
```
martcpp: Switch to window length W = E − ttl and delete the runway branch. It's simpler, and the steady-state 304 rate goes from 33% to 50% with the defaults.

src/server/handlers/status_list/get_status_list.rs
```diff
@@ -263,7 +367,10 @@ fn client_accepts_gzip(headers: &HeaderMap) -> bool {
 }
 
 fn build_cache_control(token_ttl_secs: u64) -> String {
-    format!("max-age={}, immutable", token_ttl_secs)
+    // No `immutable`: live tokens expire and their validators rotate, so a
```
martcpp: wrong explanation of immutable

docs/openapi.yaml
martcpp, line 1: make the OpenAPI text match the actual rules (E − ttl, the dead zone, claim names instead of config names).

martcpp (changes requested):

ndefokou:
@martcpp The ticket is back for you to review

Hermann-Core (approved):
@ndefokou Excellent work addressing the review feedback and finalizing the token revalidation mechanism.

### Key Highlights
- **ETag Validity Window:** Keying the live ETag to the validity window `W = exp - ttl` (`token_window(now, validity).0`) ensures that any 304 response leaves at least `token_ttl_secs` of usable lifetime on the client token. Once the window rolls over, the ETag naturally changes, serving a fresh `200 OK` signed token instead of an empty `304`.
- **Degenerate Config Safety:** The guard against `ttl_secs >= exp_secs` and `exp_secs == 0` reliably forces `ExpiredToken` to prevent certifying 304s when usable lifetime cannot be guaranteed.
- **If-Modified-Since Freshness Gate:** Incorporating the `updated_at + exp_secs` horizon with the `ttl_secs` runway check prevents stale 304 responses when only IMS is supplied.
- **Cache-Control Alignment:** Removing `immutable` from live responses (`Cache-Control: max-age=<ttl>`) accurately reflects representation rotation, while preserving `immutable` for historical snapshot tokens.
- **Observability:** Added `conditional_revalidation_total` with `not_modified`, `modified`, and `expired_token` outcome labels to monitor revalidation behavior in production.
- **Test Coverage:** All requested edge cases are thoroughly covered: CWT format, gzip handling on expiry re-sign, dead-zone boundary revalidations (`window_end - ttl` vs `window_end`), degenerate configs, response headers on re-sign, content-change combined with window rollover, and metric outcome labels.
- **OpenAPI Documentation:** Cleanly documents `If-None-Match`, `If-Modified-Since`, and the `304 Not Modified` response contract.

martcpp (approved):

PR #557, opened 2026-09-17
feat: enforce 8-bit capacity for application-specific status values and update related validations

### LINKED ISSUE #481: Enforce Draft-21 bit-width {1, 2, 4, 8} and ZLIB-compressed empty lists
## Objective

Enforce OAuth Status List Draft-21 compliance by restricting bit width strictly to `{1, 2, 4, 8}` and ensuring empty status lists serialize to a valid ZLIB-compressed stream.

## Context

During protocol review against [OAuth Status List Draft-21](https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/):
1. **Bit width overflow:** When a status value >= 256 is supplied, `determine_bits` in `src/domain/models/status_list.rs` calculates `(256+1).next_power_of_two().trailing_zeros() == 9`, causing the token to emit `bits=9`. However, Draft-21 §4.2 explicitly mandates that `bits` MUST be one of `{1, 2, 4, 8}`.
2. **Empty status list representation:** Publishing a list with `statuses: []` produces `lst: ""` (empty string). An empty string is not a valid RFC 1950 ZLIB stream, causing third-party relying party decoders to fail with decompression errors.

## Deliverables

- [x] In `src/domain/models/status_list.rs`, update `determine_bits` to strictly limit bit width to 1, 2, 4, or 8.
- [x] Reject status values exceeding 8-bit capacity (values > 255 cannot fit in an 8-bit status list under Draft-21) with a clear validation error (`400 Bad Request`).
- [x] In `StatusList::create`, when `status_updates` is empty, ensure the status array is encoded through `encode_compressed` so `lst` contains a valid base64url-encoded ZLIB byte stream instead of an empty string.
- [x] Add tests validating that standard zlib decoders successfully inflate empty status list tokens.
- [x] Add unit tests verifying `bits` is never emitted as any value outside `{1, 2, 4, 8}`.

## Acceptance Criteria

- [x] `bits` in generated status list tokens is strictly one of 1, 2, 4, or 8.
- [x] Empty status list tokens contain a valid, base64url-encoded ZLIB stream that decodes cleanly without error.
- [x] Conforms to OAuth Status List Draft-21 §4.2 and §4.3.

## References

- `src/domain/models/status_list.rs` (lines 103–107, 180–210)
- `src/utils/bits_validation.rs`
- Arc review: `arc-review/finalReview/application-functional-bugs/README.md` (PoC 2)

Ngha-Boris:
**Summary**

Enforces OAuth Status List Draft-21 bit-width compliance and fixes empty status list serialization.

**Changes**

- Restricted generated `bits` values to Draft-21-supported widths: `1`, `2`, `4`, and `8`.
- Rejects status values greater than `255` with a clear validation error, mapping to `400 Bad Request`.
- Encodes empty status lists through ZLIB compression so `lst` is a valid base64url-encoded RFC 1950 stream instead of an empty string.
- Updated request deserialization to accept application-specific values within `3..=255` and reject values above 8-bit capacity.
- Added tests covering:
  - Empty list ZLIB inflation with a standard decoder.
  - Rejection of values `>255`.
  - Generated bit widths always staying within `{1, 2, 4, 8}`.
  - Unsupported stored bit widths rejected as corrupt state.

**Validation**

- `cargo fmt`
- `cargo test`

All tests pass: `192` unit tests, `31` integration tests, and `2` doctests.


src/server/handlers/status_list/utils/request.rs
```diff
@@ -27,10 +27,10 @@ impl<'de> Deserialize<'de> for Status {
             0 => Status::VALID,
             1 => Status::INVALID,
             2 => Status::SUSPENDED,
-            n if n >= 256 => Status::ApplicationSpecific(n),
+            n @ 3..=255 => Status::ApplicationSpecific(n),
```
Christiantyemele: Spec-compliance (Important): Draft-21 §7.1 reserves only 0x03 and 0x0C-0x0F (values 3 and 12-15) as application-specific; 0x04-0x0B and 0x10-0xFF are 'reserved for future registration', not application-specific. Accepting the full 3..=255 range here lets an issuer emit status lists whose status-type values are non-conformant, which a spec-conformant relying party cannot interpret. The error message 'exceeds 8-bit capacity / max 255' also conflates the 8-bit encoding capacity with the status-type registry. Suggest restricting application-specific values to {3} u {12..=15} and erroring on 4-11 and 16-255 as reserved, or at minimum tightening this arm and status_value() to match the draft's reserved ranges.

src/domain/models/status_list.rs
```diff
@@ -170,10 +172,19 @@ fn status_value(status: &Status) -> Result<u32, StatusListError> {
         Status::Valid => Ok(0),
         Status::Invalid => Ok(1),
         Status::Suspended => Ok(2),
-        Status::ApplicationSpecific(value) if *value >= 256 => Ok(*value),
-        Status::ApplicationSpecific(_) => Err(StatusListError::InvalidStatusList(
-            "ApplicationSpecific value must be >= 256".to_string(),
-        )),
+        Status::ApplicationSpecific(value) if *value <= 255 => Ok(*value),
```
Christiantyemele: Spec-compliance (Important), mirrors request.rs:30: `value <= 255` accepts 4-255 as application-specific, but Draft-21 only reserves 0x03 and 0x0C-0x0F (3, 12-15) for application-specific use; 4-11 and 16-255 are reserved for future registration. The 'exceeds 8-bit capacity' error conflates encoding capacity with the status-type registry. Recommend narrowing to {3} u {12..=15} (or documenting the broader policy) so produced lists stay conformant.

src/domain/models/status_list.rs
```diff
@@ -123,6 +123,7 @@ impl StatusList {
         }
 
         let old_bits = self.bits as usize;
+        validate_bits(old_bits)?;
```
Christiantyemele: Question (backward compatibility): the previous code allowed bit widths of 9/13/16 etc. (via next_power_of_two) for application-specific values >= 256. Any such list already persisted in storage now fails `validate_bits(old_bits)` on update and surfaces as CorruptStoredList -> HTTP 500, so existing high-bit-width rows can no longer be updated. If such data can exist from earlier releases, a migration/rejection story is needed; if it provably cannot, a comment noting that assumption would help.

Christiantyemele (changes requested):
Overall this is a solid, well-tested change: enforcing Draft-21 bit widths (1/2/4/8) via validate_bits, rejecting status values over 255, and fixing empty-list serialization so `lst` is a valid base64url-zlib stream are all correct improvements. Tests pass (192 unit, 31 integration, 2 doctests) and clippy is clean. Findings: 1 Important (spec compliance on the accepted application-specific value range), 1 Question (backward compatibility for previously stored >8-bit lists). Once the application-specific range is aligned with the draft's reserved 0x03 / 0x0C-0x0F values (or the divergence is documented), the rest is ready to approve.

Ngha-Boris:
Hello @Christiantyemele please recheck

src/domain/models/status_list.rs
```diff
@@ -358,12 +379,7 @@ fn decode_status_array(array: &[u8], bits: usize) -> Result<Vec<Status>, StatusL
             0 => Status::Valid,
             1 => Status::Invalid,
             2 => Status::Suspended,
-            value if value >= 256 => Status::ApplicationSpecific(value),
-            _ => {
-                return Err(StatusListError::CorruptStoredList(
-                    "Invalid status value in existing list".to_string(),
-                ));
-            }
+            value => Status::ApplicationSpecific(value),
```
Christiantyemele: Suggestion (symmetry): encode paths now enforce the application-specific registry via is_application_specific_status_value() (status_value, Deserialize, Serialize), but decode_status_array still accepts any value as Status::ApplicationSpecific (e.g. reserved 4-11 / 16-255 decode as app-specific). This is benign today because the only re-encode path (widening updates) routes through status_value() and rejects reserved values, and serving uses the pre-encoded lst. Consider rejecting non-registry values here as CorruptStoredList for symmetry with the encode path (or add a comment noting legacy-read tolerance). Optional, not blocking.

Christiantyemele (commented):
Good overral except for the below suggestion

src/domain/models/status_list.rs
Ngha-Boris: Good point @Christiantyemele. I’m leaning toward keeping `decode_status_array` tolerant for now because it is reading already-stored compressed state, and making it strict would turn any legacy row containing reserved values into `CorruptStoredList` immediately. The encode paths now reject reserved values, so new or mutated lists cannot emit them.

Christiantyemele (dismissed):

src/domain/models/status_list.rs
```diff
@@ -103,7 +103,7 @@ impl StatusList {
         if status_updates.is_empty() {
             return Ok(Self {
                 bits: 1,
-                lst: String::new(),
+                lst: encode_compressed(&[])?,
```
Hermann-Core, line 106: Empty status list serialization now correctly produces a valid RFC 1950 ZLIB stream (`encode_compressed(&[])?`), which ensures relying party decoders conformant with Draft-21 Section 8.3 Step 5 can inflate it without unexpected EOF errors.

However, lists created by older releases were stored in the database with `lst: ""`. When an issuer calls `update` on a legacy empty list, `decode_compressed(&self.lst)` feeds an empty slice to `ZlibDecoder`, which fails with `UnexpectedEof` and returns `StatusListError::CorruptStoredList`.

To maintain backward compatibility with previously stored empty lists, consider handling empty inputs in `decode_compressed`:
```rust
fn decode_compressed(encoded: &str) -> Result<Vec<u8>, StatusListError> {
    if encoded.is_empty() {
        return Ok(Vec::new());
    }
    let bytes = base64url::decode(encoded).map_err(...)?;
    // ...
```
This allows existing empty lists to be updated smoothly and upgraded to the new compressed format upon write.

src/domain/models/status_list.rs
```diff
@@ -170,10 +175,30 @@ fn status_value(status: &Status) -> Result<u32, StatusListError> {
         Status::Valid => Ok(0),
         Status::Invalid => Ok(1),
         Status::Suspended => Ok(2),
-        Status::ApplicationSpecific(value) if *value >= 256 => Ok(*value),
-        Status::ApplicationSpecific(_) => Err(StatusListError::InvalidStatusList(
-            "ApplicationSpecific value must be >= 256".to_string(),
-        )),
+        Status::ApplicationSpecific(value) if is_application_specific_status_value(*value) => {
```
Hermann-Core, line 225: The enforcement of Draft-21 Section 4.1 and Section 7.1 is well implemented:
- Bit widths are strictly confined to `{1, 2, 4, 8}`.
- Application-specific values correctly align with `0x03` and `0x0C..=0x0F` (`3` and `12..=15`).
- Reserved ranges (`4..=11` and `16..=255`) and values exceeding 8-bit capacity (`> 255`) are rejected with clear, distinct validation messages.

Note that because `determine_bits` derives `bits` from the highest status value in the update payload, and all currently supported status values are `<= 15`, newly created status lists will automatically select `bits: 1`, `bits: 2`, or `bits: 4`. Generating a list with `bits: 8` will only occur if an existing list already has 8-bit width or once future IANA status values (`>= 16`) or an explicit bit-width request parameter are introduced.

src/domain/models/status_list.rs
Hermann-Core: In Draft-21 Section 7.1, values `4..=11` and `16..=255` are designated as 'reserved for future registration' rather than application-specific (`3` and `12..=15`).

While mapping any unknown status to `Status::ApplicationSpecific(value)` was intended as read tolerance for stored lists, notice what happens during a widening update (`new_bits > old_bits`): `update` re-encodes decoded entries through `create_with_bits` -> `apply_updates` -> `status_value(&update.status)`. At that point, `status_value` rejects the reserved value with `StatusListError::InvalidStatusList`, returning a `400 Bad Request` to the client that initiated the update.

If a stored list contains reserved status values, it represents corrupt stored state rather than an invalid client request. Mapping reserved values to `CorruptStoredList` in `decode_status_array` ensures accurate error classification and 500 status rather than blaming the caller.

Hermann-Core (commented):
@Ngha-Boris Please review the spec-compliance and backward-compatibility feedback below:

1. Draft-21 Alignment: The enforcement of allowed bit widths `{1, 2, 4, 8}` (Draft-21 Section 4.1) and status value ranges (Section 7.1) is accurate. The distinction between application-specific values (`3`, `12..=15`), reserved values (`4..=11`, `16..=255`), and values exceeding 8-bit capacity (`>255`) correctly follows the specification.
2. Empty List Inflation: Serializing empty status lists via ZLIB compression (`encode_compressed(&[])`) produces a valid RFC 1950 stream and resolves inflation failures in standard decoders (Section 8.3 Step 5).
3. Legacy Empty List Updates: Pre-existing empty lists stored with `lst = ""` will fail with `UnexpectedEof` when updated because `decode_compressed` expects a ZLIB stream. Adding an early return `if encoded.is_empty() { return Ok(Vec::new()); }` in `decode_compressed` ensures legacy empty lists can be updated and transparently upgraded.
4. Error Classification in `decode_status_array`: During a widening update, any stored entries with reserved values (`4..=11`, `16..=255`) currently fail in `apply_updates` as `InvalidStatusList` (HTTP 400). Flagging them as `CorruptStoredList` in `decode_status_array` avoids misattributing stored corruption to the caller's update request.

martcpp (changes requested):
- there is no plan for stored lists with widths ∉ {1,2,4,8}. They produce a generic 500 on every update, which blocks revocation.
- criterion 1 and 2: no guard at token emission. Legacy rows still produce bits = 9 and lst: "".
- requests with values > 255 get 422 in plain text over HTTP, not the required 400 ApiError.

- a breaking API change with no breaking-change marker in the commits, an incorrect OpenAPI StatusValue schema, and a stale compliance matrix.

src/domain/models/status_list.rs
```diff
@@ -170,10 +175,30 @@ fn status_value(status: &Status) -> Result<u32, StatusListError> {
...
+                "status value {value} is reserved for future registration; application-specific status values are 3 and 12 through 15"
+            )))
+        }
+        Status::ApplicationSpecific(value) => Err(StatusListError::InvalidStatusList(format!(
+            "status value {value} exceeds 8-bit capacity; maximum supported status value is 255"
+        ))),
+    }
+}
+
+pub(crate) fn is_application_specific_status_value(value: u32) -> bool {
```
martcpp, line 234: **Suggestion (non-blocking): make illegal status values unrepresentable**

The status-value rule currently lives in four places: `status_value()`, this helper, and the request `Serialize`/`Deserialize` in `request.rs`, each with its own error text (that's how the "maximum supported status value is 255" message was copied into three places). `determine_bits` also needs a dead `16..=255 => 8` arm plus `unreachable!()`, which only holds because `status_value` ran first.

A validated newtype makes the compiler enforce this instead:

```rust
/// A Draft-21 §14.5.2 status type: 0x00–0x03 or 0x0C–0x0F.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub struct StatusType(u8); // private field: only constructible via TryFrom

impl TryFrom<u32> for StatusType {
    type Error = StatusListError;
    fn try_from(v: u32) -> Result<Self, Self::Error> {
        match v {
            0..=3 | 12..=15 => Ok(Self(v as u8)),
            _ => Err(StatusListError::InvalidStatusList(format!(
                "status value {v} is not a supported Draft-21 status type; accepted values are 0-3 and 12-15"
            ))),
        }
    }
}

impl StatusType {
    fn required_bits(self) -> BitFlag {
        let bits = match self.0 { 0 | 1 => 1, 2 | 3 => 2, _ => 4 };
        BitFlag::new(bits).expect("1, 2 and 4 are valid bit widths")
    }
}
```

What this buys:
- **One registry check, one error message.** `request.rs` deserializes straight into `StatusType`, so `Serialize` can't fail and the duplicated strings go away.
- **`determine_bits` becomes `max(required_bits)`**, with no dead 8-bit arm and no `unreachable!()`.
- **`StatusList.bits: BitFlag`** replaces the `validate_bits` calls (lines 129/162/228/342) and puts the currently unused `BitFlag` to work. Stored data still needs a single `BitFlag::new(row.bits)` check when it's loaded from the database.
- Registering a new status type later means changing one match arm.

it would also naturally close #152.

martcpp (changes requested):

Ngha-Boris:
Hello @Hermann-Core  and @martcpp please recheck

Hermann-Core (dismissed):
All four concerns from the previous round are now resolved.

Christiantyemele (dismissed):

src/domain/models/status_list.rs
martcpp: @Ngha-Boris did you consider this ??

src/domain/models/status_list.rs
```diff
@@ -196,19 +221,13 @@ fn determine_bits(
...
-                .next_power_of_two()
-                .trailing_zeros();
-            if bits_needed == 0 {
-                return Err(StatusListError::InvalidStatusList(
-                    "Status value too large".to_string(),
-                ));
-            }
-            bits_needed as usize
-        }
+        _ => unreachable!("status_value rejects values above 255"),
```
martcpp: return and error here so it fellow our coding style even if it unreachable  by design

Ngha-Boris:
Hello @martcpp please check this so we can merge

src/domain/models/status_list.rs
```diff
@@ -163,17 +168,53 @@ impl StatusList {
             lst: encode_compressed(&status_array)?,
         })
     }
+
+    pub(crate) fn token_representation(&self) -> Result<(u8, String, Vec<u8>), StatusListError> {
+        validate_bits(self.bits as usize)?;
+
+        let raw = decode_compressed(&self.lst)?;
```
martcpp: `raw` is only used in the `is_empty()` branch, which is the rare legacy case. For every normal list, each GET now does: base64-decode → full zlib inflate → discard → base64-decode the same string again. This is the hottest endpoint in the service, and the cost scales with list size, so a 100k-entry list pays an inflate per request for nothing.

The JWT path then discards `compressed` (line 294 in `token.rs`) and the CWT path discards `lst` (line 213), so each caller pays for a value it never reads.

Suggested shape  no decode in the common path, and the base64 decode only where CWT needs it:

```rust
/// The `lst` as it must appear in a token: the stored value, or a valid
/// zlib-compressed empty stream for legacy `lst = ""` rows.
pub(crate) fn token_lst(&self) -> Result<(u8, String), StatusListError> {
    validate_bits(self.bits as usize)?;
    let lst = if self.lst.is_empty() {
        encode_compressed(&[])?
    } else {
        self.lst.clone()
    };
    Ok((self.bits, lst))
}

/// Raw compressed bytes for the §4.3 CBOR byte string.
pub(crate) fn token_lst_bytes(&self) -> Result<(u8, Vec<u8>), StatusListError> {
    let (bits, lst) = self.token_lst()?;
    let bytes = base64url::decode(&lst).map_err(|err| {
        StatusListError::CorruptStoredList(format!("Invalid lst encoding: {err}"))
    })?;
    Ok((bits, bytes))
}
```

That also removes the unlabeled `(u8, String, Vec<u8>)` tuple, where both call sites currently discard a field with `_`.

Note this drops the incidental "is the stored stream inflatable" check. That check was costing an inflate per request to detect a condition that `update` already catches, and corruption surfaces anyway when someone writes to the list.

src/domain/models/status_list.rs
```diff
@@ -163,17 +168,53 @@ impl StatusList {
...
+        Status::ApplicationSpecific(value) if is_application_specific_status_value(*value) => {
+            Ok(*value)
+        }
+        Status::ApplicationSpecific(value) if *value <= 255 => {
+            Err(StatusListError::InvalidStatusList(format!(
+                "status value {value} is reserved for future registration; application-specific status values are 3 and 12 through 15"
+            )))
+        }
+        Status::ApplicationSpecific(value) => Err(StatusListError::InvalidStatusList(format!(
+            "status value {value} exceeds 8-bit capacity; maximum supported status value is 255"
```
martcpp: It says "maximum supported status value is 255", but 255 is rejected on the line above as reserved. The real maximum is **15**. An issuer that sends 300, reads this, and retries with 255 gets a different rejection.

Suggested text, used in all three places (here, `request.rs:30` and `request.rs:53`):

> `status value {v} is not a supported Draft-21 status type; accepted values are 0-3 and 12-15`

The new 400 test in `publish_status.rs` asserts `contains("exceeds 8-bit capacity")`, so it needs updating in the same commit.

src/domain/models/status_list.rs
```diff
@@ -195,20 +236,13 @@ fn determine_bits(
...
-                .next_power_of_two()
-                .trailing_zeros();
-            if bits_needed == 0 {
-                return Err(StatusListError::InvalidStatusList(
-                    "Status value too large".to_string(),
-                ));
-            }
-            bits_needed as usize
-        }
+        _ => unreachable!("status_value rejects unsupported Draft-21 status values"),
```
martcpp: Thanks for dropping the dead `16..=255 => 8` arm. The `unreachable!()` is still worth replacing: it holds only because `status_value()` — a different function — filtered the input first. If the accepted range is ever widened without updating this match, this panics inside a request task instead of returning an error.

`_ => return Err(StatusListError::InvalidStatusList(...))` costs nothing here. The stronger fix is the `TryFrom<u32>` newtype (separate comment), which makes the match exhaustive by construction.

src/domain/models/status_list.rs
```diff
@@ -195,20 +236,13 @@ fn determine_bits(
...
     };
 
+    if let Some(bits) = original_bits {
+        validate_bits(bits)?;
```
martcpp: `validate_bits` now has five call sites (129, 162, 173, 243, 361). This one doesn't belong in `determine_bits`: the function's job is to pick a width, and callers on the stored-data path (129, 173) have already validated `original_bits`.

Keep it where stored data enters  129 (`update`), 173 (token emission) and 361 (`decode_status_array`) and make 162 in `create_with_bits` a `debug_assert!`, since that input is internal.

src/domain/models/status_list.rs
```diff
@@ -163,17 +168,53 @@ impl StatusList {
...
+        ))),
+    }
+}
+
+pub(crate) fn is_application_specific_status_value(value: u32) -> bool {
+    matches!(value, 3 | 12..=15)
+}
+
+fn validate_bits(bits: usize) -> Result<(), StatusListError> {
+    match bits {
```
martcpp, line 239: `let (bits, _, lst_bytes)`  the discarded `String` was built by cloning and, on the JWT path at line 294, the discarded `Vec<u8>` was built by a base64 decode. See the note on `status_list.rs:175`: splitting this into two methods means neither caller pays for what it doesn't use.

martcpp (changes requested):
Still  open: no migration for legacy rows with `bits` ∉ {1, 2, 4, 8}

`git tag --contains 85799b8` shows the `>= 256` rule shipped in every release from v1.0.0 to v1.2.0, so production can hold rows with `bits` of 9 or 13.

The emission guard makes this **more** severe, deliberately. Those lists used to serve a non-conformant token on GET; now GET returns 500 as well as PATCH. Failing closed is the right call, but it means such a list is completely unusable — unreadable *and* unrevocable — with no tool to repair it.

Before merge this PR needs:

1. **An audit**  `SELECT list_id, bits FROM status_lists WHERE bits NOT IN (1,2,4,8)` against each environment. If the count is zero everywhere, say so in the PR and most of this concern goes away.
2. **A migration** for any rows found: lists holding no values >= 256 can be re-packed at the correct width; those that do hold them cannot be represented in Draft-21 at all and need a product decision.


### New: the ETag is computed before normalization

`generate_etag` (`utils/etag.rs:13`) hashes `record.status_list.lst`, which for a legacy row is still `""`. The token body now changes for those rows, but the validator doesn't.

A relying party that revalidates with its pre-deploy ETag therefore gets a **304** and keeps using the broken `lst: ""` token — the exact failure this PR set out to fix. The `window_start` term rotates the validator, so it self-heals after at most one token window, but the fix is defeated for that whole window.

Fix: hash the normalized representation, so the validator changes when the body does.

```rust
let (bits, lst) = record.status_list.token_lst()?;  // or inline the empty-lst normalization
hasher.update(bits.to_string().as_bytes());
hasher.update(lst.as_bytes());
```

`generate_etag` would become fallible, or take the already-normalized values from the caller at `get_status_list.rs:137`.

### Test gaps (carried over)

- **Pin the empty-list bytes.** The zlib round-trip test uses flate2 to check flate2's own output. `assert_eq!(result.lst, "eNoDAAAAAAE")` is what actually proves third-party decoders will cope.
- **`create_status_list_matches_two_bit_spec_vector` still isn't the spec vector.** Draft-21 §4.1 uses `[1,2,0,3,0,1,0,1,1,2,3,3]` → `0xC9 0x44 0xF9` → `lst = "eNo76fITAAPfAgc"`. The test uses a different array (positions 6 and 7 differ) and only asserts the length is 12. Value 3 is legal now, so it can finally assert the real vector.
- **No test updates an existing 8-bit list.** After this change `determine_bits` only ever returns 1, 2 or 4, so an 8-bit width now arises solely from legacy rows — and nothing exercises that path.

### Housekeeping

- The commits are plain `feat:`. Rejecting values >= 256, which have been accepted since v1.0.0, is breaking: use `feat!:` or add a `BREAKING CHANGE:` footer so release-plz produces the right bump.

Ngha-Boris:
Hello @martcpp please recheck this

docs/draft21-legacy-status-list-audit.md
martcpp, line 1: do he audit and attach the result not commit the result this doc not needed

martcpp (changes requested):

docs/draft21-legacy-status-list-audit.md
martcpp: @Ngha-Boris 
image
this his it

Ngha-Boris:
I ran the legacy bit-width audit and attached the terminal output.

image

martcpp (dismissed):

Ngha-Boris:
Hello @martcpp @Hermann-Core @Christiantyemele  please re approve as i just fix the conflicts

Hermann-Core (approved):

martcpp (approved):

PR #559, opened 2026-09-18
feat(crypto): support multi-curve keys via aws-lc-rs and zero-copy token minting

### LINKED ISSUE #524: Support multi-curve (P-256, P-384, P-521, Ed25519) and RSA keys with unified signing abstraction and aws-lc-rs
## Objective

Implement comprehensive multi-algorithm key loading and signing support for popular elliptic curves (P-256, P-384, Ed25519) and RSA algorithms when loading signing keys from the filesystem. Introduce a decoupled, consistent cryptographic abstraction interface for downstream consumers, and replace existing crypto libraries with `aws-lc-rs` in its latest version for signing operations.

## Context

Currently, the server's key loading and token signing implementations were tightly coupled to ECDSA P-256 (`ES256`) and the pure-Rust `p256` crate:
1. `src/setup.rs` and `src/utils/keygen.rs` only supported parsing P-256 EC PEM private keys.
2. `src/server/handlers/status_list/utils/token.rs` directly constructed `p256::ecdsa::SigningKey` and hardcoded the `ES256` algorithm.
3. CWT signing hardcoded COSE algorithm identifier `-7` (`ES256`).
4. Downstream components (`get_status_list.rs`, `cert_manager`, `file_watcher`) lacked a stable trait boundary.

Real-world deployments require support for broader cryptographic suites:
- **Elliptic Curves**: P-256 (`ES256`), P-384 (`ES384`), and Ed25519 (`EdDSA`). *(P-521 deferred per scope refinement)*.
- **RSA**: RSASSA-PKCS1-v1_5 (`RS256`) across 2048/3072/4096-bit keys. *(PS256 deferred per scope refinement)*.

Migrating to `aws-lc-rs` (AWS Libcrypto for Rust) in its latest version provides assembly-optimized signing routines, constant-time guarantees, and substantial signing throughput improvements over pure-Rust crates.

*Note: This ticket consolidates and expands upon #476.*

## Deliverables

- [x] **Cryptographic Abstraction Interface**:
  - Defined unified `SigningAlgorithm` enum (`ES256`, `ES384`, `EdDSA`, `RS256`) with mappings to JOSE algorithm strings and COSE identifiers.
  - Defined clean, decoupled `TokenSigner` trait completely isolating downstream consumers from low-level crypto primitives.
- [x] **`aws-lc-rs` Migration**:
  - Added `aws-lc-rs = "1"` to `Cargo.toml`, completely removed `p256` and `rand` crates, bumped `jsonwebtoken = "11.1"`.
  - Implemented key parsing and signature generation using `aws-lc-rs` backends.
- [x] **Filesystem Multi-Key Loading**:
  - Supported loading PKCS#8 (`BEGIN PRIVATE KEY`), SEC1 (`BEGIN EC PRIVATE KEY`), and PKCS#1 (`BEGIN RSA PRIVATE KEY`) PEM files. DER inputs are intentionally unsupported.
  - Automatically detect key type and algorithm from the PEM private-key structure.
  - Seamless hot-reloading via `FileWatcher` with X.509 certificate and private key public key validation across all algorithms.
- [x] **Dynamic Token Minting**:
  - Dynamically set the JWT `alg` header based on the loaded key's algorithm.
  - Dynamically set the COSE algorithm header in CWT tokens.
- [x] **Benchmarking & Testing**:
  - Added unit tests verifying token signing and external signature verification for all supported curves and RSA.
  - Added unit tests for dynamic JWT and CWT algorithm headers.
  - Tested PEM roundtrip and secret handling.

## Acceptance Criteria

- [x] Private keys in P-256, P-384, Ed25519, and RSA PEM formats load successfully from filesystem and sign valid JWTs and CWTs.
- [x] Downstream handlers depend on the high-level `TokenSigner` interface.
- [x] `cargo test` passes across all cryptographic and token tests.

Hermann-Core:
This PR upgrades the cryptographic engine of `status-list-server` from pure-Rust `p256` to `aws-lc-rs` and bumps `jsonwebtoken` to 11.1, introducing multi-algorithm signing capabilities and zero-copy token minting:

- **Multi-Algorithm Signing**: Adds unified `SigningKey` and `TokenSigner` supporting `ES256` (NIST P-256), `ES384` (NIST P-384), `EdDSA` (Ed25519), and `RS256` (RSA 2048).
- **Zero-Allocation Token Minting**: Pre-computes `jsonwebtoken::EncodingKey` once at key initialization, replacing the previous per-token PEM parse roundtrips with a zero-copy borrow (`signer.as_ref()`).
- **Flexible Key Ingestion**: Supports generic PEM private keys (PKCS#8 v1/v2, SEC1 / EC private keys) across local storage, Vault, and ACME renewal strategies.
- **Memory & Safety Hardening**: Discards raw private key DER bytes after backend initialization and stores public keys in immutable boxed slices (`Box<[u8]>`).

### 1. End-to-End Token Minting (Previous vs Updated)

Workload: Realistic status list with 100,000 entries (~12.5 KB compressed bitstring).

| Pipeline / Strategy | Format & Algorithm | Mean Latency (µs) | Median P50 (µs) | Tail P99 (µs) | Throughput (ops/s) | Speedup vs Previous |
| :--- | :--- | :---: | :---: | :---: | :---: | :---: |
| **Previous Implementation** | JWT — ES256 (P-256) | 43.84 µs | 41.92 µs | 82.38 µs | 22,809 ops/s | 1.00x *(Baseline)* |
| **Uncached (Single Parse)** | JWT — ES256 (P-256) | 28.83 µs | 28.50 µs | 31.88 µs | 34,681 ops/s | 1.52x faster |
| **Updated Impl (This PR)** | **JWT — ES256 (P-256)** | **26.67 µs** | **26.17 µs** | **31.33 µs** | **37,498 ops/s** | **1.64x faster (+64.4%)** 🚀 |
| **Updated Impl (This PR)** | **JWT — EdDSA (Ed25519)** | **14.39 µs** | **14.12 µs** | **29.29 µs** | **69,476 ops/s** | **3.05x faster (+205%)** ⚡ |
| **Previous Implementation** | CWT — ES256 (Parse PEM) | 29.04 µs | 28.54 µs | 33.33 µs | 34,430 ops/s | 1.00x *(CWT Base)* |
| **Updated Impl (This PR)** | **CWT — ES256 (P-256)** | **18.08 µs** | **17.96 µs** | **19.54 µs** | **55,325 ops/s** | **1.61x faster (+60.7%)** 🚀 |
| **Updated Impl (This PR)** | **CWT — EdDSA (Ed25519)** | **7.97 µs** | **7.92 µs** | **8.88 µs** | **125,507 ops/s** | **3.65x faster (+265%)** ⚡ |

### 2. Visual Comparison

```
JWT End-to-End Latency (Mean, Lower is Better):
Previous Impl (Full Parse Roundtrip) [████████████████████████████████████████] 43.84 µs (22,809 ops/s)
Updated Impl (Zero-Copy ES256) [PR]   [████████████████████████                ] 26.67 µs (37,498 ops/s) -39.2% latency
Updated Impl (Zero-Copy Ed25519) [PR] [█████████████                           ] 14.39 µs (69,476 ops/s) -67.2% latency ⚡

CWT End-to-End Latency (Mean, Lower is Better):
Previous Impl (Parse PEM per-token)  [████████████████████████████████████████] 29.04 µs (34,430 ops/s)
Updated Impl (COSE Sign1 ES256) [PR] [█████████████████████████               ] 18.08 µs (55,325 ops/s) -37.7% latency
Updated Impl (COSE Sign1 Ed25519)[PR][███████████                             ]  7.97 µs (125,507 ops/s) -72.6% latency ⚡
```

- **P99 Tail Latency**: Dropped from **82.38 µs** to **31.33 µs** (-62.0%) by eliminating runtime heap allocation and parsing jitter.
- **Ed25519 CWT**: Achieves **125k ops/s** and sub-10µs P99 latency (**8.88 µs**).

### 3. Raw Signing Throughput (`aws-lc-rs`)

Pure signature generation (`TokenSigner::sign`) on 256-byte payload (5,000 iterations):

| Algorithm | Curve / Family | Mean Latency (µs) | P50 (µs) | P99 (µs) | Throughput (ops/s) | vs ES256 |
| :--- | :--- | :---: | :---: | :---: | :---: | :---: |
| **ES256** | ECDSA P-256 (NIST) | 16.78 µs | 16.46 µs | 20.79 µs | **59,578 ops/s** | 1.00x *(Base)* |
| **ES384** | ECDSA P-384 (NIST) | 69.90 µs | 68.00 µs | 100.62 µs | **14,306 ops/s** | 0.24x |
| **EdDSA** | Ed25519 (Curve25519) | **6.69 µs** | **6.62 µs** | **7.29 µs** | **149,429 ops/s** | **2.51x faster** ⚡ |
| **RS256** | RSA 2048 PKCS#1 v1.5 | 490.19 µs | 482.46 µs | 675.58 µs | **2,040 ops/s** | 0.03x |

## Local Setup & Replication Guide for Reviewers

To replicate these benchmark results on your machine:

1. **Environment**:
   - Rust toolchain: Rust 1.86+ (2024 edition).
   - Build flags: Run benchmarks with `--release` (opt-level 3, LTO enabled).
   - Platform: Hardware acceleration for `aws-lc-rs` is automatically enabled on ARM64 (NEON) and x86-64 (AVX2/BMI2).

2. **Benchmark Harness Setup**:
   - The measurements compare:
     1. **Baseline**: Emulating the previous `p256` path: `Keypair::from_pkcs8_pem(&pem)` -> `to_pkcs8_pem_bytes()` -> `EncodingKey::from_ec_pem(&bytes)` -> `jsonwebtoken::encode` on every iteration.
     2. **PR Implementation**: Pre-computing `SigningKey` once -> borrowing via `AsRef<jsonwebtoken::EncodingKey>` into `issue_jwt` with zero key allocation on the hot path.
   - Sample status list bitstring: 100,000 entries (12.5 KB zlib-compressed bitstring).
   - Iterations: 5,000 iterations per benchmark, preceded by 500 warmup iterations to prime instruction cache and branch predictors, using `std::hint::black_box` around inputs and outputs.


Ngha-Boris (commented):
Hello @Hermann-Core thank you for this pull request and for the solid migration so far, could you please check the comments I left below

src/server/handlers/status_list/utils/token.rs
```diff
@@ -284,7 +287,7 @@ fn build_x5chain(cert_chain: &[String]) -> Result<CborValue, StatusListError> {
 
 fn issue_jwt(
     status_record: &StatusListRecord,
```
Ngha-Boris, line 286: **Issue:** `issue_jwt` is tightly coupled to `SigningKey` instead of the `TokenSigner` trait.

`issue_cwt` correctly abstracts over `&impl TokenSigner`, but `issue_jwt` requires `&SigningKey` specifically. This undermines the decoupled `TokenSigner` abstraction promised in the ticket deliverables.

**Suggestion:** Extend `TokenSigner` with a method yielding a JWT-compatible signer (e.g., `fn as_jwt_encoding_key(&self) -> Option<&EncodingKey>`) so `issue_jwt` can accept `&impl TokenSigner`. This keeps the trait as the primary boundary for both JWT and CWT paths.

src/server/handlers/status_list/utils/token.rs
```diff
@@ -235,27 +231,34 @@ fn issue_cwt(
...
             CWT_TYPE,
             CborValue::Text(STATUS_LISTS_CWT_TYPE_VALUE.into()),
         )
         .build();
 
-    let signing_key = keypair.signing_key();
-
+    let mut sign_err = None;
     let sign1 = CoseSign1Builder::new()
         .protected(protected)
```
Ngha-Boris, line 253: **Issue:** Error handling in CWT signing uses a side-effect variable.

The `create_signature` closure captures errors via an outer `sign_err` variable and returns an empty `Vec` on failure. This is brittle and relies on `build()` succeeding with an invalid signature.

**Suggestion:** If the `CoseSign1Builder` API doesn't allow returning `Result` directly, wrap the error capture in a small struct or at minimum add a comment explaining the empty-Vec fallback convention and why the builder API forces this pattern.

src/server/handlers/status_list/utils/token.rs
```diff
@@ -304,16 +307,83 @@ fn issue_jwt(
...
+        ];
+
+        for key in &test_keys {
+            let token = issue_jwt(&record, key, &cert_chain, &None, 1000, 2000, 300).unwrap();
+            let header = jsonwebtoken::decode_header(&token).unwrap();
+            let expected_alg: jsonwebtoken::Algorithm = key.algorithm().into();
+            assert_eq!(header.alg, expected_alg);
+            assert_eq!(header.typ.as_deref(), Some(STATUS_LISTS_HEADER_JWT));
+        }
+    }
```
Ngha-Boris, line 434: **Issue:** CWT tests only assert algorithm headers, not signature validity.

`test_dynamic_cwt_alg_header` verifies the COSE header but does not verify the token can be verified externally. This is weaker than the JWT tests which decode and verify claims.

**Suggestion:** Add signature verification for CWT using `CoseSign1::verify_signature` (or equivalent coset verification) with the public key to match the thoroughness of the JWT round-trip tests.

src/domain/ports.rs
```diff
@@ -80,13 +83,43 @@ pub trait StatusListSnapshotRepo: Send + Sync + 'static {
...
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct SigningMaterial {
     /// Base64 DER-encoded x509 certificate chain parts for JWT `x5c` and CWT
     /// `x5chain`.
     pub certificate_chain: Option<Vec<String>>,
-    /// PKCS#8 PEM-encoded signing key.
+    /// PEM-encoded signing key (supports PKCS#8, SEC1, or PKCS#1 format).
     pub signing_key_pem: String,
+    /// Pre-parsed, thread-safe signing key instance.
```
Ngha-Boris: **Issue:** `SigningMaterial` embeds the concrete `SigningKey` type, breaking domain/utils decoupling.

The domain port `SigningMaterial` contains `Arc<SigningKey>` (a utils-layer concrete type). The ticket states "clean, decoupled `TokenSigner` trait completely isolating downstream consumers," but `CertificateProvider` returns a concrete `SigningKey`.

**Suggestion:** Change the field to `pub signing_key: Arc<dyn TokenSigner>`. Update constructors accordingly. This respects the hexagonal architecture boundary and ensures the domain layer does not leak `aws-lc-rs` implementation details.

src/domain/ports.rs
```diff
@@ -80,13 +83,43 @@ pub trait StatusListSnapshotRepo: Send + Sync + 'static {
 }
 
 /// Certificate chain and signing key captured from one provider snapshot.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
```
Ngha-Boris: **Issue:** `PartialEq` / `Eq` derive removed from `SigningMaterial` without explanation.

The previous struct derived `PartialEq` and `Eq`. The new `SigningKey` (and `Arc<SigningKey>`) lacks them, so this is a silent breaking change for any code comparing signing material instances.

**Suggestion:** Implement `PartialEq` and `Eq` manually by comparing `algorithm()` and `public_key_bytes()`. If you switch to `Arc<dyn TokenSigner>`, implement equality via those trait methods.

src/domain/ports.rs
```diff
@@ -80,13 +83,43 @@ pub trait StatusListSnapshotRepo: Send + Sync + 'static {
...
+}
+
+impl SigningMaterial {
+    /// Construct signing material by parsing the PEM key into a cached `SigningKey`.
+    pub fn new(
+        certificate_chain: Option<Vec<String>>,
+        signing_key_pem: String,
+    ) -> Result<Self, crate::utils::crypto::Error> {
+        let signing_key = SigningKey::from_pem(&signing_key_pem)?;
+        Ok(Self {
```
Ngha-Boris: **Issue:** `SigningMaterial::new` returns `crate::utils::crypto::Error`.

The domain port constructor returns an error type from the `utils` layer. This couples the domain to a specific utils error type, violating the port/boundary abstraction.

**Suggestion:** Define a domain-level error variant or convert `crypto::Error` into `StatusListError` inside the constructor, so callers of the domain port do not need to import `crate::utils::crypto`.

src/utils/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+}
+
+struct SigningKeyInner {
+    pair: KeyPairInner,
+    encoding_key: jsonwebtoken::EncodingKey,
+    rng: SystemRandom,
+}
+
+enum KeyPairInner {
+    Ecdsa(EcdsaKeyPair),
```
Ngha-Boris, line 67: **Issue:** `to_pkcs8_der` for RSA may return PKCS#1 content wrapped in a PKCS#8 PEM tag.

For `RsaKeyPair`, `to_pkcs8_der` calls `kp.as_der()`. The `aws-lc-rs` documentation does not clearly guarantee this returns PKCS#8 for RSA. If it returns PKCS#1, then `to_pkcs8_pem` wraps it with `-----BEGIN PRIVATE KEY-----` but the body is PKCS#1, producing an invalid PEM.

**Suggestion:** Add a PEM roundtrip test for RS256 (load PEM → export via `to_pkcs8_pem` → reload) to confirm correctness. If `as_der()` returns PKCS#1, wrap it in a proper `PrivateKeyInfo` structure before PEM encoding.

src/utils/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+
+        // JWT sign & verify using AsRef<EncodingKey>
+        let header = Header::new(key.algorithm().into());
+        let claims = TestClaims {
+            sub: "test-eddsa".into(),
+            exp: 9999999999,
+        };
+        let token = encode(&header, &claims, key.as_ref()).unwrap();
+
+        let decoding_key = DecodingKey::from_ed_der(key.public_key_bytes());
```
Ngha-Boris: **Issue:** PEM roundtrip test only covers ES256.

`test_pem_roundtrip` only exercises P-256. Multi-curve support should include roundtrip tests for ES384, EdDSA, and RS256.

**Suggestion:** Parameterize the roundtrip test over all supported `SigningAlgorithm` variants, or add explicit test cases for `Es384`, `EdDsa`, and `Rs256`.

Cargo.toml
```diff
@@ -202,7 +201,6 @@ tikv-jemallocator = "0.7"
 [[bin]]
 name = "status-list-server"
 path = "src/main.rs"
-required-features = ["memory"]
```
Ngha-Boris, line 205: **Issue:** `required-features = ["memory"]` removed from `[[bin]]`.

This change is unrelated to the cryptographic ticket. Removing it means `cargo build` without the `memory` feature will still attempt to build the binary, which may fail or produce an unexpected artifact.

**Suggestion:** Revert the removal of `required-features = ["memory"]` unless there is a specific justification documented in the PR description.

src/utils/cert_manager.rs
```diff
@@ -412,8 +412,8 @@ impl CertManager {
 
         // If the secret does not exist, try to generate and store a new one
         warn!("No existing server secret found. Generating a new one...");
-        let keypair = Keypair::generate()?;
-        let key_pem = keypair.to_pkcs8_pem()?;
+        let key = SigningKey::generate(SigningAlgorithm::Es256)?;
```
Ngha-Boris, line 415: **Issue:** Auto-generated keys hardcode `Es256` without configuration.

`signing_key_pem()` hardcodes `SigningAlgorithm::Es256` when generating a new key. If the operator deploys with an RS256 certificate, ACME provisioning will mismatch because the generated key is always P-256.

**Suggestion:** Make the default algorithm configurable on `CertManager` (e.g., `with_default_signing_algorithm`) or document clearly that auto-provisioned keys are always P-256 and RS256 deployments must supply pre-generated keys.

src/outbound/cert.rs
```diff
@@ -172,16 +174,14 @@ pub(crate) fn validate_signing_material(
             )))
         })?;
     let cert_public_key = certificate.public_key().subject_public_key.data.as_ref();
-    let key_public_key = keypair.verifying_key().to_sec1_point(false);
-    if cert_public_key != key_public_key.as_bytes() {
+    let key_public_key = signing_key.public_key_bytes();
+    if cert_public_key != key_public_key {
         return Err(StatusListError::Backend(Box::new(std::io::Error::new(
             std::io::ErrorKind::InvalidData,
             "certificate public key does not match signing key",
```
Ngha-Boris, line 179: **Issue:** Public key comparison uses raw DER bytes directly.

`cert_public_key != key_public_key` compares raw ASN.1 bitstring bytes. This can be brittle if the certificate encodes the public key with different optional fields or padding.

**Suggestion:** Consider parsing both public keys into their respective SPKI structures and comparing algorithm + key material more robustly. If direct byte comparison is intentional, add a comment explaining that both sources are normalized to the same DER encoding.

src/utils/cert_manager/strategy.rs
```diff
@@ -211,17 +211,17 @@ fn decode_base64_text(value: &str) -> Option<Vec<u8>> {
         .ok()
 }
 
-fn normalize_pkcs8_key(signing_key: Vec<u8>) -> Result<String, CertError> {
+fn normalize_signing_key(signing_key: Vec<u8>) -> Result<String, CertError> {
```
Ngha-Boris, line 214: **Issue:** `normalize_signing_key` name no longer communicates output format.

The function always normalizes to PKCS#8 PEM regardless of input format, but the generic name obscures this.

**Suggestion:** Rename to `normalize_to_pkcs8_pem` for clarity, or document the output format explicitly in the function doc comment.

src/utils/crypto.rs
martcpp: hmmm no that important he it it not just to decode incoming file and move on but then good to have more code

src/utils/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+    UnsupportedKeygen(String),
+}
+
+/// Decoupled signing trait isolating consumers from low-level cryptographic backends.
+pub trait TokenSigner: Send + Sync {
+    /// Return the algorithm associated with this signer.
+    fn algorithm(&self) -> SigningAlgorithm;
+
+    /// Sign arbitrary payload bytes returning raw signature bytes.
+    fn sign(&self, data: &[u8]) -> Result<Vec<u8>, Error>;
```
ndefokou: **TokenSigner::sign has no documented encoding contract.** The trait returns raw signature bytes, but ECDSA signatures are format-sensitive: this happens to return IEEE P1363 raw R‖S (correct for BOTH JWT and COSE ES256/ES384), and RSA returns PKCS#1 v1.5. A downstream caller cannot infer this from the signature alone. The `sign` name implying a universal raw signature is misleading in a JWT context.

**Why it matters:** This is the public abstraction the whole ticket is built around. Misuse (e.g. treating the ECDSA bytes as ASN.1 DER, as several JWT libs expect) would produce undecodable tokens.

**Suggestion:** Document the exact output encoding on the trait (e.g. "raw R‖S for ECDSA (IEEE P1363), raw Ed25519, PKCS#1 v1.5 for RSA"), or better, expose higher-level `sign_jwt`/`sign_cose` methods so encoding is handled by the abstraction and consumers can stay fully decoupled.

src/utils/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+/// Unified signing key supporting ECDSA (P-256, P-384), Ed25519, and RSA (RS256).
+pub struct SigningKey {
+    algorithm: SigningAlgorithm,
+    inner: SigningKeyInner,
+    public_key: Box<[u8]>,
+}
+
+struct SigningKeyInner {
+    pair: KeyPairInner,
+    encoding_key: jsonwebtoken::EncodingKey,
```
ndefokou: **Two independent private-key copies are retained in memory.** `SigningKeyInner` stores both the aws-lc-rs `KeyPairInner` AND a `jsonwebtoken::EncodingKey` (a second, separate copy of the key material), and `SigningMaterial` additionally retains the full PEM `String`. The PR claims to "discard raw private key DER bytes" and harden memory, but the key material lives on in at least two forms.

**Why it matters:** For one key this is trivial, but it doubles+ the in-memory secret surface and partially contradicts the stated memory-hardening goal. Key material is also never explicitly zeroized (no `zeroize` is applied).

**Suggestion:** If the EncodingKey must be precomputed for the zero-copy JWT path, keep it, but note the duplication in a comment, and consider zeroizing buffers (`zeroize::Zeroizing`) that do hold copies for transient operations. At minimum document that this is an intentional tradeoff vs. re-parsing PEM.

ndefokou (commented):
@Hermann-Core 
I checked the branch out and validated it locally: `cargo build` (incl. `--features acme`), full lib suite (197/197 pass), crypto/token/cert/cert_manager tests all green, clippy clean for the new module. I also **empirically verified** the two untested ingestion paths (SEC1 `BEGIN EC PRIVATE KEY`, PKCS#1 `BEGIN RSA PRIVATE KEY` at 2048/3072 bits) using OpenSSL-generated keys — they load, auto-detect, and produce externally verifiable signatures. The cert public-key cross-validation is format-correct for all four key families (verified against aws-lc-rs 1.18 source).

No functional correctness defects found. The items below are inline — two are worth treating as must-fix (secret-leak hazard in `SigningMaterial`'s Debug, and the partial decoupling of the JWT path); the rest are test-coverage and maintainability improvements.

src/utils/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+            }
+            SigningAlgorithm::Es384 => {
+                let doc = EcdsaKeyPair::generate_pkcs8(&ECDSA_P384_SHA384_FIXED_SIGNING, &rng)?;
+                Self::from_pkcs8_der(doc.as_ref())
+            }
+            SigningAlgorithm::EdDsa => {
+                let doc = Ed25519KeyPair::generate_pkcs8(&rng)?;
+                Self::from_pkcs8_der(doc.as_ref())
+            }
+            SigningAlgorithm::Rs256 => Err(Error::UnsupportedKeygen("RS256".into())),
```
ndefokou, line 109: **`SigningKey::generate(Rs256)` deliberately errors with `UnsupportedKeygen`.** This is reasonable (RSA keygen is expensive and rarely needed), but it is surprising for a `SigningKey::generate` that otherwise advertises full multi-algorithm support, and there is no test or doc asserting this behavior.

**Why it matters:** Downstream code calling `generate` generically across algorithms will only discover RSA is unsupported at runtime. It's a config/UX footgun.

**Suggestion:** Add an explicit unit test asserting `generate(RS256)` returns `Error::UnsupportedKeygen`, and note the limitation in the method's doc comment so it's discoverable.

src/utils/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+        }
+    }
+
+    /// Load and automatically detect signing key from PEM string (supports PKCS#8, SEC1, or PKCS#1 format).
+    pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
```
ndefokou, line 114: **SEC1 (`BEGIN EC PRIVATE KEY`) and PKCS#1 (`BEGIN RSA PRIVATE KEY`) ingestion paths are entirely untested**, even though they are explicit deliverables and the only non-trivial custom code (`convert_sec1_to_pkcs8`, `convert_pkcs1_to_pkcs8`). `from_pem` only ever exercises the `PRIVATE KEY` (PKCS#8) branch in the test suite — both existing test PEMs (`ec-private.pem`, `gcloud_test_key.dummy.pem`) are PKCS#8.

**Why it matters:** These two conversion functions are hand-written ASN.1 wrappers; a regression would silently reject SEC1 or PKCS#1 keys in production.

**Suggestion:** Add tests that (a) parse a generated SEC1 EC P-256 key and verify a signature, and (b) parse PKCS#1 RSA at 2048/3072/4096 bits and verify a signature. I verified all of these work correctly with real OpenSSL-generated keys, so the coverage just needs to be locked in. (I confirmed ES256's `pem_roundtrip` only covers PKCS#8 and `generate()` for ES384/EdDSA/RS256.)

src/server/handlers/status_list/utils/token.rs
```diff
@@ -284,7 +287,7 @@ fn build_x5chain(cert_chain: &[String]) -> Result<CborValue, StatusListError> {
 
 fn issue_jwt(
     status_record: &StatusListRecord,
-    keypair: &Keypair,
+    signer: &SigningKey,
```
ndefokou: **The JWT path bypasses the `TokenSigner` abstraction and couples to the concrete `SigningKey`.** `issue_jwt` takes `signer: &SigningKey` (this line) and calls `signer.as_ref()` to obtain a `jsonwebtoken::EncodingKey`, while `issue_cwt` correctly takes `&impl TokenSigner`. So a JWT consumer cannot write against the trait — they must know the concrete type, its `AsRef<EncodingKey>`/`From<SigningKey>` impls, and the `jsonwebtoken` crate.

**Why it matters:** The ticket's headline deliverable is a "decoupled TokenSigner trait completely isolating downstream consumers from low-level crypto primitives." That goal is only half-met. It also means JWT (the default output format) and CWT are maintained on two parallel signing implementations (jsonwebtoken's EncodingKey vs. aws-lc-rs raw `sign`), which must be kept in sync whenever algorithms change.

**Suggestion:** Either add a trait method that yields a format-compatible signer (e.g. `fn sign_jwt(&self, header, claims)`) and make `issue_jwt` generic over `TokenSigner`, or document explicitly that JWT is coupled to `SigningKey` while CWT is trait-driven. Unifying on one entry point reduces drift risk.

src/server/handlers/status_list/utils/token.rs
```diff
@@ -235,27 +231,34 @@ fn issue_cwt(
...
-    let signing_key = keypair.signing_key();
-
+    let mut sign_err = None;
     let sign1 = CoseSign1Builder::new()
         .protected(protected)
         .payload(payload)
-        .create_signature(&[], |payload| {
-            let signature: Signature = signing_key.sign(payload);
-            signature.to_vec()
+        .create_signature(&[], |tbs| match signer.sign(tbs) {
```
ndefokou: **The CWT sign-error handling via a captured `sign_err` flag + empty `Vec` is convoluted.** `create_signature`'s closure must return signature bytes, so on a sign error you return an empty vec, build a garbage `CoseSign1`, then bail via `sign_err` before serialization. It works (coset's `build()` does no signature-length assertion, so no panic), but it's fragile and easy to misread.

**Why it matters:** If someone later adds a length assertion upstream or reorders the `if let` check, this silently produces an invalid CWT or panics instead of a clean error.

**Suggestion:** Compute the signature before constructing the builder, e.g. sign the TBS outside, and only call `create_signature` with a closure that cannot fail — or use coset's `try_`-style builder if available. Simpler and obviously correct.

src/domain/ports.rs
ndefokou: **`SigningMaterial` derives `Debug` while holding the full private-key PEM in `signing_key_pem`.** This is a latent secret-leak hazard: `SigningKey`'s `Debug` is properly redacted (algorithm + pub-key length only), but a single future `tracing::debug!("{:?}", signing_material)` — a very common debugging habit — would dump the private key into the logs.

**Why it matters:** Logs are often shipped to aggregators / SIEM and are a primary secret-exfiltration vector. Currently nothing logs it, so it is latent, not active — but it should not be one `Debug` away from leaking.

**Suggestion:** Replace the derived `Debug` with a redacted `impl Debug for SigningMaterial` (omit `signing_key_pem`), or make the PEM a `secrecy::SecretString` (the crate is already a dependency) so it never formats.

src/domain/ports.rs
```diff
@@ -80,13 +83,43 @@ pub trait StatusListSnapshotRepo: Send + Sync + 'static {
...
+#[derive(Clone, Debug)]
 pub struct SigningMaterial {
     /// Base64 DER-encoded x509 certificate chain parts for JWT `x5c` and CWT
     /// `x5chain`.
     pub certificate_chain: Option<Vec<String>>,
-    /// PKCS#8 PEM-encoded signing key.
+    /// PEM-encoded signing key (supports PKCS#8, SEC1, or PKCS#1 format).
     pub signing_key_pem: String,
+    /// Pre-parsed, thread-safe signing key instance.
+    pub signing_key: Arc<SigningKey>,
```
ndefokou: **`SigningMaterial` can be constructed in an internally inconsistent state.** `with_signing_key` accepts an arbitrary `signing_key_pem` alongside an already-built `signing_key`; nothing enforces that the PEM corresponds to the parsed key (e.g. `src/outbound/memory.rs:262` passes an empty `""` PEM with a real key).

**Why it matters:** A struct named `SigningMaterial` implies its fields are a coherent snapshot. If a caller later reads `signing_key_pem` (used for rotation diffing in cert_manager) and trusts it to match `signing_key`, the mismatch is a silent invariant violation.

**Suggestion:** Add a doc comment documenting the invariant ("`signing_key_pem` and `signing_key` must describe the same key"), or have `with_signing_key` just derive/coerce the PEM from the key when feasible. Non-blocking, but worth documenting.

src/outbound/cert.rs
```diff
@@ -172,16 +174,14 @@ pub(crate) fn validate_signing_material(
             )))
         })?;
     let cert_public_key = certificate.public_key().subject_public_key.data.as_ref();
-    let key_public_key = keypair.verifying_key().to_sec1_point(false);
-    if cert_public_key != key_public_key.as_bytes() {
+    let key_public_key = signing_key.public_key_bytes();
```
ndefokou, line 175: **RSA and EdDSA certificate/key matching is never exercised by the test suite.** `validate_signing_material` is the gate for loading RSA/EdDSA keys in the filesystem, inline, and store strategies, but every cert test uses `matching_cert_and_key` → an rcgen **EC P-256** cert. The only cert path under test is EC.

**Why it matters:** The byte comparison here must be format-correct per family (EC = uncompressed X9.62 point; RSA = RFC 8017 `RSAPublicKey`; Ed25519 = raw 32 bytes), each matching the X.509 `subjectPublicKey` payload. I verified from aws-lc-rs 1.18 source that all three are correct — but a regression would silently reject all RSA (a headline deliverable) or EdDSA deployments.

**Suggestion:** Add one test each for RSA and EdDSA cert+key validation (generate a self-signed cert matching an RSA/Ed25519 key and assert `validate_signing_material` succeeds), plus a deliberately mismatched case.

Cargo.toml
```diff
@@ -81,16 +81,19 @@ governor = "0.10"
 moka = { version = "0.12", features = ["future"] }
 
 # Cryptography and Security
-rand = "0.10"
+aws-lc-rs = "1"
```
ndefokou, line 84: **Out-of-scope binary semantic change:** `[[bin]] required-features = ["memory"]` was removed (Cargo.toml). This lets the binary build with no storage backend feature enabled, which may produce a binary that fails at runtime — and it is unrelated to the crypto ticket.

**Why it matters:** It silently expands the supported build matrix and could mask a misconfigured default-features setup.

**Suggestion:** If intentional, call it out in the PR description and ideally add a CI check that the binary behaves (or fails loudly) without a store feature. If not, restore `required-features = ["memory"]`.

src/utils/cert_manager.rs
Hermann-Core: Kept intentionally. This is the ACME-managed certificate-key creation path: the generated P-256 key is used immediately to obtain the matching ACME certificate, rather than selecting a token algorithm independently. Externally supplied key/certificate material supports the other algorithms. Adding an unimplemented algorithm switch here would create misleading configuration and can produce a certificate/key mismatch.

src/outbound/cert.rs
Hermann-Core: Kept the direct comparison. `x509-parser` exposes the decoded SubjectPublicKey BIT STRING payload, and aws-lc-rs exposes the same native public-key representation for each supported family. The implementation now documents that invariant and the tests cover matching P-256, Ed25519, and RSA certificate/key pairs.

src/utils/cert_manager/strategy.rs
Hermann-Core: No rename needed. PEM input is preserved after validation; only DER input is converted to PEM. The current name accurately describes that normalization behavior, while `normalize_to_pkcs8_pem` would incorrectly imply PEM input is rewritten.

Hermann-Core:
Hello @Ngha-Boris @ndefokou thank you for the reviews. Can you check again?

src/utils/crypto.rs
```diff
@@ -0,0 +1,406 @@
...
+                let pub_bytes = kp.public_key().as_ref().to_owned();
+                (KeyPairInner::Ecdsa(kp), pub_bytes.into())
+            }
+            SigningAlgorithm::EdDsa => {
+                let kp = Ed25519KeyPair::from_pkcs8(der)?;
+                let pub_bytes = kp.public_key().as_ref().to_owned();
+                (KeyPairInner::Ed25519(kp), pub_bytes.into())
+            }
+            SigningAlgorithm::Rs256 => {
+                let kp = RsaKeyPair::from_pkcs8(der)?;
```
ndefokou, line 154: **Blocking (security/defense-in-depth): enforce a minimum RSA modulus size.** `RsaKeyPair::from_pkcs8(der)` here accepts any RSA key aws-lc-rs can parse, with no lower bound on the modulus. The issue explicitly scopes RS256 to 2048/3072/4096-bit keys, but a weak (e.g. 512- or 1024-bit) key would be accepted and used to sign production RS256 tokens silently — no error, no warning. Consider rejecting moduli below 2048 bits here (or in `validate_signing_material`) using `kp.public_modulus_len()`, with a distinct error, and add a test proving such keys are refused.

src/utils/crypto.rs
```diff
@@ -0,0 +1,406 @@
...
+                        .expect("ES384 signature verifies");
+                }
+                SigningAlgorithm::EdDsa => {
+                    assert_eq!(signature.len(), 64);
+                    UnparsedPublicKey::new(&ED25519, key.public_key_bytes())
+                        .verify(message, &signature)
+                        .expect("Ed25519 signature verifies");
+                }
+                SigningAlgorithm::Rs256 => {
+                    assert_eq!(signature.len(), 256);
```
ndefokou, line 340: **Non-blocking but fragile assertion + missing coverage.** `assert_eq!(signature.len(), 256)` is only true for a 2048-bit RSA key. The issue claims support across "2048/3072/4096-bit keys," yet there is no test fixture for 3072- or 4096-bit RSA, and this hardcoded constant would silently break if a larger fixture were added. Suggest deriving the expected length from the key (e.g. `key.public_modulus_len()`) and adding 3072/4096-bit fixtures to actually demonstrate the claimed range.

src/server/handlers/status_list/utils/token.rs
```diff
@@ -304,16 +304,162 @@ fn issue_jwt(
...
+
+    let signature = signer
+        .sign(token.as_bytes())
         .map_err(|err| StatusListError::Backend(Box::new(err)))?;
+    token.push('.');
+    token.push_str(&base64url::encode(signature));
     Ok(token)
 }
+
+fn cose_algorithm(algorithm: SigningAlgorithm) -> Algorithm {
```
ndefokou, line 329: **Non-blocking maintainability: duplicate COSE-id mapping.** This `cose_algorithm()` match duplicates the mapping already defined as `SigningAlgorithm::cose_id()` in `src/domain/models/token.rs`. Two sources of truth for the same COSE identifier can drift. Consider deriving the coset `Algorithm` from `cose_id()`/the enum so there is a single authoritative mapping.

src/domain/ports.rs
```diff
@@ -80,19 +83,66 @@ pub trait StatusListSnapshotRepo: Send + Sync + 'static {
...
 /// Certificate chain and signing key captured from one provider snapshot.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone)]
 pub struct SigningMaterial {
     /// Base64 DER-encoded x509 certificate chain parts for JWT `x5c` and CWT
     /// `x5chain`.
     pub certificate_chain: Option<Vec<String>>,
-    /// PKCS#8 PEM-encoded signing key.
-    pub signing_key_pem: String,
+    /// Pre-parsed signer. The material deliberately does not retain private
```
ndefokou, line 91: **Non-blocking (doc accuracy).** The comment "The material deliberately does not retain private key encoding after a provider has validated and constructed it" is slightly misleading: the *signing* key (which holds the private key material) is retained in memory for the lifetime of the process — as it must be for signing. The comment could be read as claiming the key is zeroized/dropped. Suggest rewording (e.g. "does not retain the PEM/DER encoding") and, if practical, noting whether/where key material is zeroized so a future reader isn't misled.

ndefokou (dismissed):
## Review Summary

I reviewed the full PR diff (11 commits, base `develop`) and built the branch in an isolated worktree. `cargo test` for the crypto, token-minting, and cert-adapter suites all pass. This is a high-quality migration: the `SigningAlgorithm` enum, `TokenSigner` trait, and unified `SigningKey` cleanly decouple token encoders from crypto primitives; the aws-lc-rs migration is real (`p256` and the `rand` direct dependency are gone); and the dynamic JWT/CWT algorithm headers are correct (verified COSE IDs -7/-35/-8/-257 against the IANA registry).

The implementation fully satisfies the functional acceptance criteria. I've left inline comments; the only item I'd call blocking is a defense-in-depth hardening: enforcing a minimum RSA modulus size (2048-bit floor) so a weak key can never silently sign RS256 tokens. Everything else is non-blocking polish.

Approving as the architecture and correctness are sound.

Ngha-Boris (commented):
Re-review after fixes. The architecture and abstraction improvements are solid — `TokenSigner` trait fully decouples JWT/CWT paths, `SigningMaterial` no longer leaks concrete crypto types, and tests cover all algorithms end-to-end. Below are three remaining inline observations; they are non-blocking but worth tracking.

src/utils/cert_manager.rs
Ngha-Boris, line 415: **Issue:** Auto-generated server secrets hardcode `Es256`.

`signing_key_pem()` unconditionally generates `SigningAlgorithm::Es256` when no key exists in storage. An operator running ACME provisioning with an RSA certificate will receive a P-256 signing key that does not match the certificate algorithm.

**Suggestion:** Either add a `with_default_signing_algorithm` builder method on `CertManager` or document in the ACME setup guide that auto-provisioned keys are always P-256 and RS256 deployments must pre-seed a matching key via `StoreProvisioningStrategy::filesystem`.

src/outbound/cert.rs
```diff
@@ -172,16 +172,14 @@ pub(crate) fn validate_signing_material(
...
+    if cert_public_key != key_public_key {
         return Err(StatusListError::Backend(Box::new(std::io::Error::new(
             std::io::ErrorKind::InvalidData,
             "certificate public key does not match signing key",
         ))));
     }
```
Ngha-Boris, line 181: **Issue:** Public key comparison relies on raw DER byte equality without a comment.

`cert_public_key != key_public_key` assumes both X.509 `subjectPublicKeyInfo` and `signing_key.public_key_bytes()` produce identical byte-for-byte DER encodings. This happens to hold for all tested backends, but it is an implicit invariant that could break if a certificate issuer uses a different optional-parameter encoding (e.g., RSA with explicit NULL vs absent parameters).

**Suggestion:** Add a short comment above the comparison explaining that both values are normalized to the same SPKI DER encoding by `aws-lc-rs` and `x509-parser`, or parse both into `SubjectPublicKeyInfo` and compare algorithm OID + key bitstring separately.

src/utils/crypto.rs
Ngha-Boris, line 154: **Issue:** No minimum RSA modulus size enforced during key loading.

`RsaKeyPair::from_pkcs8` accepts any valid PKCS#8 RSA key. A deployment could accidentally load a 1024-bit (or smaller) key for `RS256`, which is below current NIST/BSI recommendations and would silently sign tokens with weak material.

**Suggestion:** After `RsaKeyPair::from_pkcs8(der)`, check `kp.public_modulus_len()` and reject keys below 256 bytes (2048 bits) with a new `Error::UnsupportedAlgorithm("RSA key must be at least 2048 bits".into())`.

Hermann-Core: Verified with a valid 1024-bit PKCS#8 fixture: aws-lc-rs rejects it in RsaKeyPair::from_pkcs8 with KeyRejected("TooSmall") before application code can inspect it. I added a regression test for that rejection; an additional application-side modulus check would therefore be unreachable. The existing 2048-bit RSA fixture continues to load and sign.

src/utils/crypto.rs
Hermann-Core: Updated the RSA assertion to derive its expected signature length from RsaKeyPair::public_modulus_len(). I am not adding 3072/4096 fixtures: production allocation already uses that same backend value and has no size-specific branch.

src/server/handlers/status_list/utils/token.rs
Hermann-Core: Keeping the exhaustive adapter mapping. SigningAlgorithm remains a project-owned domain type and must not expose coset types. Converting cose_id() through Algorithm::from_i64() would add a fallible runtime conversion to an invariant currently enforced by exhaustive matching.

src/domain/ports.rs
Hermann-Core: Clarified the comment: PEM/DER encoding is discarded, but parsed private-key material remains resident in the signer for its required signing lifetime.

src/utils/cert_manager.rs
Hermann-Core: The ACME path generates the key before creating the CSR, so a newly generated P-256 key is used to obtain its matching certificate. This path does not select an externally supplied RSA certificate. RSA deployments provide matching external key/certificate material, so a default-algorithm switch here would be misleading.

src/outbound/cert.rs
Hermann-Core: Added the requested explanation and P-384 coverage. The comparison is specifically between the certificate subjectPublicKey BIT STRING payload and aws-lc-rs matching family-specific public-key representation; it is not whole-SPKI DER comparison, so optional SPKI algorithm parameters are not part of the equality check.

src/utils/crypto.rs
Hermann-Core: Verified with a valid 1024-bit PKCS#8 fixture: aws-lc-rs rejects it in RsaKeyPair::from_pkcs8 with KeyRejected("TooSmall") before application code can inspect it. I added a regression test for that rejection; an additional application-side modulus check would therefore be unreachable. The existing 2048-bit RSA fixture continues to load and sign.

Hermann-Core:
@Ngha-Boris @ndefokou @martcpp Could you please take another look? Thanks.

ndefokou (approved):

ndefokou (dismissed):

Ngha-Boris (dismissed):

martcpp:
checking this @Hermann-Core

martcpp (commented):
good work @Hermann-Core  work on this final touch 
1. Read the file with tokio::fs::read (bytes, not a string).
2. If the bytes contain -----BEGIN , decode them as UTF-8 and call from_pem. Otherwise call a new SigningKey::from_der that tries PKCS#8, then SEC1, then PKCS#1.
3. Reuse from_der in normalize_signing_key so both provisioning paths behave the same.Add DER fixtures for all three formats. If DER is intentionally out of scope, please drop "and DER" from the issue and fix .env.template:44 instead. It still says "Private keys must be PKCS#8 PEM or DER".

src/server/handlers/status_list/utils/token.rs
martcpp: these derives serialize as "Es256" / "EdDsa" (Rust variant names), not the JOSE names "ES256" / "EdDSA" that jose_name() returns. Nothing uses them today, so drop Serialize, Deserialize until something needs them.

src/server/handlers/status_list/utils/token.rs
martcpp: this duplicates SigningAlgorithm::cose_id(), and production only uses this copy. cose_id() is only exercised by test_algorithm_identifiers. Meanwhile test_dynamic_cwt_alg_header computes its expected value by calling cose_algorithm() itself, so a wrong mapping here (e.g. Es384 => ES256) would still pass. Suggest deleting this function, using Algorithm::from_i64(signer.algorithm().cose_id()) in issue_cwt, and hard-coding the expected algorithms in the CWT test

src/server/handlers/status_list/utils/token.rs
```diff
@@ -140,13 +146,10 @@ async fn build_status_list_token_inner(
...
 
     tokio::task::spawn_blocking(move || {
-        let keypair = Keypair::from_pkcs8_pem(&signing_key_pem)
-            .map_err(|e| StatusListError::Backend(Box::new(e)))?;
-
         let token_bytes = match accept.as_str() {
             ACCEPT_STATUS_LISTS_HEADER_CWT => issue_cwt(
                 &status_record,
-                &keypair,
+                &*signing_key,
```
martcpp: RSA makes each public token response much more expensive, and nothing is cached

Every 200 from GET /status-lists/{list_id} signs a new token (iat = now, so there's no reuse), and that endpoint is public. Before this PR the only option was ES256, which is cheap. Now, loading an RSA key makes each response roughly 30–70× more CPU-expensive. I measured about 6.4 ms per RS256 signature with a 4096-bit key versus about 0.09 ms for Ed25519. This bites at token-expiry boundaries, when all clients re-fetch at once through the ExpiredToken path. The per-IP rate limiter doesn't help there, because the traffic is legitimate.

For this PR: please document ES256 as the recommended default and note RSA's CPU cost, and consider a startup warning for RSA keys of 3072 bits or more.

@Hermann-Core create a follow-up issue: anchor iat/exp to token_window (which already anchors the ETag) and cache signed tokens per list, format and window, keyed on the signing key so rotation invalidates them. Signing cost would then depend on the number of lists rather than on traffic

src/domain/ports.rs
```diff
@@ -80,19 +83,67 @@ pub trait StatusListSnapshotRepo: Send + Sync + 'static {
...
+impl PartialEq for SigningMaterial {
+    fn eq(&self, other: &Self) -> bool {
+        self.certificate_chain == other.certificate_chain
+            && self.signing_key.algorithm() == other.signing_key.algorithm()
+            && self.signing_key.public_key_bytes() == other.signing_key.public_key_bytes()
+    }
 }
 
+impl Eq for SigningMaterial {}
+
```
martcpp: these hand-written PartialEq/Eq impls look like they are only here to keep the old #[derive(PartialEq, Eq)]. I couldn't find any caller or test that compares SigningMaterial. They also define equality by public-key bytes, which depends on every TokenSigner returning the same byte format. A KMS signer returning SPKI DER would compare unequal for the same key. Suggest deleting both impls. If change detection is needed later, add an explicit, documented method instead of a general ==.

martcpp (changes requested):

src/server/handlers/status_list/utils/token.rs
martcpp: you can close this @ndefokou his already on the fellow up https://github.com/adorsys/status-list-server/issues/524

src/server/handlers/status_list/utils/token.rs
Hermann-Core: Updated in 64050cc3: `cose_id()` is now authoritative and the local adapter converts it with `Algorithm::from_i64`, so the domain remains independent of coset. The CWT test now uses explicit expected COSE algorithm constants, so a faulty mapping cannot validate itself.

src/server/handlers/status_list/utils/token.rs
Hermann-Core: Removed in 64050cc3. No serialized contract exists here, and Rust variant names would not be the JOSE identifiers; `jose_name()` remains the explicit serialization boundary.

src/domain/ports.rs
Hermann-Core: Removed in 64050cc3. There is no caller that needs general equality, and a future change-detection operation should define its own explicit, documented key-identity semantics rather than inherit an accidental byte-format contract.

Hermann-Core:
@martcpp Confirmed: DER ingestion is intentionally out of scope. Updated in 6eab2f93: store certificate and signing-key sources now accept PEM text only; the raw/base64 DER conversion paths and stale documentation were removed. PKCS#8, SEC1 EC, and PKCS#1 RSA remain supported as PEM labels. The signed-token caching/performance proposal is already tracked in #524, so I am keeping it out of this crypto-format PR.

Hermann-Core:
@martcpp @Ngha-Boris @ndefokou The latest review updates are in 64050cc3 and 6eab2f93: signer-boundary cleanup plus PEM-only certificate/key provisioning (including rejection tests and documentation). The resolved threads contain the rationale. Please re-review when convenient.

martcpp (changes requested):
Commit 6eab2f9 removes DER support (raw DER files and base64/base64url DER in secret storage) for both certificates and keys in the store strategy. That support first appeared in 8521443 and has shipped in every release from v1.0.0 to v1.2.0. The old .env.template told operators to use it.

Anyone who followed those docs will have certificate provisioning fail on upgrade, at startup or at the next renewal run.
There's no CHANGELOG entry for the change.
The issue still says "PEM and DER files".
Removing certificate DER support has nothing to do with key algorithms.
A DER file on the filesystem path fails with the UTF-8 read error (failed to read certificate PEM file …: stream did not contain valid UTF-8). That doesn't tell the operator what's wrong.
it will be best to  revert 6eab2f9. Instead, add SigningKey::from_der that tries PKCS#8, then SEC1, then PKCS#1, and use it in normalize_signing_key. That keeps compatibility and meets the issue. If the team really wants PEM-only, do it in a separate PR with:

a Breaking CHANGELOG entry,
a major version bump or a deprecation period,
an error message that includes an openssl conversion command.

Hermann-Core:
@martcpp You are right that DER was a documented released input path. We are intentionally withdrawing it under the now-explicit PEM-only provisioning contract, so I am not restoring raw/base64 DER or adding `SigningKey::from_der`. The PR title now carries the breaking-change marker that will survive the squash merge and drive release automation; `CHANGELOG.md` remains release-PR managed. Commit 80ee3917 adds an actionable filesystem diagnostic and documents PEM conversion commands.

docs/troubleshooting.md
```diff
@@ -400,18 +400,18 @@ non-disruptive but leaves the app on the previous credential until the file is v
...
-- `signing key PEM is not valid UTF-8: ...`
-- `failed to read certificate file '...': ...` / `failed to read signing key file '...': ...`
+- `{certificate|signing key} material must be PEM text`
+- `{certificate|signing key} material must be PEM text; file '...' is not valid UTF-8 (DER is unsupported; convert it to PEM)`
+- `failed to read certificate PEM file '...': ...` / `failed to read signing key PEM file '...': ...`
 - `store certificate key '...' was not found` / `store signing key '...' was not found`
 - Store validation: both-paths-and-keys, missing file, or missing key errors from the
   `StoreProvisioningStrategy` builder
 
 _Source: `src/utils/cert_manager/strategy.rs:92-222`, `src/utils/cert_manager/builder.rs:171`, `src/setup.rs:494-505` (`store` provider selection and `spawn_cert_rotation`)_
```
martcpp: Stale reference strategy.rs:92-222 in troubleshooting.md:409 the file is now 218 lines or so

src/utils/crypto.rs
```diff
@@ -0,0 +1,417 @@
...
+                let doc = Ed25519KeyPair::generate_pkcs8(&rng)?;
+                Self::from_pkcs8_der(doc.as_ref())
+            }
+            SigningAlgorithm::Rs256 => Err(Error::UnsupportedKeygen("RS256".into())),
+        }
+    }
+
+    /// Load and automatically detect signing key from PEM string (supports PKCS#8, SEC1, or PKCS#1 format).
+    pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
+        let parsed = pem::parse(pem_str.as_bytes())?;
```
martcpp: till parses only the first PEM block. With PEM now the only accepted format, output from openssl ecparam -genkey has no workaround except hand-editing the file

martcpp (changes requested):
Thanks for making PEM-only explicit and marking it !. I'm fine with that decision as long as the squash commit keeps the ! (ideally with a BREAKING CHANGE: footer noting DER support shipped since v1.0.0 is removed), the migration docs also cover base64 DER stored in secret backends, and the issue text drops "and DER".

docs/troubleshooting.md
Hermann-Core: Updated in d740a1b1. The troubleshooting guide no longer contains source paths, line ranges, or internal component names; it now stays at the stable operator-facing behavior and exact error strings.

src/utils/crypto.rs
Hermann-Core: Updated in 237570ca. `from_pem` now parses all PEM blocks, ignores auxiliary blocks such as `EC PARAMETERS`, and accepts exactly one supported private-key block. Multiple private-key blocks are rejected as ambiguous. Tests cover the OpenSSL-style EC parameters plus SEC1 key bundle and the ambiguity case.

Hermann-Core:
@martcpp The remaining review-level items are addressed: #524 now defines the filesystem contract as PEM-only, and the migration guidance explains base64/base64url DER secret values must be decoded, converted to PEM, and replaced with PEM text. The PR title retains `feat(crypto)!`, which is the repository-defined breaking-change marker for the eventual squash commit; no manual changelog entry was added.

Hermann-Core:
@Christiantyemele @ndefokou @Ngha-Boris @martcpp Fixed the CI-only revalidation test flake in 0a18eda5 by selecting deterministic token-window boundaries which was introduced by a previous PR; production behavior is unchanged. Please re-review when convenient.

martcpp:
pls take away the [fix(test): stabilize revalidation metrics test](https://github.com/adorsys/status-list-server/pull/559/commits/0a18eda516ea3644e3ad8a871285144fe64dbd0f) and just fix it as a stand alone and then merge because it does not only fail on you end other are having and it out of scope of this issue thanks

Hermann-Core:
> pls take away the [fix(test): stabilize revalidation metrics test](https://github.com/adorsys/status-list-server/pull/559/commits/0a18eda516ea3644e3ad8a871285144fe64dbd0f) and just fix it as a stand alone and then merge because it does not only fail on you end other are having and it out of scope of this issue thanks

It was making my CI to fail sometimes because it is not deterministic. The change is small so we can afford it here instead of creating an issue just to solve this.

martcpp (approved):

ndefokou (approved):

